PR #2, opened 2021-06-14
Add an init command

alexeagle:
This is just the start of logic to help bootstrap a new directory onto Bazel and Aspect


alexeagle:
Moving to silo

PR #11, opened 2021-09-03
feat: setup wizard

alexeagle:


gregmagolan:
This should go in upstream in silo

PR #39, opened 2021-10-01
fix: use interactive flag rather than check isatty

alexeagle:


f0rmiga:
Closing this in favour of the approach in https://github.com/aspect-build/aspect-cli/pull/64.

PR #57, opened 2021-10-23
feat(query): add query verb with initial preset features

mattem:
Add query functionality. Users can provide regular bazel queries that will be passed directly to bazel. Users can also take advantage of predefined queries that will simply prompt the user for bazel label(s) as needed

### DIFF
--- a/cmd/aspect/query/query.go
+++ b/cmd/aspect/query/query.go
@@ -32,6 +32,7 @@ func NewQueryCommand(streams ioutils.Streams, bzl bazel.Spawner) *cobra.Command 
 	cmd := &cobra.Command{
 		Use:   "query",
 		Short: "Executes a dependency graph query.",
+		Long: "Executes a query language expression over a specified subgraph of the build dependency graph.",
 		RunE:  q.Run,
 	}
 
--- a/cmd/aspect/query/query.go
+++ b/cmd/aspect/query/query.go
@@ -37,7 +37,5 @@ func NewQueryCommand(streams ioutils.Streams, bzl bazel.Bazel) *cobra.Command {
 		RunE:  q.Run,
 	}
 
-	cmd.PersistentFlags().BoolVarP(&q.ShowGraph, "show_graph", "", false, `Shows a graph for of the result`)
-
 	return cmd
 }
--- a/cmd/aspect/query/query.go
+++ b/cmd/aspect/query/query.go
@@ -20,7 +20,8 @@ func NewDefaultQueryCmd() *cobra.Command {
 func NewQueryCommand(streams ioutils.Streams, bzl bazel.Bazel) *cobra.Command {
 	q := query.New(streams, bzl, true)
 
-	// Load these from somewhere
+	// TODO: Queries should be loadable from the plugin config
+	// https://github.com/aspect-build/aspect-cli/issues/98
 	q.Presets = []*query.PresetQuery{
 		{
 			Name:        "why",
@@ -32,7 +33,7 @@ func NewQueryCommand(streams ioutils.Streams, bzl bazel.Bazel) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "query",
 		Short: "Executes a dependency graph query.",
-		Long: "Executes a query language expression over a specified subgraph of the build dependency graph.",
+		Long:  "Executes a query language expression over a specified subgraph of the build dependency graph.",
 		RunE:  q.Run,
 	}
 
--- a/cmd/aspect/query/query.go
+++ b/cmd/aspect/query/query.go
@@ -20,16 +20,6 @@ func NewDefaultQueryCmd() *cobra.Command {
 func NewQueryCommand(streams ioutils.Streams, bzl bazel.Bazel) *cobra.Command {
 	q := query.New(streams, bzl, true)
 
-	// TODO: Queries should be loadable from the plugin config
-	// https://github.com/aspect-build/aspect-cli/issues/98
-	q.Presets = []*query.PresetQuery{
-		{
-			Name:        "why",
-			Description: "Determine why targetA depends on targetB",
-			Query:       "somepath(?targetA, ?targetB)",
-		},
-	}
-
 	cmd := &cobra.Command{
 		Use:   "query",
 		Short: "Executes a dependency graph query.",
--- a/cmd/aspect/query/query.go
+++ b/cmd/aspect/query/query.go
@@ -24,8 +24,8 @@ func NewQueryCommand(streams ioutils.Streams, bzl bazel.Bazel) *cobra.Command {
 	q.Presets = []*query.PresetQuery{
 		{
 			Name:        "why",
-			Description: "Determine why a target depends on another",
-			Query:       "somepath(?target, ?dependency)",
+			Description: "Determine why targetA depends on targetB",
+			Query:       "somepath(?targetA, ?targetB)",
 		},
 	}
 
--- a/cmd/aspect/query/query.go
+++ b/cmd/aspect/query/query.go
@@ -1,0 +1,41 @@
+/*
+Copyright © 2021 Aspect Build Systems
+
+Not licensed for re-use
+*/
+
+package query
+
+import (
+	"aspect.build/cli/pkg/aspect/query"
+	"aspect.build/cli/pkg/bazel"
+	"aspect.build/cli/pkg/ioutils"
+	"github.com/spf13/cobra"
+)
+
+func NewDefaultQueryCmd() *cobra.Command {
+	return NewQueryCommand(ioutils.DefaultStreams, bazel.New())
+}
+
+func NewQueryCommand(streams ioutils.Streams, bzl bazel.Spawner) *cobra.Command {
+	q := query.New(streams, bzl, true)
+
+	// Load these from somewhere
+	q.Presets = []*query.PresetQuery{
+		{
+			Name:        "why",
+			Description: "Determine why a target depends on another",
+			Query:       "somepath(?target, ?dependency)",
+		},
+	}
+
+	cmd := &cobra.Command{
+		Use:   "query",
+		Short: "Executes a dependency graph query.",
+		RunE:  q.Run,
+	}
+
+	cmd.PersistentFlags().BoolVarP(&q.ShowGraph, "show_graph", "", false, `Shows a graph for of the result`)
+
+	return cmd
+}
--- a/cmd/aspect/query/query.go
+++ b/cmd/aspect/query/query.go
@@ -17,7 +17,7 @@ func NewDefaultQueryCmd() *cobra.Command {
 	return NewQueryCommand(ioutils.DefaultStreams, bazel.New())
 }
 
-func NewQueryCommand(streams ioutils.Streams, bzl bazel.Spawner) *cobra.Command {
+func NewQueryCommand(streams ioutils.Streams, bzl bazel.Bazel) *cobra.Command {
 	q := query.New(streams, bzl, true)
 
 	// Load these from somewhere
--- a/cmd/aspect/root/BUILD.bazel
+++ b/cmd/aspect/root/BUILD.bazel
@@ -13,6 +13,7 @@ go_library(
         "//cmd/aspect/clean",
         "//cmd/aspect/docs",
         "//cmd/aspect/info",
+        "//cmd/aspect/query",
         "//cmd/aspect/run",
         "//cmd/aspect/test",
         "//cmd/aspect/version",
--- a/docs/help/topics/BUILD.bazel
+++ b/docs/help/topics/BUILD.bazel
@@ -11,6 +11,7 @@ bindata(
     package = "topics",
 )
 
+# keep
 go_library(
     name = "topics",
     srcs = [":bindata"],  # keep
--- a/docs/help/topics/BUILD.bazel
+++ b/docs/help/topics/BUILD.bazel
@@ -11,7 +11,6 @@ bindata(
     package = "topics",
 )
 
-# keep
 go_library(
     name = "topics",
     srcs = [":bindata"],  # keep
--- a/pkg/aspect/query/query.go
+++ b/pkg/aspect/query/query.go
@@ -32,22 +32,27 @@ type PresetQuery struct {
 type Query struct {
 	ioutils.Streams
 
-	bzl           bazel.Bazel
-	isInteractive bool
+	Bzl           bazel.Bazel
+	IsInteractive bool
 
 	Presets   []*PresetQuery
 	ShowGraph bool
+
+	GetAPrompt func(label string) PromptRunner
+	GetASelect func(presetNames []string) SelectRunner
 }
 
 func New(streams ioutils.Streams, bzl bazel.Bazel, isInteractive bool) *Query {
 	return &Query{
 		Streams:       streams,
-		bzl:           bzl,
-		isInteractive: isInteractive,
+		Bzl:           bzl,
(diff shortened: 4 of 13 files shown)


JesseTatasciore:
Can pass through directly to bazel query:
passthrough-query


Can prompt the user to use a preset query when no arguments are given to query:
promt-for-preset


Can specify which preset query to use when calling query:
give-preset-name


Can specify a preset query and give its argument all in the same command: 
give-full-preset

alexeagle:
"determine why a target depends on another" the prompts are "target" and "dependency" and at least to me, it's not clear which direction the arrow points.
Maybe the prompt could include placeholders, like "determine why target A depends on target B" and then "target A?" "target B?"

But, then in your next demo that won't work since `aspect query why` you wouldn't know what "A" and "B" refer to 🤔

cmd/aspect/root/BUILD.bazel
```diff
@@ -13,10 +13,11 @@ go_library(
         "//cmd/aspect/clean",
         "//cmd/aspect/docs",
         "//cmd/aspect/info",
+        "//cmd/aspect/query",
         "//cmd/aspect/root/flags",
         "//cmd/aspect/test",
         "//cmd/aspect/version",
-        "//docs/help/topics",
+        "//docs/help/topics",  # keep
```
mattem, line 20: I think this `keep` can be removed now - the issue I was working around has been resolved

docs/help/topics/BUILD.bazel
```diff
@@ -11,6 +11,7 @@ bindata(
     package = "topics",
 )
 
+# keep
```
mattem, line 14: Same here, can be removed iirc

mattem (commented):
Some notes off the top of my head for this:

* Have we agreed on the preset format? Where does this live? I see we have both an .aspect and a plugins file. I generally dislike so many files in the repo root, and adding another one for presets just adds to that. Can we merge these?

* There seems to be quite a large amount of boiler plate to add a verb, and we want presets for query, cquery and aquery. As an extension to this, it seems tricky to get the flags to pass though?

JesseTatasciore:
@alexeagle does something like this work better?
promt-for-preset-2

give-preset-name-2

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,170 @@
...
+			Err:      err,
+			ExitCode: exitCode,
+		}
+		return err
+	}
+
+	return nil
+}
+
+func (q *Query) RunQueryAndOpenResult(query string, flags []string) error {
```
f0rmiga, line 137: I find this a strange way to divide code. I don't think you need `RunQueryAndOpenResult`. I think you only need `RunQuery`:

1. Check if `q.ShowGraph` and append `--output=graph`.
2. Execute Bazel like you execute here but check again for `q.ShowGraph` and based on the value you either do `io.Copy` from the pipe reader o the stdout (from the default streams) or consume in the way you are consuming now here.

If you want to separate the code for more intelligibility, you can then have 2 private functions that are specialized for printing to output or opening the graph viz.

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,170 @@
...
+	defer close(bazelErrs)
+	go func() {
+		defer w.Close()
+		_, err := q.bzl.RunCommand(bazelCmd, w)
+		bazelErrs <- err
+	}()
+
+	bazelQueryOutput, err := ioutil.ReadAll(r)
+	if err != nil {
+		return fmt.Errorf("failed to get bazel query response: %w", err)
```
f0rmiga, line 157: The more appropriate error message is `failed to fun query %q: %w` like my suggestion above.

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if err := <-bazelErrs; err != nil {
+		return fmt.Errorf("failed to get bazel query response: %w", err)
+	}
+
+	graphVizUrl := fmt.Sprintf("https://edotor.net/?engine=dot#%s", url.PathEscape(string(bazelQueryOutput)))
+	if err := browser.OpenURL(graphVizUrl); err != nil {
+		return fmt.Errorf("failed to open link in the browser: %w", err)
```
f0rmiga, line 166: The error message context (the first part of the error message) should not change in this function. This is important to assure we comply with the best practice and that we can use the error handling in Go 2 once released: https://dev.to/dean/go-2-draft-error-handling-3loo.

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,170 @@
...
+	bazelQueryOutput, err := ioutil.ReadAll(r)
+	if err != nil {
+		return fmt.Errorf("failed to get bazel query response: %w", err)
+	}
+
+	if err := <-bazelErrs; err != nil {
+		return fmt.Errorf("failed to get bazel query response: %w", err)
+	}
+
+	graphVizUrl := fmt.Sprintf("https://edotor.net/?engine=dot#%s", url.PathEscape(string(bazelQueryOutput)))
```
f0rmiga, line 164: Now we have a hard dependency on an online service we don't control? I think this is blocking merging this feature.

pkg/aspect/query/query_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"aspect.build/cli/pkg/ioutils"
+)
+
+type confirm struct{}
+
+func (p confirm) Run() (string, error) {
+	return "", nil
+}
+
+func TestQuery(t *testing.T) {
```
f0rmiga, line 26: We could use some more tests that go over more code paths.

f0rmiga (changes requested):

pkg/aspect/query/query.go
mattem: I think this whole part should be removed, it was a PoC and we would build our own UI and launch that

f0rmiga: We should consider using https://github.com/wailsapp/wails.

cmd/aspect/query/query.go
```diff
@@ -0,0 +1,42 @@
...
+
+func NewDefaultQueryCmd() *cobra.Command {
+	return NewQueryCommand(ioutils.DefaultStreams, bazel.New())
+}
+
+func NewQueryCommand(streams ioutils.Streams, bzl bazel.Bazel) *cobra.Command {
+	q := query.New(streams, bzl, true)
+
+	// Load these from somewhere
+	q.Presets = []*query.PresetQuery{
```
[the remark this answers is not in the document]
JesseTatasciore: Ill make it more clear. Basically the presets are hardcoded right there in that presets list. I think the comment was meant to mean that "some of these should be loaded from the plugin config". Which there is an open issue for btw so it will come in a later PR

cmd/aspect/query/query.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+func NewDefaultQueryCmd() *cobra.Command {
+	return NewQueryCommand(ioutils.DefaultStreams, bazel.New())
+}
+
+func NewQueryCommand(streams ioutils.Streams, bzl bazel.Bazel) *cobra.Command {
+	q := query.New(streams, bzl, true)
+
+	// Load these from somewhere
```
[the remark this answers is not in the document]
JesseTatasciore: Ill make it more clear. Basically the presets are hardcoded right there in that presets list. I think the comment was meant to mean that "some of these should be loaded from the plugin config". Which there is an open issue for btw so it will come in a later PR

pkg/aspect/query/query_test.go
JesseTatasciore: Agreed. It needs some more tests. I got a bit hung up trying to figure out how to mock the inputs to `promptui`. (As several of the code paths rely on its inputs)

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+func (q *Query) Run(_ *cobra.Command, args []string) error {
+	presets := make(map[string]*PresetQuery)
+	var names []string
+	for _, p := range q.Presets {
+		presets[p.Name] = p
+		names = append(names, fmt.Sprintf("%s: %s", p.Name, p.Description))
+	}
```
f0rmiga, line 56: 1. I'd choose a more descriptive label for `names` like `presetNames`.
2. Since this is simply mapping `q.Presets` to another format, you know the final length of `names`, you better ask for the correct length since initialization and avoid memory copying on each `append`.
3. You would get weird displaying bugs for duplicated preset names. While we don't have a mechanism for allowing users to override query presets, my suggestion below should protect you.

```suggestion
	presetNames := make([]string, len(q.Presets))
	for i, p := range q.Presets {
		if _, exists := presets[p.Name]; exists {
			err := fmt.Errorf("duplicated preset query name %q", p.Name)
			return fmt.Errorf("failed to run 'aspect %s': %w", cmd.Use, err)
		}
		presets[p.Name] = p
		presetNames[i] = fmt.Sprintf("%s: %s", p.Name, p.Description)
	}
```

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,169 @@
...
+
+	var preset *PresetQuery
+	if len(args) == 0 {
+		selectQueryPrompt := &promptui.Select{
+			Label: "Select a preset query",
+			Items: names,
+		}
+
+		i, _, err := selectQueryPrompt.Run()
```
f0rmiga, line 65: Answering your question about how to test this, you would need to do some dependency injection here. This is not trivial in this case, but the best you can do is to have a `promptui.Select` factory that gets injected in the `Query` constructor and allows you to mock it during testing to return another mocked interface that gets `Run()` and you control both `i` and `err`. Let's pair on this and I show you how to do it.

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,169 @@
...
+		preset = q.Presets[i]
+	} else {
+		maybeQueryOrPreset := args[0]
+		if value, ok := presets[maybeQueryOrPreset]; ok {
+			// Treat this as the name of the preset query, so don't prompt for it.
+			fmt.Fprintf(q.Streams.Stdout, "Preset query \"%s\" selected\n", value.Name)
+			fmt.Fprintf(q.Streams.Stdout, "%s: %s\n", value.Name, value.Description)
+			preset = value
+		} else {
+			// Treat this as a raw query expression
```
f0rmiga, line 79:
```suggestion
			// Treat this as a raw query expression.
```

pkg/aspect/query/query.go
JesseTatasciore: That would be great. Let me know when you have some time. No rush

pkg/aspect/query/query.go
JesseTatasciore: Function has been removed

pkg/aspect/query/query.go
JesseTatasciore: I have been removed the showGraph functionality and created an issue: https://github.com/aspect-build/aspect-cli/issues/114

pkg/aspect/query/query.go
JesseTatasciore: Function has been removed

pkg/aspect/query/query.go
JesseTatasciore: Function has been removed

cmd/aspect/query/query.go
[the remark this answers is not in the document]
JesseTatasciore: Moved

cmd/aspect/query/query.go
```diff
@@ -0,0 +1,31 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "query",
+		Short: "Executes a dependency graph query.",
+		Long:  "Executes a query language expression over a specified subgraph of the build dependency graph.",
+		RunE:  q.Run,
+	}
+
+	return cmd
+}
```
f0rmiga, line 31: Perfect.

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,163 @@
...
+type Query struct {
+	ioutils.Streams
+
+	Bzl           bazel.Bazel
+	IsInteractive bool
+
+	Presets []*PresetQuery
+
+	GetAPrompt func(label string) PromptRunner
+	GetASelect func(presetNames []string) SelectRunner
```
f0rmiga, line 37: These function names were just an example. Not sure they are the most appropriate names. We can rename them on a follow-up PR.

f0rmiga: Maybe a more Go-idiomatic would be:

```suggestion
	Prompt func(label string) PromptRunner
	Select func(presetNames []string) SelectRunner
```

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	var preset *PresetQuery
+	if len(args) == 0 {
+		selectQueryPrompt := q.GetASelect(presetNames)
+
+		i, _, err := selectQueryPrompt.Run()
+
+		if err != nil {
+			return fmt.Errorf("prompt failed: %w", err)
```
f0rmiga, line 80:
```suggestion
			return fmt.Errorf("failed to run 'aspect %s': %w", cmd.Use, err)
```

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,163 @@
...
+			fmt.Fprintf(q.Streams.Stdout, "%s: %s\n", value.Name, value.Description)
+			preset = value
+		} else {
+			// Treat this as a raw query expression.
+			return q.RunQuery(maybeQueryOrPreset)
+		}
+	}
+
+	if preset == nil {
+		return fmt.Errorf("unable to determine preset query")
```
f0rmiga, line 98:
```suggestion
		err := fmt.Errorf("unable to determine preset query")
		return fmt.Errorf("failed to run 'aspect %s': %w", cmd.Use, err)
```

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,163 @@
...
+			query = strings.ReplaceAll(query, placeholder, args[i+1])
+		}
+	} else if len(placeholders) > 0 {
+		for _, placeholder := range placeholders {
+			label := fmt.Sprintf("Value for '%s'", strings.TrimPrefix(placeholder, "?"))
+			prompt := q.GetAPrompt(label)
+			val, err := prompt.Run()
+
+			if err != nil {
+				return fmt.Errorf("prompt failed: %w", err)
```
f0rmiga, line 115:
```suggestion
				return fmt.Errorf("failed to run 'aspect %s': %w", cmd.Use, err)
```

pkg/aspect/query/query.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (q *Query) RunQuery(query string) error {
+	bazelCmd := []string{
+		"query",
+		query,
+	}
+
+	bazelCmd = append(bazelCmd)
+
+	if exitCode, err := q.Bzl.Spawn(bazelCmd); exitCode != 0 {
```
f0rmiga, line 154:
```suggestion
	if exitCode, err := q.Bzl.Spawn(bazelCmd); exitCode != 0 {
```

f0rmiga:
@JesseTatasciore We have to keep our errors compliant with https://go.googlesource.com/proposal/+/master/design/go2draft-error-handling-overview.md.

Kinda hard to enforce but we should strive for it.

f0rmiga:
I think you have to rebase before merging because of my changes to docs generation.

f0rmiga (approved):

PR #61, opened 2021-11-11
feat: interactive mode for `aspect build`

cvarier:
Addresses https://github.com/aspect-build/aspect-cli/issues/42. Will implement same for `aspect test` in subsequent PR, and then close issue.

### DIFF
--- a/pkg/aspect/build/build.go
+++ b/pkg/aspect/build/build.go
@@ -10,9 +10,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"time"
 
-	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
 	buildeventstream "aspect.build/cli/bazel/buildeventstream/proto"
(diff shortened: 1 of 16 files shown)


pkg/aspect/build/build.go
```diff
@@ -29,7 +30,8 @@ const (
 	skipPromptKey = "build.skip_prompt"
 
 	SpecifiedFolderOption = "All targets in a specified folder (path relative to your WORKSPACE)"
-	CurrentFolderOption   = "All targets in current folder"
+	CurrentFolderOption   = "All targets within current folder"
+	CurrentPackageOption  = "All targets within current package (nearest parent if current folder is not a package)"
```
alexeagle, line 34: I don't think you need the parenthetical - the "current package" is always defined this way

pkg/aspect/build/build.go
```diff
@@ -29,7 +30,8 @@ const (
 	skipPromptKey = "build.skip_prompt"
 
 	SpecifiedFolderOption = "All targets in a specified folder (path relative to your WORKSPACE)"
-	CurrentFolderOption   = "All targets in current folder"
+	CurrentFolderOption   = "All targets within current folder"
```
alexeagle, line 33: I don't think we need this one

PR #64, opened 2021-11-15
feat: plugin system

### LINKED ISSUE #45: feature: plugin system using hashicorp/go-plugin
Initially, this should support the current `fix-visibility` plugin.

f0rmiga:
Closes #45.


pkg/plugins/v1alpha1/plugin/interface.go
```diff
@@ -0,0 +1,21 @@
...
+*/
+
+package plugin
+
+import (
+	buildeventstream "aspect.build/cli/bazel/buildeventstream/proto"
+	"aspect.build/cli/pkg/ioutils"
+)
+
+// Plugin determines how an aspect Plugin should be implemented.
```
alexeagle, line 14: this is handy for plugins written in Go, should we mention that it's optional for plugins in other languages?

also, why isn't this produced in the protoc output .pb.go file instead?

alexeagle (approved):
All looks great to me. We should pivot to writing a basic plugin authors guide, and confirm we can write a plugin outside the repo.

pkg/plugins/v1alpha1/plugin/interface.go
f0rmiga: I think we can mention it in the SDK doc.

As for your second question, I wanted to abstract the internal RPC communication so that the user doesn't need to care whether it's gRPC or not. It doesn't feel like the user of the SDK is doing gRPC, it's just implementing a simple Go interface.

alexeagle (approved):

PR #68, opened 2021-11-17
fix: all of `build`, `test`, `clean`, `info` should fail immediately when invoked outside of workspace

cvarier:
Currently `build` hangs indefinitely if invoked outside of a workspace. `test`, `clean`, `info` will chug for a bit before Bazel spits out the error. This PR will cause all commands to fail immediately if invoked outside of a workspace.

### DIFF
--- a/cmd/aspect/clean/clean.go
+++ b/cmd/aspect/clean/clean.go
@@ -7,16 +7,16 @@ Not licensed for re-use.
 package clean
 
 import (
-	rootFlags "aspect.build/cli/cmd/aspect/root/flags"
-	"aspect.build/cli/pkg/pathutils"
 	"github.com/spf13/cobra"
 
+	rootFlags "aspect.build/cli/cmd/aspect/root/flags"
 	"aspect.build/cli/pkg/aspect/clean"
+	"aspect.build/cli/pkg/pathutils"
 )
 
 // NewDefaultCleanCmd creates a new clean cobra command.
 func NewDefaultCleanCmd() *cobra.Command {
-	c := clean.NewDefault()
+	cleanCmd := clean.NewDefault()
 
 	cmd := &cobra.Command{
 		Use:   "clean",
@@ -52,24 +52,24 @@ Workaround inconistent state:
 	If you ever find an incorrect incremental build, please file a bug report,
 	and only use clean as a temporary workaround.`,
 		RunE: func(cmd *cobra.Command, args []string) (exitErr error) {
-			return pathutils.InvokeCmdInsideWorkspace("clean", func() error {
+			return pathutils.InvokeCmdInsideWorkspace(cmd.Use, func() error {
 				isInteractiveMode, err := cmd.Root().PersistentFlags().GetBool(rootFlags.InteractiveFlagName)
 				if err != nil {
 					return err
 				}
 
-				return c.Run(cmd, args, isInteractiveMode)
+				return cleanCmd.Run(isInteractiveMode)
 			})
 		},
 	}
 
-	cmd.PersistentFlags().BoolVarP(&c.Expunge, "expunge", "", false, `Remove the entire output_base tree.
+	cmd.PersistentFlags().BoolVarP(&cleanCmd.Expunge, "expunge", "", false, `Remove the entire output_base tree.
 This removes all build output, external repositories,
 and temp files created by Bazel.
 It also stops the Bazel server after the clean,
 equivalent to the shutdown command.`)
 
-	cmd.PersistentFlags().BoolVarP(&c.ExpungeAsync, "expunge_async", "", false, `Expunge in the background.
+	cmd.PersistentFlags().BoolVarP(&cleanCmd.ExpungeAsync, "expunge_async", "", false, `Expunge in the background.
 It is safe to invoke a Bazel command in the same
 workspace while the asynchronous expunge continues to run.
 Note, however, that this may introduce IO contention.`)
--- a/cmd/aspect/clean/clean.go
+++ b/cmd/aspect/clean/clean.go
@@ -7,7 +7,7 @@ Not licensed for re-use.
 package clean
 
 import (
-	"aspect.build/cli/pkg/ioutils"
+	rootFlags "aspect.build/cli/cmd/aspect/root/flags"
 	"aspect.build/cli/pkg/pathutils"
 	"github.com/spf13/cobra"
 
@@ -16,8 +16,7 @@ import (
 
 // NewDefaultCleanCmd creates a new clean cobra command.
 func NewDefaultCleanCmd() *cobra.Command {
-	isInteractive := ioutils.IsInteractive()
-	c := clean.NewDefault(isInteractive)
+	c := clean.NewDefault()
 
 	cmd := &cobra.Command{
 		Use:   "clean",
@@ -58,7 +57,12 @@ Workaround inconistent state:
 				return err
 			}
 
-			return c.Run(cmd, args)
+			isInteractiveMode, err := cmd.Root().PersistentFlags().GetBool(rootFlags.InteractiveFlagName)
+			if err != nil {
+				return err
+			}
+
+			return c.Run(cmd, args, isInteractiveMode)
 		},
 	}
 
--- a/cmd/aspect/clean/clean.go
+++ b/cmd/aspect/clean/clean.go
@@ -7,9 +7,7 @@ Not licensed for re-use.
 package clean
 
 import (
-	"os"
-
-	"github.com/mattn/go-isatty"
+	"aspect.build/cli/pkg/ioutils"
 	"github.com/spf13/cobra"
 
 	"aspect.build/cli/pkg/aspect/clean"
@@ -17,7 +15,7 @@ import (
 
 // NewDefaultCleanCmd creates a new clean cobra command.
 func NewDefaultCleanCmd() *cobra.Command {
-	isInteractive := isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd())
+	isInteractive := ioutils.IsInteractive()
 	b := clean.NewDefault(isInteractive)
 
 	cmd := &cobra.Command{
--- a/cmd/aspect/clean/clean.go
+++ b/cmd/aspect/clean/clean.go
@@ -52,17 +52,14 @@ Workaround inconistent state:
 	If you ever find an incorrect incremental build, please file a bug report,
 	and only use clean as a temporary workaround.`,
 		RunE: func(cmd *cobra.Command, args []string) (exitErr error) {
-			err := pathutils.CmdNotInvokedInsideWorkspace("clean")
-			if err != nil {
-				return err
-			}
+			return pathutils.InvokeCmdInsideWorkspace("clean", func() error {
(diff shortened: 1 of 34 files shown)


pkg/ioutils/prompt.go
```diff
@@ -6,10 +6,19 @@ Not licensed for re-use.
...
+)
 
 // PromptRunner is the interface that wraps the promptui.Prompt and makes a call
 // to it from the aspect CLI Core.
 type PromptRunner interface {
 	Run(prompt promptui.Prompt) (string, error)
 }
+
+// Check if the CLI can be run in interactive mode
+func IsInteractive() bool {
```
alexeagle, line 22: you shouldn't check this, because there's also a `--interactive` flag that can be passed. Check Thulio's recent change for how to observe the value

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func CmdNotInvokedInsideWorkspace(cmdName string) error {
+	workingDirectory, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("could not resolve working directory: %w", err)
+	}
+	workspaceRoot := FindWorkspaceRoot(workingDirectory)
+	if workspaceRoot == "" {
+		return fmt.Errorf("the '%s' command is only supported from within a workspace", cmdName)
```
alexeagle, line 58: this is a great example of a bad bazel error message where we can do better. What is a workspace?

alexeagle (commented):
I'll let @f0rmiga do most of the review.

At some point the behavior of running `aspect` outside a workspace should be "would you like to initialize a new bazel workspace in this folder" but not shovel-ready yet.

cmd/aspect/clean/clean.go
```diff
@@ -53,16 +51,25 @@ Workaround inconistent state:
 	Such problems are fixable and these bugs are a high priority.
 	If you ever find an incorrect incremental build, please file a bug report,
 	and only use clean as a temporary workaround.`,
-		RunE: b.Run,
+		RunE: func(cmd *cobra.Command, args []string) (exitErr error) {
+			return pathutils.InvokeCmdInsideWorkspace("clean", func() error {
```
f0rmiga, line 55: Instead of hardcode `"clean"`, you can do:

```suggestion
			return pathutils.InvokeCmdInsideWorkspace(cmd.Use, func() error {
```

cmd/aspect/info/info.go
```diff
@@ -18,7 +19,7 @@ func NewDefaultInfoCmd() *cobra.Command {
 }
 
 func NewInfoCmd(streams ioutils.Streams) *cobra.Command {
-	v := info.New(streams)
+	i := info.New(streams)
```
f0rmiga, line 22: Single-letter variables are tricky to use. The rules are implicit and they can usually lead to a lack of intelligibility of the code.

E.g. `i` is well-known to represent `index` in any language. Perhaps we should stop using single-letter vars for these commands as the scope of the `New<name>Cmd` functions have been increasing recently.

cmd/aspect/info/info.go
```diff
@@ -42,10 +43,14 @@ the bazel User Manual, and can be programmatically obtained with
 See also 'bazel version' for more detailed bazel version
 information.`,
 		Args: cobra.MaximumNArgs(1),
-		RunE: v.Run,
+		RunE: func(cmd *cobra.Command, args []string) (exitErr error) {
+			return pathutils.InvokeCmdInsideWorkspace("info", func() error {
```
f0rmiga, line 47: Same here, `cmd.Use`.

cmd/aspect/test/test.go
```diff
@@ -7,6 +7,7 @@ Not licensed for re-use
 package test
 
 import (
+	"aspect.build/cli/pkg/pathutils"
```
f0rmiga, line 10: Please, move this import statement to the bottom. The convention is usually:

```
import (
<std packages>

<third-party packages>

<first-party packages>
)
```

cmd/aspect/test/test.go
```diff
@@ -35,7 +36,11 @@ don't forget to pass all your 'build' options to 'test' too.
 See 'bazel help target-syntax' for details and examples on how to
 specify targets.
 `,
-		RunE: v.Run,
+		RunE: func(cmd *cobra.Command, args []string) (exitErr error) {
+			return pathutils.InvokeCmdInsideWorkspace("test", func() error {
```
f0rmiga, line 40: `cmd.Use`

pkg/aspect/clean/clean.go
```diff
@@ -70,20 +68,17 @@ type Clean struct {
 // New creates a Clean command.
 func New(
 	streams ioutils.Streams,
-	bzl bazel.Spawner,
-	isInteractiveMode bool) *Clean {
+	bzl bazel.Spawner) *Clean {
```
f0rmiga, line 71:
```suggestion
	bzl bazel.Spawner,
) *Clean {
```

pkg/aspect/clean/clean.go
```diff
@@ -107,9 +102,9 @@ func NewDefault(isInteractive bool) *Clean {
 }
 
 // Run runs the aspect build command.
```
f0rmiga, line 104: Wooops, copy-pasta left-over.

pkg/aspect/clean/clean.go
```diff
@@ -107,9 +102,9 @@ func NewDefault(isInteractive bool) *Clean {
...
-func (c *Clean) Run(_ *cobra.Command, _ []string) error {
+func (c *Clean) Run(_ *cobra.Command, _ []string, isInteractiveMode bool) error {
```
f0rmiga, line 105: Since you are here - if we are not using `cmd` and `args`, and we don't plan to use them, let's just remove the args here. The commands don't satisfy any interface to enforce this signature.

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,66 @@
+/*
```
f0rmiga, line 1: Nice package, but we will need some tests for this.

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,66 @@
...
+
+package pathutils
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func IsValidFile(path string) bool {
```
f0rmiga, line 14: `IsValidFile` does not represent well what this function does. `IsFile` is probably sufficient. The `Valid` word implies it does some sort of validation, which it does not. Alternatively, you can probe if the file is readable/writable:

```
func IsFileReadable(path string) bool {
  const R_OK uint32 = 4
  return syscall.Access(path, R_OK) != syscall.EACCES
}
```

For reference, the second argument to `syscall.Access` can be copied from the C code `unistd.h`:

```
#define	R_OK	4		/* Test for read permission.  */
#define	W_OK	2		/* Test for write permission.  */
#define	X_OK	1		/* Test for execute permission.  */
#define	F_OK	0		/* Test for existence.  */
```

You see, `F_OK` can be used for verifying if it exists or not too.

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,66 @@
...
+	if IsValidWorkspace(path) {
+		return path
+	}
+
+	parentDirectory := filepath.Dir(path)
+	if parentDirectory == path {
+		return ""
+	}
+
+	return FindWorkspaceRoot(parentDirectory)
```
f0rmiga, line 48: Hmmm... I'd avoid recursive functions in the Aspect CLI in general as it keeps growing the stack.

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,66 @@
...
+
+func FindWorkspaceRoot(path string) string {
+	if IsValidWorkspace(path) {
+		return path
+	}
+
+	parentDirectory := filepath.Dir(path)
+	if parentDirectory == path {
+		return ""
+	}
```
f0rmiga, line 46: I don't really understand this check. It would only be true if the path is relative, and it is `.`. Otherwise, if I'm reading correctly, the recursive function will run forever if it's not a Bazel workspace. Combined with the growing stack, you will run out of memory.

f0rmiga (changes requested):

cmd/aspect/info/info.go
cvarier: That's a good point. Ok, I'll refactor all of these.

pkg/ioutils/prompt.go
alexeagle: there's existing logic that consults both the flag, and the execution environment, you need to find and reuse that

pkg/pathutils/pathutils.go
cvarier: will add shortly

pkg/pathutils/pathutils.go
cvarier: Nice suggestion. I don't think probing whether we have r/w access to the file is necessary, as we are using this helper to perform a simple check if a file exists.

pkg/pathutils/pathutils.go
cvarier: Good point. Theoretically this could end up in stack overflow, but I'd imagine the limit would be quite hard to reach - the project would have to have an insanely deeply nested structure. I thought I'd leave it as is since I moved it over from bazelisk.go.

pkg/ioutils/prompt.go
f0rmiga: @alexeagle I synced with Chai yesterday over a call to go over this already.

pkg/pathutils/pathutils.go
f0rmiga: Alright. Doing the `syscall.Access` may make this whole package easier to test.

pkg/pathutils/pathutils.go
f0rmiga: Actually, a simple mistake like the one I pointed out above can lead to this undesired behaviour.
You don't need to fix it in this PR, but we should fix it right next.

pkg/pathutils/pathutils.go
cvarier: No, this actually works correctly. This stopping condition occurs when we've reached the root directory on disk, ie. when the current folder's parent is itself.

cvarier: `filepath.Dir(path)` gets the parent directory of `path`. I'll add a comment describing this.

pkg/pathutils/pathutils.go
cvarier: I just tried out your suggestion, but doesn't seem to build even after importing `syscall`. I think this is an os-dependent API, and according to https://pkg.go.dev/syscall, it seems like `syscall` is deprecated. Their suggestion is to use the packages here https://pkg.go.dev/golang.org/x/sys according to the platform that we're targetting.

Since we want our CLI to be cross-compatible, we probably shouldn't overcomplicate this with platform-specific syscall branches.

cmd/aspect/test/test.go
cvarier: I'm thinking we should define a pre-commit hook for this - I don't know if there's some 3rd-party linter that does this?

PR #70, opened 2021-11-18
refactor: use buildeventstream alias for clarity

f0rmiga:
Since the buildeventstream proto is vendored and lives under third-party, we do some workarounds to make it appear to come from //bazel/buildeventstream/proto.

We enforce it with Gazelle now. Makes it simpler to reason about when importing `aspect.build/cli/bazel/buildeventstream/proto`.

### DIFF
--- a/bazel/buildeventstream/proto/BUILD.bazel
+++ b/bazel/buildeventstream/proto/BUILD.bazel
@@ -1,1 +1,7 @@
 # gazelle:exclude dummy.go
+
+alias(
+    name = "proto",
+    actual = "//third-party/github.com/bazelbuild/bazel/src/main/java/com/google/devtools/build/lib/buildeventstream/proto",
+    visibility = ["//visibility:public"],
+)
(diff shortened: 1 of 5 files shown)


bazel/buildeventstream/proto/BUILD.bazel
```diff
@@ -1 +1,7 @@
 # gazelle:exclude dummy.go
+
+alias(
+    name = "proto",
+    actual = "//third-party/github.com/bazelbuild/bazel/src/main/java/com/google/devtools/build/lib/buildeventstream/proto",
```
alexeagle, line 5: nit: I think it's also fine for us to vendor the file itself here, the third-party directory structure isn't mandated anywhere

alexeagle (approved):

bazel/buildeventstream/proto/BUILD.bazel
f0rmiga: Yeah, I believe there's no obligation to do the way I did. Still, I like to keep it there as it makes it clear the file is vendored and not authored by us.

PR #79, opened 2021-11-23
feat: workspace finder

cvarier:
- Introduce `pkg/pathutils/pathutils.go` with `WorskpaceFinder`
- In a subsequent PR, will remove the corresponding methods from `bazelisk.go` and refactor `build`, `test`, `clean`, `info` commands to use `InvokeCmdInsideWorkspace`


pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,80 @@
+/*
+Copyright © 2021 Aspect Build Systems Inc
+Not licensed for re-use.
+*/
```
f0rmiga, line 4:
```suggestion
/*
Copyright © 2021 Aspect Build Systems Inc

Not licensed for re-use.
*/
```

f0rmiga: This is just to match the rest of the files we have in the repo.

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,80 @@
...
+	"path"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+)
+
+// https://github.com/bazelbuild/bazel/blob/8346ea4c/src/main/cpp/workspace_layout.cc#L37
+var workspaceFilenames = []string{"WORKSPACE", "WORKSPACE.bazel"}
+
+type RunFn func(cmd *cobra.Command, args []string) (exitErr error)
```
f0rmiga, line 21:
```suggestion
// RunFn is the function signature for the cobra RunE function in the cobra.Command.
type RunFn func(cmd *cobra.Command, args []string) (exitErr error)
```

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,80 @@
...
+
+func InvokeCmdInsideWorkspace(toRun RunFn) (wrapped RunFn) {
```
f0rmiga, line 23:
```suggestion
// InvokeCmdInsideWorkspace verifies if the current working directory is inside a Bazel workspace, then invokes the provided toRun function.
func InvokeCmdInsideWorkspace(toRun RunFn) RunFn {
```

f0rmiga: I also removed the named return argument `wrapped` since we don't use it as a named arg.

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,80 @@
...
+			return fmt.Errorf("failed to run command %q inside workspace: %w", cmd.Use, err)
+		}
+		return toRun(cmd, args)
+	}
+}
+
+// WorkspaceFinder is the interface that wraps the simple Find method that performs the
+// finding of the WORKSPACE file in the user's Bazel project.
+type WorkspaceFinder struct {
+	osGetwd func() (string, error)
```
f0rmiga, line 46: You don't need this since the working directory is being passed as an argument to `Find`.

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,80 @@
...
+		if workspacePath == "" {
+			err = fmt.Errorf("the current working directory %q is not a bazel workspace " +
+				"(below a directory having a WORKSPACE file)", cwd)
+			return fmt.Errorf("failed to run command %q inside workspace: %w", cmd.Use, err)
+		}
+		return toRun(cmd, args)
+	}
+}
+
+// WorkspaceFinder is the interface that wraps the simple Find method that performs the
```
f0rmiga, line 43: The comment here is wrong. It's not an interface.

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,80 @@
...
+// finding of the WORKSPACE file in the user's Bazel project.
+type WorkspaceFinder struct {
+	osGetwd func() (string, error)
+	osStat  func(string) (fs.FileInfo, error)
+}
+
+// NewDefaultWorkspaceFinder instantiates a default internal implementation of the WorkspaceFinder
```
f0rmiga, line 50: It's not an internal implementation since the `WorkspaceFinder` is an exported struct (WorkspaceFinder vs workspaceFinder).

pkg/pathutils/pathutils.go
```diff
@@ -0,0 +1,80 @@
...
+func NewDefaultWorkspaceFinder() *WorkspaceFinder {
+	return &WorkspaceFinder{
+		osGetwd: os.Getwd,
+		osStat:  os.Stat,
+	}
+}
+
+// Find finds the WORKSPACE file under a Bazel workspace. If the returned
+// path is empty and no error was produced, the user's current working directory
+// is not a Bazel workspace.
```
f0rmiga, line 61:
```suggestion
// Find tries to find a file that marks the root of a Bazel workspace (WORKSPACE or WORKSPACE.bazel). If the returned
// path is empty and no error was produced, the user's current working directory
// is not a Bazel workspace.
```

pkg/pathutils/pathutils_test.go
```diff
@@ -0,0 +1,64 @@
+package pathutils
```
f0rmiga, line 1: Missing license header.

pkg/pathutils/pathutils_test.go
```diff
@@ -0,0 +1,64 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+
+	infocmd "aspect.build/cli/cmd/aspect/info"
+	infopkg "aspect.build/cli/pkg/aspect/info"
+	"aspect.build/cli/pkg/ioutils"
+	"aspect.build/cli/pkg/pathutils"
+)
+
+func TestInvokeCmdInsideWorkspace(t *testing.T) {
```
f0rmiga, line 18: This was supposed to be a unit test, not an integration test. Performing disk IO operations with fixtures is not what we want in the `pkg` sub-packages.

To properly test the logic without calling `os.Stat` and `os.Getwd`, you will need to mock them.

f0rmiga (changes requested):

.gitignore
```diff
@@ -2,3 +2,5 @@
 /bazel-*
 /.idea
 /.cache
+/.ijwb
+/Library
```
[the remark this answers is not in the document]
cvarier: I think my unit test is doing that when I set $HOME temporarily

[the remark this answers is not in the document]
f0rmiga: Yeah, that testing is bad. We want to do proper integration tests, but not this way.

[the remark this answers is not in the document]
f0rmiga: Revert this change, please. You can keep `/.ijwb`, which I believe is from intellij?

pkg/pathutils/pathutils_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/spf13/cobra"
+)
+
+func getMockCmd(osGetwd func() (string, error)) *cobra.Command {
+	return &cobra.Command{
+		Use:   "mock",
+		Short: "This is a mock command",
+		Long:  "This is a mock command",
+		Args:  cobra.MaximumNArgs(1),
+		RunE: invokeCmdInsideWorkspace(
```
f0rmiga, line 22: Don't worry about trying to integrate with the user. From a unit test perspective, you don't care who your user is, you only care that the unitary logic behaves correctly with the input that you allow to be passed.

The tests should be split into 2 top-level `Test*` functions:
- TestWorkspaceFinder - you will mock the `osStat` to make it fail and assert that the `Find` method fails as expected, and play with the various possible `cwd` strings that could be passed to the `Find` method under ideal circumstances.
- TestInvokeInsideWorkspace - you will have to pass the `cobra.Command` to the returned function, which you will call it, but you will not assert that the `cmd` gets called. You will have to assert that the `cmd` gets passed to the `toRun`, though. You will never use the `defaultWorkspaceFinder` on your tests. You can create a `mock` directory and use the same pattern we've been using for other packages using [gomock](https://github.com/golang/mock).

f0rmiga (changes requested):

PR #86, opened 2021-12-11
feat: interceptors system

f0rmiga:
Adds a system similar to the one seen on gRPC that is able to intercept the command executions and inject arbitrary values into the context.

It makes the separation of concerns and the implementation of extra functionality simpler for the commands.

Currently, we use 2 interceptors:
- The one that verifies whether a command is being run inside a Bazel workspace, injecting the workspace root into the context.
- The one that starts and injects a BES backend for the build command.

For more details, see the `Example` test for the interceptors `Run` function.


JesseTatasciore (approved):
My golang and CLI knowledge is still pretty basic but it looks okay to me

PR #89, opened 2022-01-05
fix: CI cache and rename pipeline

f0rmiga:
Since bazelrc doesn't support environment variables, we need to set the disk cache and repository cache explicitly in the bazel command.

Also, renamed the CI workflow to Default since CI is too generic. While Default is still generic, it's enough to differentiate from the Release workflow.


f0rmiga:
After the fix, the commit https://github.com/aspect-build/aspect-cli/pull/89/commits/b3a0e04a56ae82304d11ac8c1ecd0da8eb2771bc was completely empty and ran in 2m with the cache properly restored.

.github/workflows/ci.bazelrc
```diff
@@ -2,6 +2,4 @@
 # Included with a --bazelrc option in the call to bazel in main.yml
 build --announce_rc
 test --test_output=errors
-build --disk_cache=.cache/bazel
```
alexeagle, line 5: I still don't quite understand why it doesn't work to have this relative to the working directory?

Also, this is one of many spots we need to change, if it's actually wrong. https://github.com/bazel-contrib/rules-template/blob/main/.github/workflows/ci.bazelrc#L10-L11 has been stamped out into a number of rulesets including some of our own.

.github/workflows/default.yml
```diff
@@ -0,0 +1,35 @@
...
+          path: |
+            ~/.cache/bazel
+            ~/.cache/bazel-repo
+          key: bazel-cache
+      - name: bazel test //...
+        env:
+          XDG_CACHE_HOME: ~/.cache/bazel-repo
+        run: |
+          bazel \
+            --bazelrc=.github/workflows/ci.bazelrc \
```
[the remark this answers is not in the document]
f0rmiga: I'm not sure yet, this is why I kept it.

.github/workflows/ci.bazelrc
f0rmiga: I don't understand either, but it didn't restore the cache.

For environment variables in bazelrc:
https://github.com/bazelbuild/bazel/issues/10904

alexeagle: So do we have the same problem in aspect-build/bazel-lib for example? I just want to make sure we don't accidentally fix only a single instance of a general problem.

f0rmiga: I think they have the same problem we used to have in this repo, which is a directory named $HOME inside the workspace.

image

f0rmiga: And if you look at their caches, they don't change, it's always a cache hit. On the `aspect-build/bazel-lib` we are fortunate that it's a small repo.

alexeagle (approved):

PR #91, opened 2022-01-05
feat(plugins): Wire the test verb up to the CLI plugin infrastructure

JesseTatasciore:
plugins-via-test


pkg/aspect/test/test_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"aspect.build/cli/pkg/ioutils"
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/gomega"
+	bep_mock "aspect.build/cli/pkg/plugin/system/bep/mock"
```
f0rmiga, line 11: Could you reorder the import statements here to separate the third-party from the first-party packages, please?

pkg/plugin/sdk/v1alpha1/plugin/grpc.go
```diff
@@ -74,6 +74,25 @@ func (m *GRPCServer) PostBuildHook(
 		m.Impl.PostBuildHook(req.IsInteractiveMode, prompter)
 }
 
+// PostTestHook translates the gRPC call to the Plugin PostTestHook
```
f0rmiga, line 77: You will have to copy the whole `v1alpha1` package into `v1alpha2` package. `v1alpha1` should be immutable at this point since we already released it.

f0rmiga (changes requested):
Awesome!

pkg/plugin/sdk/v1alpha1/plugin/grpc.go
f0rmiga: Thinking about this, we should have a test on CI that prevents any changes to previous SDKs. Maybe the best way is to detect changes in `pkg/plugins/sdk/` and only allow for the next release - hardcoded -, in this case, `v1alpha2`.

pkg/aspect/test/test_test.go
JesseTatasciore: Done. I am assuming golang has some linters available. Can we set one up to catch issues like this?

f0rmiga: I don't think it's needed. The auto importer and formatter will detect our logic and keep them correct going forward.

JesseTatasciore: Sounds good

pkg/plugin/sdk/v1alpha1/plugin/grpc.go
JesseTatasciore: Ive created `v1alpha2`. Are you wanting me to add this test now? Or should I create an issue for it?

f0rmiga: Addressed in https://github.com/aspect-build/aspect-cli/pull/99.

f0rmiga (approved):

PR #100, opened 2022-01-07
test: should fail

f0rmiga:
This is just a test to assert the CI is doing good.


f0rmiga:
https://github.com/aspect-build/aspect-cli/runs/4742290600?check_suite_focus=true

PR #106, opened 2022-01-08
fix(docs): add docs for run command

mattem:
Fixes 404 on [aspect.build](https://aspect.build/aspect_run.md)


f0rmiga (approved):

f0rmiga:
We should fix our tests to assert this doesn't get missed again.

PR #112, opened 2022-01-11
feat: enforce generation of docs for commands

f0rmiga:
Whenever we create a new top-level command, the feature being introduced in this PR will keep the docs up-to-date.
This will avoid future mistakes e.g. the one that https://github.com/aspect-build/aspect-cli/commit/e4a9dd4fb3df30a4a197bc2fb06e42599c972331 fixes.

### DIFF
--- a/cmd/docgen/main.go
+++ b/cmd/docgen/main.go
@@ -60,14 +60,13 @@ func NewGenMarkdownCmd(aspectRootCmd *cobra.Command) *cobra.Command {
 		Use:   "gen-markdown",
 		Short: "Generates the markdown documentation.",
 		RunE: func(_ *cobra.Command, _ []string) error {
-			if outputDir == "" {
-				return fmt.Errorf("missing --output-dir")
-			}
 			return doc.GenMarkdownTree(aspectRootCmd, outputDir)
 		},
 	}
 
-	cmd.PersistentFlags().StringVar(&outputDir, "output-dir", "", "The path to the output directory.")
+	outputDirFlag := "output-dir"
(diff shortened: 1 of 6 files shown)


cmd/docgen/main.go
```diff
@@ -12,18 +13,61 @@ import (
...
+
+func NewGenMarkdownCmd(aspectRootCmd *cobra.Command) *cobra.Command {
+	var outputDir string
+
+	cmd := &cobra.Command{
+		Use:   "gen-markdown",
+		Short: "Generates the markdown documentation.",
+		RunE: func(_ *cobra.Command, _ []string) error {
+			if outputDir == "" {
+				return fmt.Errorf("missing --output-dir")
```
mattem, line 64: Doesn't cobra do this for you? As in, marking a flag as required?

mattem (approved):
Not going to block, but I feel the implementation / naming confuses me a little. As in, it outputs text that has doc and structure of a bzl file, but requires the output to be redirected to a file - what else would someone do with it?

Perhaps if it was named `generate_bzl_command_list` or something more specific? I dunno 🤷

cmd/docgen/main.go
f0rmiga: Right... I tried `MarkFlagRequired` and since it didn't work I thought 3 lines of code was just easier in this case. I took a look again and it wants `MarkPersistentFlagRequired`.

Fixed in 24ef379.

f0rmiga:
@mattem Yeah, makes sense the naming (fixed in 7f690ce). As for the command printing the bzl file instead of saving it as a file directly, I don't see any issue. Bash does this easier and the command is only used with a genrule anyways.

PR #116, opened 2022-01-16
feat: use repository cache for release builds

kormide:


kormide:
I suppose the only way to test this is to actually tag a release?

.github/workflows/release.yml
```diff
@@ -11,6 +11,18 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Bazel caches
+        uses: actions/cache@v2
+        with:
+          # Only restore the repository cache. Don't restore the build cache
+          # as stamped bazel builds won't rebuild targets when the workplace
+          # status variables change.
```
kormide, line 19: Is my assumption here correct? We always want to build a release build from scratch so that the artifacts will be properly stamped.

.github/workflows/release.yml
```diff
@@ -11,6 +11,18 @@ jobs:
...
+          path: |
+            ~/.cache/bazel-repo
+          key:
+            bazel-cache-${{ hashFiles('**/*.go', '**/BUILD.bazel', '**/*.bzl',
+            'WORKSPACE') }}
```
kormide, line 24: Prettier wanted to put this on a separate line. I hope that doesn't break the key..

f0rmiga: It's subtle with YAML. Since you used `>` to start the string, it will replace the new lines `\n` with spaces.

.github/workflows/release.yml
```diff
@@ -11,6 +11,16 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Bazel caches
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/bazel
+            ~/.cache/bazel-repo
+          key: >
```
f0rmiga, line 20:
```suggestion
          key: >-
```

f0rmiga:
> I suppose the only way to test this is to actually tag a release?

Please, do not release. My review is enough. We already validated the changes in the PR pipeline.

f0rmiga (approved):

PR #117, opened 2022-01-18
feat: aquery and cquery verbs

JesseTatasciore:
Add aquery and cquery verbs.

`aquery` and `cquery` can do basic passthroughs to bazel
Passthrough cquery and aquery

`aquery` and `cquery` can have predefined queries
Precanned cquery and aquery

`query` verb can list `aquery` and `cquery` predefined queries when configured to do so by the user
Query can list precanned aqueries and cqueries

`query` verb can be configured to use `cquery` under the hood
Query can use cquery


cmd/aspect/query/aquery/aquery.go
```diff
@@ -0,0 +1,31 @@
...
+Not licensed for re-use
+*/
+
+package aquery
+
+import (
+	"aspect.build/cli/pkg/aspect/query/aquery"
+	"aspect.build/cli/pkg/bazel"
+	"aspect.build/cli/pkg/ioutils"
+	"github.com/spf13/cobra"
```
f0rmiga, line 13:
```suggestion
	"github.com/spf13/cobra"

	"aspect.build/cli/pkg/aspect/query/aquery"
	"aspect.build/cli/pkg/bazel"
	"aspect.build/cli/pkg/ioutils"
```

cmd/aspect/query/cquery/cquery.go
```diff
@@ -0,0 +1,31 @@
...
+Not licensed for re-use
+*/
+
+package cquery
+
+import (
+	"aspect.build/cli/pkg/aspect/query/cquery"
+	"aspect.build/cli/pkg/bazel"
+	"aspect.build/cli/pkg/ioutils"
+	"github.com/spf13/cobra"
```
f0rmiga, line 13:
```suggestion
	"github.com/spf13/cobra"

	"aspect.build/cli/pkg/aspect/query/cquery"
	"aspect.build/cli/pkg/bazel"
	"aspect.build/cli/pkg/ioutils"
```

cmd/aspect/root/BUILD.bazel
```diff
@@ -13,7 +13,9 @@ go_library(
         "//cmd/aspect/clean",
         "//cmd/aspect/docs",
         "//cmd/aspect/info",
-        "//cmd/aspect/query",
+        "//cmd/aspect/query/aquery",
+        "//cmd/aspect/query/cquery",
+        "//cmd/aspect/query/query",
```
f0rmiga, line 18: I don't think they should live under `aspect/query/*`. They map to `aspect <cmd>`, so it makes more sense to have:

```
//cmd/aspect/aquery
//cmd/aspect/cquery
//cmd/aspect/query
```

f0rmiga (changes requested):

f0rmiga:
Did a first pass. Will do a second pass as soon as you move the files, so my review matches easier later.

cmd/aspect/root/BUILD.bazel
JesseTatasciore: Moved. Would you want the commands under `pkg/aspect/query` to be moved as well? Those ones are grouped because they share a lot of code

cmd/aspect/query/cquery/cquery.go
JesseTatasciore: Done. I still think we should have a linter that would catch these types of things

f0rmiga: You could try to use `goimports -local aspect.build/cli` to see if it does the right thing. If it does, feel free to add to https://github.com/aspect-build/aspect-cli/blob/main/.github/workflows/tidy_up_repository.sh.

cmd/aspect/root/BUILD.bazel
f0rmiga: I think it makes sense to move those too. I haven't checked yet how they share code but we can always shuffle the shared code around, if necessary.

JesseTatasciore: Moved. `pkg/aspect/aquery` and `pkg/aspect/cquery` have been created. Shared code currently lives under `pkg/aspect/query/shared`

pkg/aspect/query/query.go
```diff
@@ -8,154 +8,115 @@ package query
...
+		q.Prefs.Set(baseInquiredKey, true)
 
-func Select(presetNames []string) SelectRunner {
-	return &promptui.Select{
-		Label: "Select a preset query",
-		Items: presetNames,
-	}
-}
+		// if user types in y or Y then err will be nil. Any other input will not result in nil
+		_, someErr := q.Confirmation(question).Run()
```
f0rmiga, line 113:
```suggestion
		// Y = no error; N = error
		_, err := q.Confirmation(question).Run()
```

pkg/aspect/query/query_test.go
```diff
@@ -180,32 +256,46 @@ func TestQuery(t *testing.T) {
 			Streams:       streams,
 			Bzl:           spawner,
 			IsInteractive: true,
-			Prompt: func(label string) query.PromptRunner {
+			Prompt: func(label string) shared.PromptRunner {
+				fmt.Println("label:", label)
```
f0rmiga, line 260:
```suggestion
```

pkg/aspect/query/shared/query.go
```diff
@@ -0,0 +1,220 @@
...
+			ExitCode: exitCode,
+		}
+		return fmt.Errorf("failed to run %q %q: %w", verb, query, err)
+	}
+
+	return nil
+}
+
+func ReplacePlaceholders(query string, args []string, p func(label string) PromptRunner) (string, error) {
+	placeholders := regexp.MustCompile(`(\?[a-zA-Z]*)`).FindAllString(query, -1)
```
f0rmiga, line 152: Move the `MustCompile` to the package scope since this step is expensive and the result of the compiled regex can be considered a "const".

pkg/aspect/query/shared/query.go
```diff
@@ -0,0 +1,220 @@
...
+
+	if len(placeholders) == len(args)-1 {
+		for i, placeholder := range placeholders {
+			// todo.... Print out targetA was set to //foo and targetB was set to //bar
```
f0rmiga, line 156: Formal TODO? Issue?

f0rmiga (changes requested):

pkg/aspect/query/shared/query.go
JesseTatasciore: Just implemented instead.

f0rmiga (approved):

PR #118, opened 2022-01-19
feat(plugin): Add error augmentor plugin

JesseTatasciore:
Wire up SetupHook that will pass properties to their respective plugins

Add an error augmentor plugin that will parse bazel output and provide help
Error Augmentor

### DIFF
--- a/cmd/aspect/main.go
+++ b/cmd/aspect/main.go
@@ -38,17 +38,8 @@ func main() {
 	}
 
 	pluginSystem := system.NewPluginSystem()
-	if err := pluginSystem.ConfigurePluginSystem(ioutils.DefaultStreams); err != nil {
+	if err := pluginSystem.Configure(ioutils.DefaultStreams); err != nil {
 		fmt.Fprintln(os.Stderr, "Error:", err)
-		os.Exit(1)
-	}
-
-	setupErrors := pluginSystem.ConfigurePlugins().Errors()
-
-	if len(setupErrors) > 0 {
-		for _, err := range setupErrors {
-			fmt.Fprintln(os.Stderr, "Error:", err)
-		}
 		os.Exit(1)
 	}
 
--- a/cmd/aspect/main.go
+++ b/cmd/aspect/main.go
@@ -38,12 +38,12 @@ func main() {
 	}
 
 	pluginSystem := system.NewPluginSystem()
-	if err := pluginSystem.Configure(ioutils.DefaultStreams); err != nil {
+	if err := pluginSystem.ConfigurePluginSystem(ioutils.DefaultStreams); err != nil {
 		fmt.Fprintln(os.Stderr, "Error:", err)
 		os.Exit(1)
 	}
 
-	setupErrors := pluginSystem.SetupPlugins().Errors()
+	setupErrors := pluginSystem.ConfigurePlugins().Errors()
 
 	if len(setupErrors) > 0 {
 		for _, err := range setupErrors {
--- a/cmd/aspect/main.go
+++ b/cmd/aspect/main.go
@@ -42,6 +42,16 @@ func main() {
 		fmt.Fprintln(os.Stderr, "Error:", err)
 		os.Exit(1)
 	}
+
+	setupErrors := pluginSystem.SetupPlugins().Errors()
+
+	if len(setupErrors) > 0 {
+		for _, err := range setupErrors {
+			fmt.Fprintln(os.Stderr, "Error:", err)
+		}
+		os.Exit(1)
+	}
+
 	defer pluginSystem.TearDown()
 
 	cmd := root.NewDefaultRootCmd(pluginSystem)
--- a/pkg/plugin/sdk/v1alpha2/plugin/grpc.go
+++ b/pkg/plugin/sdk/v1alpha2/plugin/grpc.go
@@ -60,8 +60,7 @@ func (m *GRPCServer) SetupHook(
 	ctx context.Context,
 	req *proto.SetupHookReq,
 ) (*proto.SetupHookRes, error) {
-	return &proto.SetupHookRes{},
-		m.Impl.SetupHook(req.Properties)
+	return &proto.SetupHookRes{}, m.Impl.SetupHook(req.Properties)
 }
 
 // PostBuildHook translates the gRPC call to the Plugin PostBuildHook
--- a/pkg/plugin/sdk/v1alpha2/plugin/grpc.go
+++ b/pkg/plugin/sdk/v1alpha2/plugin/grpc.go
@@ -136,7 +136,7 @@ func (m *GRPCClient) BEPEventCallback(event *buildeventstream.BuildEvent) error 
 
 // SetupHook is called from the Core to execute the Plugin SetupHook.
 func (m *GRPCClient) SetupHook(
-	properties string,
+	properties []byte,
 ) error {
 	req := &proto.SetupHookReq{
 		Properties: properties,
--- a/pkg/plugin/sdk/v1alpha2/plugin/grpc.go
+++ b/pkg/plugin/sdk/v1alpha2/plugin/grpc.go
@@ -55,12 +55,12 @@ func (m *GRPCServer) BEPEventCallback(
 	return &proto.BEPEventCallbackRes{}, m.Impl.BEPEventCallback(req.Event)
 }
 
-// SetupHook translates the gRPC call to the Plugin SetupHook implementation.
-func (m *GRPCServer) SetupHook(
+// Setup translates the gRPC call to the Plugin Setup implementation.
+func (m *GRPCServer) Setup(
 	ctx context.Context,
-	req *proto.SetupHookReq,
-) (*proto.SetupHookRes, error) {
-	return &proto.SetupHookRes{}, m.Impl.SetupHook(req.Properties)
+	req *proto.SetupReq,
+) (*proto.SetupRes, error) {
+	return &proto.SetupRes{}, m.Impl.Setup(req.Properties)
 }
 
 // PostBuildHook translates the gRPC call to the Plugin PostBuildHook
@@ -134,14 +134,14 @@ func (m *GRPCClient) BEPEventCallback(event *buildeventstream.BuildEvent) error 
 	return err
 }
 
-// SetupHook is called from the Core to execute the Plugin SetupHook.
-func (m *GRPCClient) SetupHook(
+// Setup is called from the Core to execute the Plugin Setup.
+func (m *GRPCClient) Setup(
 	properties []byte,
 ) error {
-	req := &proto.SetupHookReq{
+	req := &proto.SetupReq{
 		Properties: properties,
 	}
-	_, err := m.client.SetupHook(context.Background(), req)
+	_, err := m.client.Setup(context.Background(), req)
 	return err
 }
 
--- a/pkg/plugin/sdk/v1alpha2/plugin/grpc.go
+++ b/pkg/plugin/sdk/v1alpha2/plugin/grpc.go
@@ -55,6 +55,15 @@ func (m *GRPCServer) BEPEventCallback(
 	return &proto.BEPEventCallbackRes{}, m.Impl.BEPEventCallback(req.Event)
 }
 
+// SetupHook translates the gRPC call to the Plugin SetupHook implementation.
+func (m *GRPCServer) SetupHook(
+	ctx context.Context,
+	req *proto.SetupHookReq,
+) (*proto.SetupHookRes, error) {
+	return &proto.SetupHookRes{},
+		m.Impl.SetupHook(req.Properties)
+}
+
 // PostBuildHook translates the gRPC call to the Plugin PostBuildHook
 // implementation. It starts a prompt runner that is passed to the Plugin
 // instance to be able to perform prompt actions to the CLI user.
@@ -123,6 +132,17 @@ type GRPCClient struct {
 // BEPEventCallback.
 func (m *GRPCClient) BEPEventCallback(event *buildeventstream.BuildEvent) error {
 	_, err := m.client.BEPEventCallback(context.Background(), &proto.BEPEventCallbackReq{Event: event})
+	return err
+}
+
+// SetupHook is called from the Core to execute the Plugin SetupHook.
+func (m *GRPCClient) SetupHook(
+	properties string,
+) error {
+	req := &proto.SetupHookReq{
+		Properties: properties,
+	}
+	_, err := m.client.SetupHook(context.Background(), req)
 	return err
 }
 
--- a/pkg/plugin/sdk/v1alpha2/plugin/interface.go
+++ b/pkg/plugin/sdk/v1alpha2/plugin/interface.go
@@ -14,7 +14,7 @@ import (
 // Plugin determines how an aspect Plugin should be implemented.
 type Plugin interface {
 	BEPEventCallback(event *buildeventstream.BuildEvent) error
-	SetupHook(properties string) error
+	SetupHook(properties []byte) error
 	PostBuildHook(
 		isInteractiveMode bool,
 		promptRunner ioutils.PromptRunner,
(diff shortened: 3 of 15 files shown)


plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+
+		// if aborted then there will not be a progress messages so exit early
+		return nil
+	}
+
+	progress := event.GetProgress()
+
+	if progress != nil {
+		stderr := progress.GetStderr()
+		plugin.processErrorMessage(stderr)
```
f0rmiga, line 83: I think you want at least to check if `stderr` is not empty before calling `processErrorMessage`. That method is quite expensive to run on every event coming from Bazel.

Ideally, we should be able to do a check on the event to see if it represents an error. Could you check this? If not, take my following suggestion:

```suggestion
		if stderr != "" {
			plugin.processErrorMessage(stderr)
		}
```

f0rmiga: You could also spawn a goroutine as the error message processor and communicate via non-blocking channels, so Bazel doesn't wait for you to do all the regex matching before it proceeds. I can help pairing with you on this.

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+				str = strings.ReplaceAll(str, fmt.Sprint("$", i), match)
+			}
+
+			hintExists := false
+			for _, existingHint := range plugin.helpfulHints {
+				if existingHint == str {
+					hintExists = true
+					break
+				}
+			}
```
f0rmiga, line 109: Yet another loop... getting too expensive here. How could we eliminate this?

Seems like you are trying to represent a set. I bet I already implemented an ordered set data structure somewhere in this repository. If not, I can pair with you again to implement it. Proper data structures make the code cleaner and more efficient.

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	for _, hint := range plugin.helpfulHints {
+		plugin.printMiddle("- " + hint)
+	}
+
+	plugin.printBreak()
+	return nil
+}
+
+func (plugin *ErrorAugmentorPlugin) printBreak() {
+	var b strings.Builder
```
f0rmiga, line 140: I guess you can just print everything straight to stdout instead of creating a buffer.

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+		fmt.Fprintf(&b, "-")
+	}
+
+	fmt.Fprintf(&b, " ")
+
+	fmt.Println(b.String())
+}
+
+func (plugin *ErrorAugmentorPlugin) printMiddle(str string) {
+	var b strings.Builder
```
f0rmiga, line 154: Same here. Unnecessary buffer.

f0rmiga (changes requested):

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
```diff
@@ -55,6 +55,15 @@ func (m *GRPCServer) BEPEventCallback(
 	return &proto.BEPEventCallbackRes{}, m.Impl.BEPEventCallback(req.Event)
 }
 
+// SetupHook translates the gRPC call to the Plugin SetupHook implementation.
+func (m *GRPCServer) SetupHook(
+	ctx context.Context,
+	req *proto.SetupHookReq,
+) (*proto.SetupHookRes, error) {
+	return &proto.SetupHookRes{},
+		m.Impl.SetupHook(req.Properties)
```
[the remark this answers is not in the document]
JesseTatasciore: Will fix it. A large number of the functions in this file are laid out like that. I had just copied from one of them

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	"aspect.build/cli/pkg/plugin/sdk/v1alpha2/config"
+)
+
+func main() {
+	goplugin.Serve(config.NewConfigFor(NewDefaultPlugin()))
+}
+
+type ErrorAugmentorPlugin struct {
+	properties          string
+	regex               map[*regexp.Regexp]string
```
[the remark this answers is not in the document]
JesseTatasciore: My variable naming sucks. Going with hintMap

cmd/aspect/main.go
```diff
@@ -42,6 +42,16 @@ func main() {
 		fmt.Fprintln(os.Stderr, "Error:", err)
 		os.Exit(1)
 	}
+
+	setupErrors := pluginSystem.SetupPlugins().Errors()
```
f0rmiga, line 46: From the perspective of the main command, both `Configure` and `SetupPlugins` sound like they are doing the same thing. We can perhaps combine them all in a single method? Or maybe give different names to those methods so it's clearer what does what?

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -14,6 +14,7 @@ import (
 // Plugin determines how an aspect Plugin should be implemented.
 type Plugin interface {
 	BEPEventCallback(event *buildeventstream.BuildEvent) error
+	SetupHook(properties []byte) error
```
f0rmiga, line 17: I don't think it's hooking onto anything. IMO, `Setup` represents the intention better.

pkg/plugin/system/aspectplugins.go
```diff
@@ -24,10 +24,11 @@ const (
...
 type AspectPlugin struct {
-	Name       string                 `yaml:"name"`
-	From       string                 `yaml:"from"`
-	LogLevel   string                 `yaml:"log_level"`
-	Properties map[string]interface{} `yaml:"properties"`
+	Name          string                 `yaml:"name"`
+	From          string                 `yaml:"from"`
+	LogLevel      string                 `yaml:"log_level"`
+	PropertiesMap map[string]interface{} `yaml:"properties"`
+	properties    []byte
```
f0rmiga, line 31: How about this? At least they match YAML -> struct field.
```suggestion
	Properties map[string]interface{} `yaml:"properties"`
	propertiesBytes    []byte
```

pkg/plugin/system/aspectplugins.go
```diff
@@ -114,5 +117,16 @@ func (p *parser) Parse(aspectpluginsPath string) ([]AspectPlugin, error) {
...
 	}
-	return aspectplugins, nil
+
+	var processedPlugins []AspectPlugin
+	for _, plugin := range aspectplugins {
+		marshalledProperties, err := p.yamlMarshal(plugin.PropertiesMap)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse .aspectplugins: %w", err)
+		}
+		plugin.properties = marshalledProperties
```
f0rmiga, line 127: Continuing the suggestion above, see how it looks clearer?
```suggestion
		propertiesBytes, err := p.yamlMarshal(plugin.Properties)
		if err != nil {
			return nil, fmt.Errorf("failed to parse .aspectplugins: %w", err)
		}
		plugin.propertiesBytes = propertiesBytes
```

pkg/plugin/system/aspectplugins.go
f0rmiga: Alternatively, `propertiesRaw`?

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,231 @@
...
+func NewPlugin() *ErrorAugmentorPlugin {
+	return &ErrorAugmentorPlugin{
+		properties:          nil,
+		hintMap:             map[*regexp.Regexp]string{},
+		yamlUnmarshalStrict: yaml.UnmarshalStrict,
+		helpfulHints:        &helpfulHintSet{nodes: make(map[helpfulHintNode]struct{})},
+	}
+}
+
+// will be used to unmarshal plugin properties specific to this plugin
```
f0rmiga, line 50: Nit: prefer comments with capitalization and punctuation. I.e. more formal grammar.

```suggestion
// errorMappings is used to unmarshal the error mappings from the plugin properties provided by the CLI core.
```

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (plugin *ErrorAugmentorPlugin) SetupHook(
+	properties []byte,
+) error {
+	plugin.properties = properties
+
+	var processedProperties errorMappings
+	if err := plugin.yamlUnmarshalStrict(properties, &processedProperties); err != nil {
+		return fmt.Errorf("failed to setup: failed to parse properties: %w", err)
```
f0rmiga, line 62: +1 - perfect error wrapping message with the first part being the current context, the second part an additional context from the generic unmarshal function, and the actual error from the call.

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,231 @@
...
+	// change map keys into regex objects now so they are ready to use and we only need to compile the regex once
+	for r, m := range processedProperties.ErrorMappings {
+		// r for regex, m for message
+		regex, err := regexp.Compile(r)
+		if err != nil {
+			return err
+		}
+
+		plugin.hintMap[regex] = m
+	}
```
f0rmiga, line 74:
```suggestion
	for pattern, message := range processedProperties.ErrorMappings {
		regex, err := regexp.Compile(pattern)
		if err != nil {
			return err
		}

		plugin.hintMap[regex] = message
	}
```

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,231 @@
...
+
+	plugin.errorMessages = make(chan string, 64)
+	for i := 0; i < 4; i++ {
```
f0rmiga, line 77: I think you could make this `4` a Plugin property. And the channel buffer size a multiple of the property. Something like `processor_count`? Not sure...

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,231 @@
...
+		plugin.errorMessages <- aborted.Description
+
+		// We exit early here because there will not be a progress message when the event was of type "aborted".
+		return nil
+	}
+
+	progress := event.GetProgress()
+
+	if progress != nil {
+		// TODO: Should we also check for stdout?
```
f0rmiga, line 97: Did you figure this one out at some point?

plugins/error-augmentor/plugin.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (plugin *ErrorAugmentorPlugin) processErrorMessage(errorMessage string) {
+	for regex, helpfulHint := range plugin.hintMap {
+		matches := regex.FindStringSubmatch(errorMessage)
+
+		if len(matches) > 0 {
+
+			// apply regex capture group replacements to given hint
+			for i, match := range matches {
+				if i == 0 {
```
f0rmiga, line 122: Why are we skipping index 0 again? I don't recall. It's worth a comment?

f0rmiga: Oh, reading the docs from `FindStringSubmatch` refreshed my memory. I still think a comment helps clarify.

cmd/aspect/main.go
JesseTatasciore: `Configure` configures the plugin system. `SetupPlugins` configures the plugins themselves. I think they deserve to be separated. `ConfigurePluginSystem` and `ConfigurePlugins` might make the differences more clear.

pkg/plugin/system/aspectplugins.go
JesseTatasciore: I like `propertiesBytes`

pkg/plugin/system/aspectplugins.go
JesseTatasciore: Yes it does look cleaner like this. I agree

plugins/error-augmentor/plugin.go
JesseTatasciore: :)

plugins/error-augmentor/plugin.go
JesseTatasciore: Done

plugins/error-augmentor/plugin.go
JesseTatasciore: I spent a little time but didnt find much. I think we should. Ive added it

pkg/plugin/system/system.go
```diff
@@ -30,10 +30,11 @@ import (
 // PluginSystem is the interface that defines all the methods for the aspect CLI
 // plugin system intended to be used by the Core.
 type PluginSystem interface {
-	Configure(streams ioutils.Streams) error
+	ConfigurePluginSystem(streams ioutils.Streams) error
```
jbedard, line 33: Do we need the word "Plugin" or "System" Within the "Plugin System" class?

pkg/plugin/system/system.go
```diff
@@ -30,10 +30,11 @@ import (
...
 	TearDown()
 	BESBackendInterceptor() interceptors.Interceptor
 	BuildHooksInterceptor(streams ioutils.Streams) interceptors.Interceptor
+	ConfigurePlugins() *aspecterrors.ErrorList
```
jbedard, line 37: Do we need 2 methods? Do we need to return all errors instead of just the first one?

pkg/plugin/system/system.go
```diff
@@ -236,6 +251,7 @@ func (l *PluginList) insert(p plugin.Plugin) {
 
 // PluginNode is a node in the PluginList linked list.
 type PluginNode struct {
-	next   *PluginNode
-	plugin plugin.Plugin
+	next       *PluginNode
+	plugin     plugin.Plugin
+	properties []byte
```
jbedard, line 256: @f0rmiga do we want to be adding more then just `next` + `plugin` to linked-lists like this?

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -14,6 +14,7 @@ import (
 // Plugin determines how an aspect Plugin should be implemented.
 type Plugin interface {
 	BEPEventCallback(event *buildeventstream.BuildEvent) error
+	Setup(properties []byte) error
```
jbedard, line 17: What do the bytes represent? Is it weird having bytes in the public API?

pkg/plugin/system/system.go
f0rmiga: I had the same comment on the first review. Every time I look at this, I'm more convinced that a single method is what we want. I'd say that `Configure` should do both things and return a list of errors. I don't see why not return all errors possible.

pkg/plugin/sdk/v1alpha2/plugin/interface.go
f0rmiga: This is public. I have been thinking about this since yesterday, and I could be convinced to change the API here to take a `map[string]interface{}`. The disadvantage is that we limit the users from parsing the bytes straight into a typed struct.

pkg/plugin/system/system.go
f0rmiga: Conceptually, I don't see why not. In this case, both `plugin` and `properties` form the payload of a list node. For clarity, it could be expressed like:

```go
type PluginNode struct {
    next *PluginNode
    payload struct {
        plugin plugin.Plugin
        properties []byte
    }
}
```

pkg/plugin/system/system.go
JesseTatasciore: Just one `Configure` method again

pkg/plugin/system/system.go
JesseTatasciore: Done

pkg/plugin/sdk/v1alpha2/plugin/interface.go
JesseTatasciore: I do agree that its a little weird to have `[]byte` on a public API like this. However it is the most convenient form for that data to be in. Makes it super easy for the plugin to parse it into its expected type. If we pass it through as something else then most plugins are just going to need to unmarshal it again creating extra work on their end

PR #119, opened 2022-01-19
bump: plugin protocol

f0rmiga:
We missed bumping the protocol version.


alexeagle (approved):
I'm still curious what we can do to verify our versioning is right so it's not a "remember to"
maybe when I get back into this code some more :)

f0rmiga:
Agree. We need a better story. Feel free to dive into it. If you don't get to it, I will at some point. I just think that our other tasks are more important at this time - and I may be wrong.

f0rmiga:
Created https://github.com/aspect-build/aspect-cli/issues/120 for this.

PR #121, opened 2022-01-20
build: move ci cache flags to bazelrc

kormide:


.prettierrc
```diff
@@ -1 +0,0 @@
-proseWrap: always
```
kormide, line 1: @alexeagle Was this option intentionally added or just copied from another file? If we include this, then prettier likes to wrap yaml keys without adding the proper `>-` directives.

alexeagle: it's copy-pasta - it's useful for markdown though.

alexeagle (approved):

PR #122, opened 2022-01-24
feat: Allow predefined queries to be specified in aspect configuration

JesseTatasciore:
predefined queries in config


integration_tests/query/long_preset_query_test.sh
```diff
@@ -0,0 +1,32 @@
...
+ASPECT="$TEST_SRCDIR/build_aspect_cli/cmd/aspect/aspect_/aspect"
+export HOME
+touch WORKSPACE
+
+mkdir foo
+echo "genrule(" > foo/BUILD
+echo "    name = \"foo\"," >> foo/BUILD
+echo "    outs = [\"foo.txt\"]," >> foo/BUILD
+echo "    cmd = \"touch \$@\"," >> foo/BUILD
+echo ")" >> foo/BUILD
```
f0rmiga, line 15:
```suggestion
cat > foo/BUILD <<'EOF'
genrule(
    name = "foo",
    outs = ["foo.txt"],
    cmd = "touch $@",
)
EOF
```

f0rmiga: You can do the same for the other `echo` lines.

f0rmiga (approved):

PR #124, opened 2022-01-27
feat: Add the ability to cleanup all (or at least many) bazel workspaces

JesseTatasciore:


pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,457 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+	// Goroutines for calculating sizes of directories
+	for i := 0; i < 5; i++ {
+		go c.sizeCalculator(sizeCalcQueue, confirmationQueue, &sizeCalcWaitGroup)
+		sizeCalcWaitGroup.Add(1)
+	}
+
+	// Goroutines for deleting directories.
+	for i := 0; i < 8; i++ {
+		go c.deleteProcessor(deleteQueue, &deleteWaitGroup, sizeQueue, errorQueue)
+	}
```
alexeagle, line 221: this isn't symmetric with the loop right before - doesn't it also need to add to the deleteWaitGroup?

or do the deletes only start filling in when the user has been prompted below? then maybe a comment would be useful here about why it's the only place we start goroutines but then don't add to a WaitGroup

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,457 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+	defer os.RemoveAll(tempDir)
+
+	bepLocation := filepath.Join(tempDir, "bep.json")
+
+	// Running an invalid query should ensure that repository rules are not executed.
+	// However, bazel will still emit its BEP containing the flag that we are interested in.
+	// This will ensure it returns quickly and allows us to easily access said flag.
+	c.bzl.RunCommand([]string{
+		"query",
+		"//",
```
alexeagle, line 332: that's a kinda crazy workaround. did you try `bazel info`, `bazel version` etc

also if I only specify `build --disk_cache=something` in my bazelrc, then `bazel query` won't have that flag applied right?

pkg/aspect/clean/windows.go
```diff
@@ -0,0 +1,35 @@
...
+
+	return smallestTime
+}
+
+func (c *Clean) MoveFolderToTmp(dir string, name string) string {
+	return ""
+}
+
+func (c *Clean) ChangeFolderPermissions(folder string) ([]byte, error) {
+	return nil, nil
```
alexeagle, line 34: is this still TODO?

pkg/aspect/clean/clean.go
JesseTatasciore: `deleteWaitGroup` is used slightly differently than the other wait groups which is why its not added here. I have added a comment

pkg/aspect/clean/windows.go
JesseTatasciore: Yes it is still a TODO. Should I add a comment or make a ticket?

pkg/aspect/clean/clean.go
JesseTatasciore: `bazel info` and `bazel version` do not spit out BES events. Best guess as to why, is that the server actually needs to start and info / version dont turn it on far enough? Maybe? Not sure. I agree that running the query like this is a bit crazy but its quick (doesnt run repo rules / installs) and it spits out what we need.

Even though we are running a query, BES will spit out all of the flags. Including the ones for build / test. We get an `unstructuredCommandLine` section witch contains useful lines such as:
```
"--default_override\u003d2:build\u003d--disk_cache\u003d/Users/jesse/Development/bazel_cache",
```

We can ignore the verb its tagged to and just use some regex to look for the path after `--disk_cache`

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+		"--build_event_json_file=" + bepLocation,
+
+		// We dont want bazel to print anything to the command line.
+		// We are only interested in the BEP output
+		"--ui_event_filters=-fatal,-error,-warning,-info,-progress,-debug,-start,-finish,-subcommand,-stdout,-stderr,-pass,-fail,-timeout,-cancelled,-depchecker",
```
alexeagle, line 341: why bother trying to give a correct and comprehensive set of flags to turn off bazel output, if we throw all the output away anyway? it feels redundant and harder to maintain than just `io.Discard` with no other flags

pkg/aspect/clean/windows.go
alexeagle: yes, ideally both a ticket and a TODO here pointing to it. What's the behavior right now if you run on Windows? It should either not offer the new features at all (behave the same as `bazel clean`), or throw an error that it doesn't work. The worst case is for it to silently be broken.

pkg/aspect/clean/linux.go
```diff
@@ -0,0 +1,46 @@
+//go:build linux
```
alexeagle, line 1: I guess this comment tells the `go build` tool to include a different file depending on the host platform? does this function under Bazel at all?

pkg/aspect/clean/darwin.go
```diff
@@ -0,0 +1,46 @@
...
+		smallestTime = timeSinceCreation
+	} else if timeSinceModified < timeSinceAccess && timeSinceModified < timeSinceCreation {
+		smallestTime = timeSinceModified
+	}
+
+	return smallestTime
+}
+
+func (c *Clean) MoveFolderToTmp(dir string, name string) string {
+	newFolder := "/private/var/tmp/aspect_delete/" + strings.Replace(dir, "/", "", -1)
```
alexeagle, line 35: this feels fishy, I think you want to use https://pkg.go.dev/os#TempDir rather than hard-code the path

pkg/aspect/clean/darwin.go
```diff
@@ -0,0 +1,46 @@
...
+	newPath := newFolder + "/" + name
+	os.MkdirAll(newFolder, os.ModePerm)
+	os.Rename(dir+"/external/"+name, newPath)
+
+	return newPath
+}
+
+func (c *Clean) ChangeFolderPermissions(folder string) ([]byte, error) {
+	cmd := exec.Command("chmod", "-R", "777", folder)
```
alexeagle, line 44: use the `os` package rather than spawn a subprocess, for better portability https://pkg.go.dev/os#Chmod

alexeagle:
seems like it just gets stuck in a loop if I say No

```
% ./bazel-bin/cmd/aspect/aspect_/aspect clean
✔ Reclaim disk space for all Bazel workspaces
? Workspace: Unknown Workspace, Age: 37h54m10.254816344s, Size: 356.52 KB. Would you like to remove?? [y/N]
? Workspace: Unknown Workspace, Age: 37h54m10.254816344s, Size: 356.52 KB. Would you like to remove?? [y/N]
? Workspace: Unknown Workspace, Age: 37h54m10.254816344s, Size: 356.52 KB. Would you like to remove?? [y/N]
? Workspace: Unknown Workspace, Age: 37h54m10.254816344s, Size: 356.52 KB. Would you like to remove?? [y/N]
? Workspace: Unknown Workspace, Age: 37h54m10.254816344s, Size: 356.52 KB. Would you like to remove?? [y/N]
? Workspace: Unknown Workspace, Age: 37h54m10.254816344s, Size: 356.52 KB. Would you like to remove?? [y/N]
? Workspace: Unknown Workspace, Age: 37h54m10.254816344s, Size: 356.52 KB. Would you like to remove?? [y/N]
```

pkg/aspect/clean/clean.go
JesseTatasciore: The `io.Discard` does not capture all of the output from the bazel server. If I leave the bazel output on I will still get the following printed to my terminal:
```
ERROR: the empty string is not a valid target
INFO: Build Event Protocol files produced successfully.
```
Even though everything should be discarded.

pkg/aspect/clean/linux.go
JesseTatasciore: Correct. This tells `go build` to include this file when building the binary for `linux`.

What do you mean by "Does this function under bazel at all?". At the moment it appears to be working. `//release:aspect` which builds the binaries for darwin / linux / windows appears to be picking up the correct files as long as I can tell. And I can confirm that the darwin one is being picked up for my local build. Have not conclusively tested for windows and linux.

pkg/aspect/clean/darwin.go
JesseTatasciore: My reasoning for doing it this way was that there is no recursive mode for https://pkg.go.dev/os#Chmod so I will need to walk the file tree in order to change everything to have deletable permissions. But I can definitely switch. Thats not a problem. Thoughts?

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+			path:               filepath.Join(bazelBaseDir, workspace.Name()),
+			isCurrentWorkspace: workspace.Name() == currentWorkingBase,
+			isCache:            false,
+		}
+
+		workspaceInfo.accessTime = c.GetAccessTime(workspace)
+
+		execrootFiles, readDirErr := ioutil.ReadDir(filepath.Join(bazelBaseDir, workspace.Name(), "execroot"))
+		if readDirErr != nil {
+			// install and cache folders will end up here. Dont want to remove these
```
f0rmiga, line 414:
```suggestion
			// The install and cache folders will end up here. We must not remove these.
```

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+			continue
+		}
+
+		// We expect these folders to have up to 2 files / folders
+		//   - Firstly, a file named "DO_NOT_BUILD_HERE"
+		//   - Secondly, a folder named after the given workspace
```
f0rmiga, line 420:
```suggestion
		// We expect these folders to have up to 2 files / folders:
		//   - Firstly, a file named "DO_NOT_BUILD_HERE".
		//   - Secondly, a folder named after the given workspace.
```

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+		// We can use the given second folder to determine the name of the workspace. We want this
+		// so that we can ask the user if they want to remove a given workspace
```
f0rmiga, line 422:
```suggestion
		// so that we can ask the user if they want to remove a given workspace.
```

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+		if (len(execrootFiles) == 1 && execrootFiles[0].Name() == "DO_NOT_BUILD_HERE") || len(execrootFiles) > 2 {
+			// TODO: Group up unkown workspaces into one prompt?
```
f0rmiga, line 424: Make this an issue and link here?

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+		// therefore happen in parallel.
+		externalFolders, _ := ioutil.ReadDir(filepath.Join(bazelDir.path, "external"))
+		for _, folder := range externalFolders {
+			newPath := c.MoveFolderToTmp(bazelDir.path, folder.Name())
+
+			if newPath != "" {
+				waitGroup.Add(1)
+				deleteQueue <- bazelDirInfo{
+					path: newPath,
+					size: 0, // dont want to calculate the size for the subfolders. Just include the parent size
```
f0rmiga, line 487: The comment is not very clear. Could you improve it, please?

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+				}
+			}
+		}
+
+		// otherwise certain files will throw a permission error when we try to remove them
```
f0rmiga, line 492:
```suggestion
		// The permissions set in the directories being removed don't allow write access,
		// so we change the permissions before removing those directories.
```

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+		if _, err := c.ChangeFolderPermissions(bazelDir.path); err != nil {
+			waitGroup.Done()
+			errors <- fmt.Errorf("failed to delete %q: %w", bazelDir.path, err)
+			continue
+		}
+
+		// remove entire folder
```
f0rmiga, line 499:
```suggestion
		// Remove the entire directory tree.
```

f0rmiga: I see we have a mix of references to directories versus folders. Could we pick one and use it everywhere?

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+
+func (c *Clean) sizePrinter(sizeQueue <-chan float64, waitGroup *sync.WaitGroup) {
+	var totalSize float64 = 0
+
+	for size := range sizeQueue {
+		totalSize = totalSize + size
+	}
+
+	_, hRSpaceReclaimed, unit := c.makeBytesHumanReadable(totalSize)
+	fmt.Printf("Space reclaimed: %.2f%s\n", hRSpaceReclaimed, unit)
```
f0rmiga, line 527: Any time we print something to the console, we should use `fmt.Fprintf` with the streams injected into the object by the cobra command constructor. This way, we can properly unit test these methods.

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+	}
+
+	files, err := ioutil.ReadDir(cwd)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to find Bazel base dir: %w", err)
+	}
+
+	for _, file := range files {
+
+		// bazel-bin, bazel-out, etc... will be symlinks, so we can eliminate non-symlinks immediately
```
f0rmiga, line 545:
```suggestion
		// bazel-bin, bazel-out, etc... will be symlinks, so we can eliminate non-symlinks immediately.
```

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+		if file.Mode()&os.ModeSymlink != 0 {
+			actualPath, err := os.Readlink(filepath.Join(cwd, file.Name()))
+			if err != nil {
+				return "", "", fmt.Errorf("failed to find Bazel base dir: %w", err)
+			}
+
+			if strings.Contains(actualPath, "bazel") && strings.Contains(actualPath, "/execroot/") {
```
f0rmiga, line 552: If we end up supporting Windows, not sure this is a good thing.
You may want to use https://pkg.go.dev/path#Split and then check if one of the parts of the path matches `execroot`.

pkg/aspect/clean/darwin.go
f0rmiga: Since this is a specialized code for macOS, I'm fine with the subprocess. The `chmod` command is already optimized enough and with lots of baggage for doing this job correctly.

pkg/aspect/clean/linux.go
f0rmiga: We need automated tests for this, so we can request CI to do the integration testing for multiple platforms.

f0rmiga (changes requested):

pkg/aspect/clean/clean.go
```diff
@@ -165,3 +189,461 @@ func (c *Clean) Run(_ *cobra.Command, _ []string) error {
...
+		}
+	}
+	confirmationWaitGroup.Done()
+}
+
+func (c *Clean) findDiskCaches(
+	sizeCalcQueue chan<- bazelDirInfo,
+	errors chan<- error,
+) {
+	tempDir, err := ioutil.TempDir("", "tmp_bazel_output")
```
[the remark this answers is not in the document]
JesseTatasciore: https://github.com/aspect-build/aspect-cli/issues/210

pkg/aspect/clean/clean.go
JesseTatasciore: `path.Split` does not do this split. It simply splits the the directory from the filename. I have instead changed my approach to use `os.PathSeparator`. Hopefully that addresses your concerns?

JesseTatasciore:
Related tickets for completing some of the missing work:
https://github.com/aspect-build/aspect-cli/issues/210
https://github.com/aspect-build/aspect-cli/issues/211

f0rmiga (approved):

PR #127, opened 2022-01-28
fix: ensure the workspaceRoot is set before running Bazel

alexeagle:
Otherwise some of our commands don't tell bazelisk how to find the .bazelversion file, and we run the wrong Bazel.
This results in churning the Bazel server when running these aspect commands, and is clearly wrong for users as well.

### DIFF
--- a/pkg/bazel/bazel.go
+++ b/pkg/bazel/bazel.go
@@ -51,6 +51,10 @@ func (b *bazel) Spawn(command []string) (int, error) {
 
 func (b *bazel) RunCommand(command []string, out io.Writer) (int, error) {
 	repos := b.createRepositories()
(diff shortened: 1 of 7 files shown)


pkg/bazel/bazel.go
```diff
@@ -51,6 +51,10 @@ func (b *bazel) Spawn(command []string) (int, error) {
 
 func (b *bazel) RunCommand(command []string, out io.Writer) (int, error) {
 	repos := b.createRepositories()
+	if len(b.workspaceRoot) < 1 {
+		panic("Illegal state: running bazel without the workspaceRoot set")
```
f0rmiga, line 55: 👍

f0rmiga (approved):

PR #132, opened 2022-02-03
build: allow PRs from forks

jbedard:


.github/workflows/validate_changed_files.sh
```diff
@@ -7,10 +7,8 @@ VALID_FILE_PATHS=('pkg/plugin/sdk/v1alpha2') # Array of filepaths that are allow
 
 # by default the only local branch will be pull/PR#/merge
 # fetch only the latest commit from the 2 branches in question to avoid fetching the entire repo which could be costly
-git fetch --depth 1 origin "$1"
-git fetch --depth 1 origin "$2"
 
-git diff "origin/$1..origin/$2" --name-only | while read -r file; do
+for file in "$@"; do
```
jbedard, line 12: fwiw if I put `echo "file: $file"` here it does correctly loop over each file in the commit

jbedard: this is the same as the PR but with the echo: https://github.com/aspect-build/aspect-cli/runs/5057836349

.github/workflows/default.yml
```diff
@@ -13,10 +13,16 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v2
+      - id: files
+        name: Get changed files
+        uses: jitterbit/get-changed-files@v1
```
f0rmiga, line 20: I'd avoid bringing an external Action for something that's easily solved with git.

f0rmiga (changes requested):

.github/workflows/default.yml
jbedard: I was thinking the same but I don't think it is easily solved with the information the git workflows give you. Look at the implementation of some of these things:

https://github.com/tj-actions/changed-files/blob/main/entrypoint.sh
https://github.com/lots0logs/gh-action-get-changed-files/blob/master/index.js
https://github.com/jitterbit/get-changed-files/blob/master/src/main.ts

jbedard: We could also always change it later. Doing this for now doesn't block us from making changes in the future etc.

This thing seems well used and we version lock it so I think it's stable for now. Unless we can figure it out with a 1-2 line git command I don't want to waste anymore time on this.

.github/workflows/validate_changed_files.sh
```diff
@@ -7,10 +7,14 @@ VALID_FILE_PATHS=('pkg/plugin/sdk/v1alpha2') # Array of filepaths that are allow
...
-git fetch --depth 1 origin "$2"
 
-git diff "origin/$1..origin/$2" --name-only | while read -r file; do
+git fetch --depth 1 origin $GITHUB_BASE_REF
+git fetch --depth 1 origin $GITHUB_REF
+
+main_head=`git rev-parse --short origin/$GITHUB_BASE_REF`
+pr_head=$GITHUB_SHA
+
+git diff --name-only ${main_head}..${pr_head} | while read -r file; do
```
f0rmiga, line 17: In this case, it doesn't matter, but to make a habit, always wrap Bash variables with double quotes to avoid word splitting.
```suggestion
git fetch --depth 1 origin "${GITHUB_BASE_REF}"
git fetch --depth 1 origin "${GITHUB_REF}"

main_head=$(git rev-parse --short "origin/${GITHUB_BASE_REF}")
pr_head="${GITHUB_SHA}"

git diff --name-only "${main_head}..${pr_head}" | while read -r file; do
```

.github/workflows/validate_changed_files.sh
```diff
@@ -7,10 +7,14 @@ VALID_FILE_PATHS=('pkg/plugin/sdk/v1alpha2') # Array of filepaths that are allow
...
 # by default the only local branch will be pull/PR#/merge
 # fetch only the latest commit from the 2 branches in question to avoid fetching the entire repo which could be costly
-git fetch --depth 1 origin "$1"
-git fetch --depth 1 origin "$2"
 
-git diff "origin/$1..origin/$2" --name-only | while read -r file; do
+git fetch --depth 1 origin $GITHUB_BASE_REF
+git fetch --depth 1 origin $GITHUB_REF
+
+main_head=`git rev-parse --short origin/$GITHUB_BASE_REF`
```
f0rmiga, line 14: Nit: it's not necessarily the main's HEAD here. We usually call it the base.

f0rmiga (approved):

PR #133, opened 2022-02-04
feat: Add the ability to have custom commands be called from plugins

JesseTatasciore:


plugins/fix-visibility/plugin.go
```diff
@@ -172,6 +173,10 @@ func (plugin *FixVisibilityPlugin) PostRunHook(
 	return plugin.PostBuildHook(isInteractiveMode, promptRunner)
 }
 
+func (plugin *FixVisibilityPlugin) CustomCommands() ([]*aspectplugin.Command, error) {
```
jbedard, line 176: Do we have to implement it even if we have none? To make it fully implement the `Plugin` interface? :/

pkg/plugin/system/system.go
```diff
@@ -45,6 +46,11 @@ type pluginSystem struct {
 	clients       []ClientProvider
 	plugins       *PluginList
 	promptRunner  ioutils.PromptRunner
+	callbackMap   map[string]CustomCommandCallbackFn
```
jbedard, line 49: Are we using the key anywhere?

jbedard: Actually I don't see this used anywhere where it's actually required...?

pkg/plugin/system/system.go
```diff
@@ -105,12 +112,61 @@ func (ps *pluginSystem) Configure(streams ioutils.Streams) error {
 		}
 
 		aspectplugin := rawplugin.(plugin.Plugin)
-		ps.plugins.insert(aspectplugin)
+
+		node := &PluginNode{
```
jbedard, line 116: Why do we create a new `PluginNode` type instead of putting the custom commands into the `aspectplugin`?

I'm assuming `aspectplugin := rawplugin.(plugin.Plugin)` converts the public API into an internal representation? Should that also wrap the custom commands?

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
```diff
@@ -147,6 +188,41 @@ func (m *GRPCClient) PostBuildHook(isInteractiveMode bool, promptRunner ioutils.
...
+	customCommandsPB, err := m.client.CustomCommands(context.Background(), req)
+
+	customCommands := make([]*Command, 0)
+
+	for _, pbCommand := range customCommandsPB.Commands {
+		customCommands = append(customCommands, &Command{
+			Use:       pbCommand.Use,
+			ShortDesc: pbCommand.ShortDesc,
+			LongDesc:  pbCommand.LongDesc,
+			Run:       nil,
```
jbedard, line 204: Why is the implementation `nil` in this case? Is this just to transfer the metadata about it?

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -27,3 +33,40 @@ type Plugin interface {
 		promptRunner ioutils.PromptRunner,
 	) error
 }
+
+type CustomCommandFn (func(ctx context.Context, args []string) error)
+
+var fnMap map[string]CustomCommandFn = make(map[string]CustomCommandFn)
```
jbedard, line 39: Why do we need to save them globally?

jbedard:
Sorry I'm going to be asking random questions to try to understand what's going on, forget me if they are stupid questions :P

cmd/aspect/main.go
```diff
@@ -45,6 +45,7 @@ func main() {
 	defer pluginSystem.TearDown()
 
 	cmd := root.NewDefaultRootCmd(pluginSystem)
+	cmd = pluginSystem.AddCustomCommands(cmd)
```
jbedard, line 48: Why do this here instead of in `NewDefaultRootCmd` where (I assume?) the rest of the plugin-data-extraction is done?

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
jbedard: Should the metadata vs implementation be split up? 🤔

pkg/plugin/system/system.go
```diff
@@ -45,6 +46,11 @@ type pluginSystem struct {
 	clients       []ClientProvider
 	plugins       *PluginList
 	promptRunner  ioutils.PromptRunner
+	callbackMap   map[string]CustomCommandCallbackFn
+}
+
+type CustomCommandCallbackFn interface {
```
jbedard, line 52: Why an interface instead of just a plain type (alias? not sure what it would be called). If it's an interface that wraps a callback I don't think the name itself should be called a "callback function"?

plugins/fix-visibility/plugin.go
JesseTatasciore: I think. I am not sure there is a way around this. The plugins need to implement all the other hooks even if they dont need them

pkg/plugin/system/system.go
JesseTatasciore: Its being used below in `AddCustomCommands`

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
JesseTatasciore: The run field can actually just be removed.

cmd/aspect/main.go
JesseTatasciore: I think you're right. Thats probably a better spot for it

pkg/plugin/sdk/v1alpha2/plugin/interface.go
JesseTatasciore: Definitely possible I didnt set this up properly here. Basically I needed this map to be shared between the calls of both `saveRunFunctions` and `executeRunFunction`

pkg/plugin/system/system.go
JesseTatasciore: I wanted the function `CustomCommandCallback` to be available when casted to this type. There definitely may be a better way to set this up. If you know how to change it so it doesnt use the interface then I am happy to change it. I just wasnt sure how to do that myself in this case

pkg/plugin/system/system.go
JesseTatasciore: `aspectplugin := rawplugin.(plugin.Plugin)` is casting `rawplugin` to `plugin.Plugin` which is public facking and I dont think we want those fields on there.

plugins/fix-visibility/plugin.go
jbedard: IMO we're going to have to think about that...

pkg/plugin/system/system.go
jbedard: That's unnecessary though, I think you can just delete this...

plugins/fix-visibility/plugin.go
JesseTatasciore: I agree that its not ideal. I just dont know how we could implement it differently. Might be worth getting some input from @f0rmiga / discussing it in our planning meeting

pkg/plugin/system/system.go
JesseTatasciore: If I remove this from the struct it compiles fine but throws a runtime error of:
```
panic: assignment to entry in nil map

goroutine 1 [running]:
aspect.build/cli/pkg/plugin/system.(*pluginSystem).AddCustomCommands(0x14000132070, 0x1400016e000)
	pkg/plugin/system/system.go:140 +0x210
aspect.build/cli/cmd/aspect/root.NewRootCmd({{0x10089cbe0, 0x14000010010}, {0x10089cc20, 0x14000010018}, {0x10089cc20, 0x14000010020}}, {0x1008b3950, 0x14000132070}, 0x1)
	cmd/aspect/root/root.go:134 +0xdf8
aspect.build/cli/cmd/aspect/root.NewDefaultRootCmd({0x1008b3950, 0x14000132070})
	cmd/aspect/root/root.go:41 +0xe0
main.main()
	cmd/aspect/main.go:47 +0x2ac
```

jbedard: Why?

Good chance I'm miss-reading it, but it seems like the only place we assign a value into this map is also the only place we read from it? In `AddCustomCommands`...

plugins/fix-visibility/plugin.go
f0rmiga: https://github.com/aspect-build/aspect-cli/pull/140

JesseTatasciore: I have remove this now. Thanks @f0rmiga !

pkg/plugin/system/system.go
JesseTatasciore: It is used twice inside of `AddCustomCommands`. We add to the map when `AddCustomCommands` is called but we access it as the `cobra` `RunE` execution happens. So its happening at a different point in time than the original call. We no longer have access to the `node` that we pulled `customCommandCallback` off of when the `RunE` function is called. So we can loop back through the list of nodes and find the node we need to get the `customCommandCallback` or we can save those callbacks somewhere so we can easily access them when we need them

jbedard: Can't you just use a local variable scoped to the `for` loop though? Something such as the `next` var in pkg/interceptors/run.go?

JesseTatasciore: Yeah that worked. Switched it out for that

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -54,3 +65,41 @@ func (*Base) PostTestHook(bool, ioutils.PromptRunner) error {
...
+func GetWorkspaceRoot(ctx context.Context) string {
+	return ctx.Value(interceptors.WorkspaceRootKey).(string)
+}
+
+func GetBazel(ctx context.Context) bazel.Bazel {
+	workspaceRoot := GetWorkspaceRoot(ctx)
+	bzl := bazel.New()
+	bzl.SetWorkspaceRoot(workspaceRoot)
+	return bzl
+}
```
f0rmiga, line 105: I think these 2 functions should not exist at all. I'd argue that what we want is:

```
type CustomCommandFn (func(ctx context.Context, args []string, bzl bazel.Bazel) error)
```

And then you make it totally hidden and the user doesn't need to ask for it to be constructed.

pkg/plugin/sdk/v1alpha2/proto/plugin.proto
```diff
@@ -10,6 +10,8 @@ option go_package = "aspect.build/cli/pkg/plugin/sdk/v1alpha2/proto";
 service Plugin {
   rpc BEPEventCallback(BEPEventCallbackReq) returns (BEPEventCallbackRes);
   rpc PostBuildHook(PostBuildHookReq) returns (PostBuildHookRes);
+  rpc CustomCommands(CustomCommandsReq) returns (CustomCommandsRes);
```
f0rmiga, line 13: Could you sort this whole section in alphabetical order, please?

pkg/plugin/system/system.go
```diff
@@ -105,12 +110,52 @@ func (ps *pluginSystem) Configure(streams ioutils.Streams) error {
...
+			node.customCommandCallback = customCommandCallback
+		}
+
+		ps.plugins.insert(node)
 	}
 
 	return nil
 }
 
+func (ps *pluginSystem) AddCustomCommands(cmd *cobra.Command) (*cobra.Command, error) {
```
f0rmiga, line 129: How about `RegisterCustomCommands`?

pkg/plugin/system/system.go
```diff
@@ -105,12 +110,52 @@ func (ps *pluginSystem) Configure(streams ioutils.Streams) error {
...
+	for node := ps.plugins.head; node != nil; node = node.next {
+		result, err := node.plugin.CustomCommands()
+		if err != nil {
+			return cmd, fmt.Errorf("unable to retrieve custom commands from plugin: %w", err)
```
f0rmiga, line 133: Wrong error context. It should be `failed to register custom commands: %w`.

plugins/hello-world/plugin.go
```diff
@@ -0,0 +1,57 @@
...
+			bzl := aspectplugin.GetBazel(ctx)
+
+			bzl.Spawn([]string{"help"})
+
+			return nil
+		},
+	})
+
+	return commandList, nil
+}
```
f0rmiga, line 57:
```suggestion
func (plugin *HelloWorldPlugin) CustomCommands() ([]*aspectplugin.Command, error) {
	return []*aspectplugin.Command{
		{
			Use:       "hello-world",
			ShortDesc: "Print 'Hello World!' to the command line.",
			LongDesc:  "Print 'Hello World!' to the command line. Echo any given argument. Then run a 'bazel help'",
			Run: func(ctx context.Context, args []string) error {
				fmt.Println("Hello World!")
				fmt.Print("Arguments passed to command: ")
				fmt.Println(args)
				fmt.Println("Going to run: 'bazel help'")

				bzl := aspectplugin.GetBazel(ctx)

				bzl.Spawn([]string{"help"})

				return nil
			},
		}
	}, nil
}
```

f0rmiga (changes requested):

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
```diff
@@ -74,6 +75,46 @@ func (m *GRPCServer) PostBuildHook(
...
+	ctx context.Context,
+	req *proto.CustomCommandsReq,
+) (*proto.CustomCommandsRes, error) {
+	customCommands, err := m.Impl.CustomCommands()
+
+	saveRunFunctions(customCommands)
+
+	pbCommands := make([]*proto.Command, 0)
+	for _, command := range customCommands {
+		pbCommand := &proto.Command{
```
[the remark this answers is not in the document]
JesseTatasciore: The issue I have with this is that when a plugin wants to return a `Command` they will be forced to add a `Command: &proto.Command{}` to their `Command` since you cannot currently set embedded struct values directly. Would look like this: 
```
&plugin.Command{
		Command: &proto.Command{
			Use:       "foo",
			ShortDesc: "bar",
			LongDesc:  "bar long",
		},
		Run: func(ctx context.Context, args []string) error {
			return nil
		},
	}
```

I much prefer how it currently looks:
```
&plugin.Command{
		Use:       "foo",
		ShortDesc: "bar",
		LongDesc:  "bar long",
		Run: func(ctx context.Context, args []string) error {
			return nil
		},
	}
```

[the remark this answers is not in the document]
JesseTatasciore: I feel like the second option creates a much better user experience.

pkg/plugin/sdk/v1alpha2/plugin/interface.go
JesseTatasciore: This has been fixed now

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
```diff
@@ -84,6 +92,46 @@ func (m *GRPCServer) PostBuildHook(
 		m.Impl.PostBuildHook(req.IsInteractiveMode, prompter)
 }
 
+// CustomCommands translates the gRPC call to the Plugin CustomCommands
+// implementation. It returns a list of commands that the plugin implements.
+func (m *GRPCServer) CustomCommands(
+	ctx context.Context,
+	req *proto.CustomCommandsReq,
+) (*proto.CustomCommandsRes, error) {
+	customCommands, err := m.Impl.CustomCommands()
```
jbedard, line 101: Why do we proceed here and not return if there is an `err`?

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -48,6 +54,11 @@ func (*Base) BEPEventCallback(*buildeventstream.BuildEvent) error {
 	return nil
 }
 
+// BEPEventCallback satisfies Plugin.BEPEventCallback.
```
jbedard, line 57: `BEPEventCallback` => `CustomCommands`?

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -62,3 +73,42 @@ func (*Base) PostTestHook(bool, ioutils.PromptRunner) error {
...
+}
+
+type PluginCommandManager struct {
+	commands map[string]CustomCommandFn
+}
+
+func (cm *PluginCommandManager) Save(commands []*Command) error {
+	for _, cmd := range commands {
+		if _, exists := cm.commands[cmd.Use]; exists {
+			return fmt.Errorf("command '%s' is declared more than once by plugin", cmd.Use)
```
jbedard, line 98: if the command is `build` will it fail here?

plugins/error-augmentor/plugin.go
```diff
@@ -20,13 +20,16 @@ import (
...
 	"aspect.build/cli/pkg/plugin/sdk/v1alpha2/config"
+	"aspect.build/cli/pkg/plugin/sdk/v1alpha2/plugin"
 )
 
 func main() {
 	goplugin.Serve(config.NewConfigFor(NewDefaultPlugin()))
 }
 
 type ErrorAugmentorPlugin struct {
+	plugin.Base
```
jbedard, line 31: Will the name of of this be `plugin` or `base`?

pkg/plugin/client/client.go
```diff
@@ -85,4 +98,6 @@ type Provider interface {
 type PluginInstance struct {
 	plugin.Plugin
 	Provider
+	CustomCommands        map[string]*plugin.Command
```
jbedard, line 101: @f0rmiga this is the kind of info I thought would end up in here. WDYT?

pkg/plugin/client/client.go
```diff
@@ -66,8 +74,13 @@ func (*clientFactory) New(aspectplugin loader.AspectPlugin, streams ioutils.Stre
 	}
 
 	res := &PluginInstance{
-		Plugin:   rawplugin.(plugin.Plugin),
-		Provider: goclient,
+		Plugin:         rawplugin.(plugin.Plugin),
+		Provider:       goclient,
+		CustomCommands: map[string]*plugin.Command{},
```
jbedard, line 79: Where does this one get used? (I'm probably just not finding it..?)

pkg/plugin/client/client.go
f0rmiga: Let's get this in, then we can think again about how to split this struct.

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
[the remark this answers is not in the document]
f0rmiga:
```go
func NewCommand(
    use string,
    shortDesc string,
    longDesc string,
    run func(ctx context.Context, args []string) error,
) *plugin.Command {
    return &plugin.Command{
        Command: &proto.Command{
            Use: use,
            ShortDesc: shortDesc,
            LongDesc: longDesc,
        },
        Run: run,
    }
}
```

pkg/plugin/sdk/v1alpha2/plugin/interface.go
f0rmiga: I don't think so. @JesseTatasciore we need some guard rails on the cobra side when constructing the commands.

plugins/error-augmentor/plugin.go
f0rmiga: I don't understand the question.

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -62,3 +73,42 @@ func (*Base) PostTestHook(bool, ioutils.PromptRunner) error {
...
+type Command struct {
+	Use       string
+	ShortDesc string
+	LongDesc  string
+	Run       CustomCommandFn
+}
+
+type CommandManager interface {
+	Save(commands []*Command) error
+	Execute(command string, ctx context.Context, args []string) error // Not sure about this signature, haven't read the whole code yet.
```
jbedard, line 88: Lets make sure we address or delete the comment here...

pkg/plugin/system/system.go
```diff
@@ -47,6 +48,13 @@ type pluginSystem struct {
 	promptRunner  ioutils.PromptRunner
 }
 
+// CustomCommandExecutor requires the Plugin implementations to provide the
+// ExecuteCustomCommand method so that the Core can ask over gRPC for a specific command to
+// be executed. `cmdName` is the name of the custom command the plugin created.
+type CustomCommandExecutor interface {
```
jbedard, line 54: This interface is duplicated?

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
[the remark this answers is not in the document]
JesseTatasciore: Added to the interface file so that its easy to use

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
JesseTatasciore: Fixed

pkg/plugin/sdk/v1alpha2/plugin/interface.go
JesseTatasciore: Fixed

plugins/error-augmentor/plugin.go
jbedard: Basically I don't understand (or remember) what this does :P

pkg/plugin/sdk/v1alpha2/plugin/interface.go
JesseTatasciore: Check added. Sample error: `Error: failed to register custom commands: plugin implements a command with a protected name: test`

pkg/plugin/client/client.go
JesseTatasciore: Its not. Good catch. I think I had added this about midway through my rebase with your changes (when I would have still needed it) and then never removed it when I got to my later changes where its was no longer needed

pkg/plugin/sdk/v1alpha2/plugin/interface.go
JesseTatasciore: Removing. Ive tested and the signature works

plugins/error-augmentor/plugin.go
JesseTatasciore: It adds the fields from `plugin.Base` to `ErrorAugmentorPlugin` so that it doesnt need to implement all of the methods. Ive added it here because there is now a `CustomCommands` method that this plugin does not implement

pkg/plugin/system/system.go
JesseTatasciore: Leftover from rebase I think. This one is not being used. Removed

pkg/plugin/client/client.go
```diff
@@ -66,8 +74,13 @@ func (*clientFactory) New(aspectplugin loader.AspectPlugin, streams ioutils.Stre
...
+	}
+
+	if customCommandExecutor, ok := rawplugin.(CustomCommandExecutor); ok {
+		res.CustomCommandExecutor = customCommandExecutor
```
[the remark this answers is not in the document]
JesseTatasciore: I am not sure? Me being new to Go might be showing through here but the point of this is to make sure the plugin has the method `ExecuteCustomCommand` that we can call. This is not part of `Plugin` at all. Its a method added internally on the GRPC interactions so that we can run the commands provided by the plugin on the running instance of the given plugin. I dont think we want to add it to the Plugin as then it may show up to the plugins as something they could possibly implement?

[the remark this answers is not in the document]
JesseTatasciore: We dont want the users to know about `ExecuteCustomCommand`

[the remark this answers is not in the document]
jbedard: This might be me being new to Go as well...

But both `CustomCommandExecutor` and `Plugin` are just the `rawplugin` casted, so they are references to the same object? If it's always references to the same object I think it would be nice to avoid having a different field for each object type.

Some languages support things like `CustomCommandExecutor & Plugin` to indicate that it is both. I'm not sure if Go has anything similar?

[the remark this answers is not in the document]
f0rmiga: We don't want to mix these 2 interfaces. The `Plugin` interface is the public API, it's what we enforce the plugin implementations to provide.

Because of the internal juggling we do over gRPC, the runtime `Plugin`, which is the `rawplugin` is not exactly the implementation provided by the plugin author. It's our internal representation of it, i.e. the gRPC implementation that wraps the real plugin logic implementation.

CLI core -> gRPC wrapper that satisfies the Plugin interface -> real implementation that satisfies the Plugin interface

Having said that, we could use some documentation here to avoid any future confusion. Or maybe pick names that translate better? E.g. `InternalExecutor`, or put this interface inside an `internal` package.

pkg/plugin/client/client.go
```diff
@@ -85,4 +97,5 @@ type Provider interface {
 type PluginInstance struct {
 	plugin.Plugin
 	Provider
+	CustomCommandExecutor CustomCommandExecutor
```
f0rmiga, line 100:
```suggestion
	CustomCommandExecutor
```

pkg/plugin/sdk/v1alpha2/plugin/grpc.go
```diff
@@ -168,6 +215,36 @@ func (m *GRPCClient) PostBuildHook(isInteractiveMode bool, promptRunner ioutils.
 	return err
 }
 
+// CustomCommands is called from the Core to execute the Plugin CustomCommands.
+// It returns a list of commands that the plugin implements.
+func (m *GRPCClient) CustomCommands() ([]*Command, error) {
+	req := &proto.CustomCommandsReq{}
+	customCommandsPB, err := m.client.CustomCommands(context.Background(), req)
+
+	customCommands := make([]*Command, 0)
```
f0rmiga, line 224:
```suggestion
	customCommands := make([]*Command, 0, len(customCommandsPB.Commands))
```

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -62,3 +74,56 @@ func (*Base) PostTestHook(bool, ioutils.PromptRunner) error {
 func (*Base) PostRunHook(bool, ioutils.PromptRunner) error {
 	return nil
 }
+
+type CustomCommandFn (func(ctx context.Context, args []string, bzl bazel.Bazel) error)
```
f0rmiga, line 78: Missing godoc on all these new types and functions.

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -62,3 +74,56 @@ func (*Base) PostTestHook(bool, ioutils.PromptRunner) error {
...
+type Command struct {
+	*proto.Command
+	Run CustomCommandFn
+}
+
+func NewCommand(
+	use string,
+	shortDesc string,
+	longDesc string,
+	run func(ctx context.Context, args []string, bzl bazel.Bazel) error,
```
f0rmiga, line 89:
```suggestion
	run CustomCommandFn,
```

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -62,3 +74,56 @@ func (*Base) PostTestHook(bool, ioutils.PromptRunner) error {
...
+}
+
+type PluginCommandManager struct {
+	commands map[string]CustomCommandFn
+}
+
+func (cm *PluginCommandManager) Save(commands []*Command) error {
+	for _, cmd := range commands {
+		if _, exists := cm.commands[cmd.Use]; exists {
+			return fmt.Errorf("command '%s' is declared more than once by plugin", cmd.Use)
```
f0rmiga, line 113: The `%q` is very helpful in Go when you want to quote a string when formatting it.

```suggestion
			return fmt.Errorf("command %q is declared more than once by plugin", cmd.Use)
```

pkg/plugin/system/system.go
```diff
@@ -98,6 +99,44 @@ func (ps *pluginSystem) addPlugin(plugin *client.PluginInstance) {
 	ps.plugins.insert(plugin)
 }
 
+func (ps *pluginSystem) RegisterCustomCommands(cmd *cobra.Command) error {
```
f0rmiga, line 102: Missing godoc.

pkg/plugin/system/system.go
```diff
@@ -98,6 +99,44 @@ func (ps *pluginSystem) addPlugin(plugin *client.PluginInstance) {
...
+	existingCommands := make(map[string]*cobra.Command)
```
f0rmiga, line 103:
```suggestion
	internalCommands := make(map[string]struct{})
```

pkg/plugin/system/system.go
```diff
@@ -98,6 +99,44 @@ func (ps *pluginSystem) addPlugin(plugin *client.PluginInstance) {
...
+
+	for _, command := range cmd.Commands() {
+		existingCommands[command.Use] = command
```
f0rmiga, line 106:
```suggestion
		internalCommands[command.Use] = struct{}{}
```

pkg/plugin/system/system.go
```diff
@@ -98,6 +99,44 @@ func (ps *pluginSystem) addPlugin(plugin *client.PluginInstance) {
...
+	}
+
+	for node := ps.plugins.head; node != nil; node = node.next {
+		result, err := node.payload.Plugin.CustomCommands()
+		if err != nil {
+			return fmt.Errorf("failed to register custom commands: %w", err)
+		}
+
+		for _, command := range result {
+			if _, ok := existingCommands[command.Use]; ok {
```
f0rmiga, line 116:
```suggestion
			if _, ok := internalCommands[command.Use]; ok {
```

f0rmiga (approved):

PR #135, opened 2022-02-05
test: add PluginSystem.commandHooksInterceptor tests

jbedard:
I think a lot more could be added but thought I'd open this as-is for now for any feedback...

### DIFF
--- a/pkg/plugin/system/system.go
+++ b/pkg/plugin/system/system.go
@@ -111,10 +111,14 @@ func (ps *pluginSystem) Configure(streams ioutils.Streams) error {
 			return fmt.Errorf("failed to setup plugin: %w", err)
 		}
 
-		ps.plugins.insert(aspectplugin)
+		ps.addPlugin(aspectplugin)
 	}
 
 	return nil
+}
+
+func (ps *pluginSystem) addPlugin(plugin plugin.Plugin) {
+	ps.plugins.insert(plugin)
 }
 
 // TearDown tears down the plugin system, making all the necessary actions to
@@ -180,9 +184,8 @@ func (ps *pluginSystem) commandHooksInterceptor(methodName string, streams iouti
 			return fmt.Errorf("failed to run 'aspect %s' command: %w", cmd.Use, err)
 		}
 
-		// TODO(f0rmiga): test this hook.
 		defer func() {
-			hasErrors := false
+			hasPluginErrors := false
 			for node := ps.plugins.head; node != nil; node = node.next {
 				params := []reflect.Value{
 					reflect.ValueOf(isInteractiveMode),
@@ -190,10 +193,10 @@ func (ps *pluginSystem) commandHooksInterceptor(methodName string, streams iouti
 				}
 				if err := reflect.ValueOf(node.plugin).MethodByName(methodName).Call(params)[0].Interface(); err != nil {
 					fmt.Fprintf(streams.Stderr, "Error: failed to run 'aspect %s' command: %v\n", cmd.Use, err)
-					hasErrors = true
+					hasPluginErrors = true
 				}
 			}
-			if hasErrors {
+			if hasPluginErrors {
 				var err *aspecterrors.ExitError
 				if errors.As(exitErr, &err) {
 					err.ExitCode = 1
--- a/pkg/plugin/system/system_test.go
+++ b/pkg/plugin/system/system_test.go
@@ -1,0 +1,259 @@
+/*
+Copyright © 2022 Aspect Build Systems Inc
+
+Not licensed for re-use.
+*/
+
+package system
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+
+	rootFlags "aspect.build/cli/pkg/aspect/root/flags"
+	"aspect.build/cli/pkg/aspecterrors"
+	"aspect.build/cli/pkg/ioutils"
+	plugin_mock "aspect.build/cli/pkg/plugin/sdk/v1alpha2/plugin/mock"
+)
+
+func createInterceptorCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "TestCommand",
+	}
+
+	// Required flags for interceptor hooks
+	cmd.PersistentFlags().Bool(rootFlags.InteractiveFlagName, false, "")
+
+	return cmd
+}
+
+func TestPluginSystemInterceptors(t *testing.T) {
+	t.Run("executes hooks in reverse order of interceptors", func(t *testing.T) {
+		g := NewGomegaWithT(t)
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		// Setup
+		var stdout strings.Builder
+		streams := ioutils.Streams{Stdout: &stdout, Stderr: &stdout}
+		ctx := context.Background()
+		cmd := createInterceptorCommand()
+
+		ps := NewPluginSystem().(*pluginSystem)
+		plugin := plugin_mock.NewMockPlugin(ctrl)
+		ps.addPlugin(plugin)
+
+		// Expect the callbacks in reverse-order of execution
+		gomock.InOrder(
+			plugin.EXPECT().PostRunHook(gomock.Any(), gomock.Any()),
+			plugin.EXPECT().PostTestHook(gomock.Any(), gomock.Any()),
+			plugin.EXPECT().PostBuildHook(gomock.Any(), gomock.Any()),
+		)
+
+		// Hook interceptors
+		buildInterceptor := ps.BuildHooksInterceptor(streams)
+		testInterceptor := ps.TestHooksInterceptor(streams)
+		runInterceptor := ps.RunHooksInterceptor(streams)
+
+		err := buildInterceptor(ctx, cmd, []string{}, func(ctx context.Context, cmd *cobra.Command, args []string) error {
+			return testInterceptor(ctx, cmd, args, func(ctx context.Context, cmd *cobra.Command, args []string) error {
+				return runInterceptor(ctx, cmd, args, func(ctx context.Context, cmd *cobra.Command, args []string) error {
+					return nil
+				})
+			})
+		})
+
+		g.Expect(err).To(BeNil())
+	})
+
+	t.Run("executes plugin hooks in order plugins are added", func(t *testing.T) {
+		g := NewGomegaWithT(t)
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		// Setup
+		var stdout strings.Builder
+		streams := ioutils.Streams{Stdout: &stdout, Stderr: &stdout}
+		ctx := context.Background()
+		cmd := createInterceptorCommand()
+
+		// Plugins to be invoked
+		ps := NewPluginSystem().(*pluginSystem)
+		plugin1 := plugin_mock.NewMockPlugin(ctrl)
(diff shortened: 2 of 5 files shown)


pkg/plugin/system/system_test.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"context"
+	"strings"
+	"testing"
+
+	buildeventstream "aspect.build/cli/bazel/buildeventstream/proto"
+	rootFlags "aspect.build/cli/pkg/aspect/root/flags"
+	"aspect.build/cli/pkg/ioutils"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
```
f0rmiga, line 18: We have been organizing the import statements in:

```
import (
    <std lib>

    <third-party>

    <first-party>
)
```

pkg/plugin/system/system_test.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+// A plugin to simply log plugin callbacks
+type LoggerPlugin struct {
```
f0rmiga, line 22: I think what you want is a mock of the `Plugin` interface. Take a look at for example https://github.com/aspect-build/aspect-cli/blob/2cc2cb276b936077943fd30f3f6aa6fa46acb069/pkg/aspect/build/build_test.go.

You can generate mocks like https://github.com/aspect-build/aspect-cli/tree/2cc2cb276b936077943fd30f3f6aa6fa46acb069/pkg/bazel/mock.

f0rmiga: This will change substantially the rest of your tests, but I think it will be cleaner and more standardized.

jbedard: Yeah this thing is a bit of a mess, and when expanding the tests more (testing error handling being a big one) it would just get worse. Hopefully mocks make that better, I'll take a look 👍

pkg/plugin/system/system_test.go
```diff
@@ -0,0 +1,252 @@
...
+		runInterceptor := ps.RunHooksInterceptor(streams)
+		e := runInterceptor(ctx, cmd, args, func(ctx context.Context, cmd *cobra.Command, args []string) error {
+			return &aspecterrors.ExitError{
+				Err:      fmt.Errorf("aspect error"),
+				ExitCode: 123,
+			}
+		})
+
+		g.Expect(e.(*aspecterrors.ExitError).Err).To(MatchError("aspect error"))
+		g.Expect(e.(*aspecterrors.ExitError).ExitCode).To(Equal(1))
```
jbedard, line 250: This is failing... does that mean https://github.com/aspect-build/aspect-cli/blob/2cc2cb276b936077943fd30f3f6aa6fa46acb069/pkg/plugin/system/system.go#L190-L195 is not working? Or do you see an issue with my test?

f0rmiga: It is. Your test needs to actually add a `DoAndReturn` to the mocked method:

```
diff --git a/pkg/plugin/sdk/v1alpha2/plugin/mock/BUILD.bazel b/pkg/plugin/sdk/v1alpha2/plugin/mock/BUILD.bazel
index b99dfa6..facdfbb 100644
--- a/pkg/plugin/sdk/v1alpha2/plugin/mock/BUILD.bazel
+++ b/pkg/plugin/sdk/v1alpha2/plugin/mock/BUILD.bazel
@@ -22,6 +22,7 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "//bazel/buildeventstream/proto",  # keep
+        "//pkg/ioutils",  # keep
         "//pkg/plugin/sdk/v1alpha2/plugin",  # keep
         "@com_github_golang_mock//gomock",  # keep
     ],
diff --git a/pkg/plugin/system/system_test.go b/pkg/plugin/system/system_test.go
index 1c721d6..ac8ddfa 100644
--- a/pkg/plugin/system/system_test.go
+++ b/pkg/plugin/system/system_test.go
@@ -231,7 +231,17 @@ func TestPluginSystem(t *testing.T) {
 		cmd.PersistentFlags().Bool(rootFlags.InteractiveFlagName, false, "")
 
 		plugin := plugin_mock.NewMockPlugin(ctrl)
-		plugin.EXPECT().PostRunHook(gomock.Any(), gomock.Any())
+		plugin.EXPECT().
+			PostRunHook(gomock.Any(), gomock.Any()).
+			DoAndReturn(func(
+				isInteractiveMode bool,
+				promptRunner ioutils.PromptRunner,
+			) interface{} {
+				return &aspecterrors.ExitError{
+					Err:      fmt.Errorf("plugin error"),
+					ExitCode: 123,
+				}
+			})
 
 		ps := NewPluginSystem().(*pluginSystem)
 		ps.Configure(streams)
```

f0rmiga:
I don't care about duplication in tests. It's often better to have "repetitive" tests that are easy to read than a super optimized file that nobody understands.

jbedard:
> It's often better to have "repetitive" tests that are easy to read than a super optimized file that nobody understands.

I want the most readable + maintainable which I think is the same as you're saying. Just let me know if you have any tips for doing that better 👍

pkg/plugin/system/system_test.go
jbedard: Oh, I'm just being stupid and returning the error from the wrong spot 🤦‍♂️ - thank you! lol

I guess there's 2 types of errors then (and I was only producing the second): the plugin invocations and the "next" invocation (which might be more plugin invocations)

pkg/plugin/system/system_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+		// Expect the callbacks in reverse-order of execution
+		gomock.InOrder(
+			plugin.EXPECT().PostRunHook(gomock.Any(), gomock.Any()),
+			plugin.EXPECT().PostTestHook(gomock.Any(), gomock.Any()),
+			plugin.EXPECT().PostBuildHook(gomock.Any(), gomock.Any()),
+		)
+
+		ps := NewPluginSystem().(*pluginSystem)
+		ps.Configure(streams)
```
f0rmiga, line 54: I think by doing this you are being too broad in your tests. Try to separate better in unit tests vs integration tests. I.e. if you are using the constructed plugin system from `NewPluginSystem`, you are already being too broad. E.g. `Configure` should have unit tests and the cases should be something like:

- Configure fails when Finder fails.
- Configure fails when Parser fails.
- Configure fails when getting an RPC Client fails (this one will be trickier and you will need to return a Client mock from a ClientFactory mock).
- Configure fails when getting a raw Client fails (same here).

pkg/plugin/system/system_test.go
```diff
@@ -0,0 +1,252 @@
...
+		ps.plugins.insert(plugin)
```
f0rmiga, line 55: As a continuation of my previous comment, you also don't want to do this in the tests. Instead, you instantiate a `pluginSystem` with the mocks and rely on their behaviour to control what happens inside the `Configure`. E.g. you will want your `Parser` to return the `aspectplugins` that then populate the `ps.plugins`.

pkg/plugin/system/system_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+		// Hook interceptors
+		buildInterceptor := ps.BuildHooksInterceptor(streams)
+		testInterceptor := ps.TestHooksInterceptor(streams)
+		runInterceptor := ps.RunHooksInterceptor(streams)
+
+		e := buildInterceptor(ctx, cmd, args, func(ctx context.Context, cmd *cobra.Command, args []string) error {
```
f0rmiga, line 62: `e` -> `err`

pkg/plugin/system/system_test.go
jbedard: I'm trying to unit test `commandHooksInterceptor`, I really don't care about `Configure` and would rather not have to invoke this at all. `Configure` should have its own tests like you're saying.

I guess that's what I don't like about the "repeated setup" that I mentioned earlier... it seems like too much and too fragile for a unit test so I'd like to either remove it or put it in a single place.

I think we're saying the same thing though...

pkg/plugin/system/system_test.go
jbedard: I'd rather not depend on anything in `Configure` though. The `commandHooksInterceptor` method really doesn't need anything from there other then a list of plugins.

One solution would be to essentially move the `commandHooksInterceptor` logic elsewhere so it can be unit tested without depending on things like `Configure`...

jbedard:
@f0rmiga wdyt about https://github.com/aspect-build/aspect-cli/pull/135/commits/6aeca0fbd9de620be2d32431f8beda9127155f72 + https://github.com/aspect-build/aspect-cli/pull/135/commits/ef9248197d6bafe45b43ec8a95379a874fb7731d

IMO that makes the tests significantly better because they are directly testing `commandHooksInterceptor` without any extra boilerplate to setup the `pluginSystem` or modify its internals...

jbedard:
@f0rmiga checkout the latest... I think I've done what we discussed in the first commit, although not exactly (I'm still using `NewPluginSystem()`).

The second 2 commits actually required changes in the code which I think are bug fixes, although it's not really clear what the original code was trying to do... it was setting the `exitErr.ExitCode` if a plugin error occurs, but `exitErr` is from interceptor errors and unrelated to plugins...

pkg/plugin/system/system.go
```diff
@@ -174,24 +178,25 @@ func (ps *pluginSystem) commandHooksInterceptor(methodName string, streams iouti
...
-			hasErrors := false
 			for node := ps.plugins.head; node != nil; node = node.next {
 				params := []reflect.Value{
 					reflect.ValueOf(isInteractiveMode),
 					reflect.ValueOf(ps.promptRunner),
 				}
 				if err := reflect.ValueOf(node.plugin).MethodByName(methodName).Call(params)[0].Interface(); err != nil {
-					fmt.Fprintf(streams.Stderr, "Error: failed to run 'aspect %s' command: %v\n", cmd.Use, err)
-					hasErrors = true
+					exitErr = &aspecterrors.ExitError{
```
jbedard, line 188: So this now propagates plugin errors to cause the interceptor to return an error. Is that correct or should it only print to `Stderror` and do nothing else?

f0rmiga (approved):

pkg/plugin/system/system.go
jbedard: Previously this error was only logged, `exitErr` was never assigned to and only came from `next()`. Assuming I'm understanding it correctly...

Previously:
* plugin errors are printed
* plugin errors set `hasErrors = true`
* interceptor errors were assigned `ExitCode = 1` **iff** `hasErrors` was true (<--- that **iff** is a bug imo)

Now:
1. plugin errors are printed
2. plugin errors are recorded and override any errors from `next()` (and override previous plugin errors)
3. all errors are assigned `ExitCode = 1`

I'm not sure which errors should have priority in 2) there though... should errors from `next()` override the plugin errors? (currently this PR does the opposite)

pkg/plugin/system/system_test.go
```diff
@@ -0,0 +1,260 @@
...
+			plugin.EXPECT().PostTestHook(gomock.Any(), gomock.Any()),
+			plugin.EXPECT().PostBuildHook(gomock.Any(), gomock.Any()),
+		)
+
+		// Hook interceptors
+		buildInterceptor := ps.BuildHooksInterceptor(streams)
+		testInterceptor := ps.TestHooksInterceptor(streams)
+		runInterceptor := ps.RunHooksInterceptor(streams)
+
+		err := buildInterceptor(ctx, cmd, []string{}, func(ctx context.Context, cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
jbedard: Isn't this (almost) equivalent to the func that `Run` creates? Each interceptor gets called within the previous...?

Issue #131, opened 2022-02-03
Passing --config flag does not override config load location


JesseTatasciore:
Current Behaviour: When I call `aspect --config ./foo.yaml someVerb` config is loaded from the default location

Expected Behaviour: When I call `aspect --config ./foo.yaml someVerb` config is ONLY loaded from the specified config file

![image](https://user-images.githubusercontent.com/11635434/152377757-3d6e90b9-971d-4002-8466-11b4c723e2a9.png)

Logic can be found in https://github.com/aspect-build/aspect-cli/blob/main/cmd/aspect/root/root.go
We attempt to set the `cfgFile` variable from the command line arguments but it appears that this never gets set. 

Current execution environment: M1 Macbook Air, zsh shell

JesseTatasciore:
- Prefix aspect flags with aspect:
- Ignore the aspect_deps flag that exists in bazel
- Ticket to deal with BES flag. We want that to be a bazel flag that we intercept. When the user provides there own bes flag we want to forward our internal BES server events to theirs
- Handle startup flags for bazel

JesseTatasciore:
Separate tickets created for dealing with BES flag and for handling startup flags:
https://github.com/aspect-build/aspect-cli/issues/136
https://github.com/aspect-build/aspect-cli/issues/137

Issue #43, opened 2021-10-01
Support common flags


alexeagle:
Currently this is an error:
```
$ aspect build --nobuild
Error: unknown flag: --nobuild
```

We have the code to parse `bazel help flags-as-proto` already, so we know all the flags that exist on the bazel build command and others.

We *could* just programatically add all the flags to all the commands in cobra. However then we're as bad as bazel, try running `bazel help build` and then scroll up forever.

Instead we probably want to:
- expose super-obvious flags in the CLI for all users
- permit "undocumented" flags to be passed through to Bazel
- [opt] user/repo settings file allows for more flags to appear.
 
Idea: maybe if you use a flag the first time, we add it to "your personal flags". But OTOH once you've used a flag, it's probably not important to you for it to appear in the help.

PR #138, opened 2022-02-08
Allow bazel flags to be passed through the aspect CLI

JesseTatasciore:
This PR current satisfies outstanding points from https://github.com/aspect-build/aspect-cli/issues/131
- Aspect specific flags are prefixed with `aspect:`
- `--aspect:config=/some/path` works properly now

This PR partially satisfies points made in https://github.com/aspect-build/aspect-cli/issues/43
- All appropriate bazel flags will be passed through to bazel. However, none of these flags are displayed to the user via `aspect help` which is mentioned in the ticket. Ticket mentions that we might want some of these to be displayed but short of doing this manually, it may be easier said than done


pkg/interceptors/bazel_flags.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"aspect.build/cli/pkg/bazel"
+	"github.com/spf13/cobra"
+)
+
+type Value interface {
```
jbedard, line 18: Where is this used?

pkg/interceptors/bazel_flags.go
```diff
@@ -0,0 +1,88 @@
...
+
+func BazelFlagInterceptor() Interceptor {
+	bzl := bazel.New()
+	return bazelflagInterceptor(bzl)
+}
+
+func bazelflagInterceptor(bzl bazel.Bazel) Interceptor {
+	return func(ctx context.Context, cmd *cobra.Command, args []string, next RunEContextFn) error {
+		workspaceRoot := ctx.Value(WorkspaceRootKey).(string)
+		bzl.SetWorkspaceRoot(workspaceRoot)
```
jbedard, line 53: Seems like currently we do this in every single `Command{RunE}`. Now we also do it within this interceptor?

What happens when `SetWorkspaceRoot` is set multiple times? Seems like the world would explode if the workspace root changes after the initial setup.

If every single `RunE` requires running `SetWorkspaceRoot`, and now this interceptor requires it... can the `WorkspaceRootInterceptor` not just set it once like it gets set here? That means any interceptor that requires `bzl.SetWorkspaceRoot()` being invoked must be run after the `WorkspaceRootInterceptor`... but that makes sense to me?

pkg/interceptors/bazel_flags.go
```diff
@@ -0,0 +1,88 @@
...
+							cmd.Flags().VarP(&key, flag, bzlFlags[flag].GetAbbreviation(), bzlFlags[flag].GetDocumentation())
+						} else {
+							cmd.Flags().StringP(flag, bzlFlags[flag].GetAbbreviation(), "", bzlFlags[flag].GetDocumentation())
+						}
+					}
+				}
+			}
+		}
+
+		if !cmd.DisableFlagParsing {
```
jbedard, line 75: Is this a requirement/prerequisite for this interceptor? Should such checks go at the top of the interceptor along with other things like the `bzl.Flags()` check (and maybe a "is workspace root set" if we address my other comment).

pkg/interceptors/bazel_flags.go
JesseTatasciore: Removed

pkg/interceptors/bazel_flags.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func bazelflagInterceptor(bzl bazel.Bazel) Interceptor {
+	return func(ctx context.Context, cmd *cobra.Command, args []string, next RunEContextFn) error {
+		workspaceRoot := ctx.Value(WorkspaceRootKey).(string)
+		bzl.SetWorkspaceRoot(workspaceRoot)
+
+		if bzlFlags, err := bzl.Flags(); err != nil {
+			return fmt.Errorf("unable to determine available bazel flags")
+		} else {
```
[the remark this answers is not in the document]
JesseTatasciore: bzlFlags is only available within the if / else when called like this

pkg/interceptors/bazel_flags.go
```diff
@@ -0,0 +1,88 @@
...
+			for flag := range bzlFlags {
+				for _, command := range bzlFlags[flag].Commands {
+					if command == cmd.Use {
+						if bzlFlags[flag].GetHasNegativeFlag() {
```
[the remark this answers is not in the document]
JesseTatasciore: Updated

pkg/interceptors/bazel_flags.go
```diff
@@ -0,0 +1,88 @@
...
+	return func(ctx context.Context, cmd *cobra.Command, args []string, next RunEContextFn) error {
+		workspaceRoot := ctx.Value(WorkspaceRootKey).(string)
+		bzl.SetWorkspaceRoot(workspaceRoot)
+
+		if bzlFlags, err := bzl.Flags(); err != nil {
+			return fmt.Errorf("unable to determine available bazel flags")
+		} else {
+			for flag := range bzlFlags {
+				for _, command := range bzlFlags[flag].Commands {
+					if command == cmd.Use {
```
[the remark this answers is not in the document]
JesseTatasciore: These loops now add all of the flags for all of the commands

pkg/interceptors/bazel_flags.go
JesseTatasciore: Removed

pkg/aspect/root/flags/bazel_flags.go
```diff
@@ -0,0 +1,95 @@
...
+	subCommands := make(map[string]*cobra.Command)
+
+	for _, command := range cmd.Commands() {
+		subCommands[command.Use] = command
+	}
+
+	bzl := bazel.New()
+	if bzlFlags, err := bzl.Flags(); err != nil {
+		return fmt.Errorf("unable to determine available bazel flags: %w", err)
+	} else {
```
jbedard, line 58: Since the `if` is just an error-check + return this doesn't need to be in an `else`?

pkg/aspect/root/flags/bazel_flags.go
```diff
@@ -0,0 +1,95 @@
...
+		for flagName := range bzlFlags {
+			flag := bzlFlags[flagName]
+			flagAbbreviation := flag.GetAbbreviation()
+			flagDoc := flag.GetDocumentation()
+
+			for _, command := range flag.Commands {
+				if subcommand, ok := subCommands[command]; ok {
```
jbedard, line 65: Is it worth putting the 2 parent `for` and `if` into a util method? So this would (I think) just look like:
```
subcommand = getSubcommand(flag.commands, more?)

if flag.GetHasNegativeFlag() {
  ... 
```
?

pkg/aspect/root/flags/bazel_flags.go
JesseTatasciore: Refactored

pkg/aspect/root/flags/bazel_flags.go
JesseTatasciore: To clarify, are you saying to add:
```
if subcommand, ok := subCommands[command]; ok {
  if flag.GetHasNegativeFlag() {
    subcommand.Flags().BoolP(flagName, flagAbbreviation, false, flagDoc)
    subcommand.Flags().Bool("no"+flagName, false, flagDoc)
    markFlagAsHidden(subcommand, flagName)
    markFlagAsHidden(subcommand, "no"+flagName)
  } else if flag.GetAllowsMultiple() {
    var key = MultiString{value: &[]string{}}
    subcommand.Flags().VarP(&key, flagName, flagAbbreviation, flagDoc)
    markFlagAsHidden(subcommand, flagName)
  } else {
    subcommand.Flags().StringP(flagName, flagAbbreviation, "", flagDoc)
    markFlagAsHidden(subcommand, flagName)
  }
}
```
to a function? What is the benefit going to be? Are you just wanting it to look cleaner?

jbedard: I think I might just be confused so maybe ignore me. Or maybe that means there's a clearer way of doing it and I'd be less confused 🤣

cmd/aspect/root/root.go
```diff
@@ -47,7 +45,7 @@ func NewRootCmd(
 	streams ioutils.Streams,
 	pluginSystem system.PluginSystem,
 	defaultInteractive bool,
-) *cobra.Command {
+) (*cobra.Command, error) {
```
f0rmiga, line 48: This is an analogue to a constructor in OO. I don't expect a constructor to return an error.

cmd/aspect/root/root.go
```diff
@@ -133,5 +89,8 @@ func NewRootCmd(
 		Long:  topics.MustAssetString("tags.md"),
 	})
 
-	return cmd
+	// Run this command after all bazel verbs have been added to "cmd".
+	err := flags.AddBazelFlags(cmd)
```
f0rmiga, line 93: Move this to outside this constructor so you don't need to return an error.

pkg/aspect/root/flags/bazel_flags.go
```diff
@@ -0,0 +1,96 @@
...
+var (
+	exposedBazelFlags = []string{
+		"keep_going",
+		"expunge",
+		"expunge_async",
+		"show_make_env",
+	}
+)
+
+type MultiString struct {
```
f0rmiga, line 27: Missing godocs for all these.

pkg/aspect/root/flags/bazel_flags.go
```diff
@@ -0,0 +1,96 @@
...
+	value *[]string
```
f0rmiga, line 28: Why make it a pointer if you are always dereferencing it over and over in the methods?

pkg/aspect/root/flags/interceptor.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"aspect.build/cli/pkg/interceptors"
+	"aspect.build/cli/pkg/ioutils"
+)
+
+var (
+	faint = color.New(color.Faint)
+)
+
+func AddGlobalFlags(cmd *cobra.Command, defaultInteractive bool) {
```
f0rmiga, line 26: Again, missing godocs.

pkg/aspect/root/flags/interceptor.go
```diff
@@ -0,0 +1,100 @@
...
+	cmd.PersistentFlags().String(ConfigFlagName, "", "config file (default is $HOME/.aspect.yaml)")
+	cmd.PersistentFlags().Bool(InteractiveFlagName, defaultInteractive, "Interactive mode (e.g. prompts for user input)")
+}
+
+func FlagsInterceptor(streams ioutils.Streams) interceptors.Interceptor {
+	return flagInterceptor(streams)
+}
+
+func flagInterceptor(streams ioutils.Streams) interceptors.Interceptor {
```
f0rmiga, line 35:
```suggestion
func FlagsInterceptor(streams ioutils.Streams) interceptors.Interceptor {
```

f0rmiga (changes requested):

BUILD.bazel
```diff
@@ -1,6 +1,11 @@
 load("@bazel_gazelle//:def.bzl", "gazelle")
 load("@io_bazel_rules_go//go:def.bzl", "TOOLS_NOGO", "nogo")
 
+exports_files([
```
alexeagle, line 4: add a comment why these files are "leaked" from the usual Bazel package encapsulation

cmd/aspect/main.go
```diff
@@ -49,6 +50,12 @@ func main() {
 
 	cmd := root.NewDefaultRootCmd(pluginSystem)
 
+	// Run this command after all bazel verbs have been added to "cmd".
+	if err := flags.AddBazelFlags(cmd); err != nil {
+		fmt.Fprintln(os.Stderr, "Error:", err)
```
alexeagle, line 55: this seems like strange error handling, is it intended?

cmd/aspect/version/version.go
```diff
@@ -36,17 +37,15 @@ func NewVersionCmd(streams ioutils.Streams, bzl bazel.Bazel) *cobra.Command {
...
 			func(ctx context.Context, cmd *cobra.Command, args []string) (exitErr error) {
-				workspaceRoot := ctx.Value(interceptors.WorkspaceRootKey).(string)
-				bzl.SetWorkspaceRoot(workspaceRoot)
 				return v.Run(bzl)
 			},
 		),
 	}
 
-	cmd.PersistentFlags().BoolVarP(&v.GNUFormat, "gnu_format", "", false, "format space-separated following GNU convention")
+	cmd.PersistentFlags().BoolVarP(&v.GNUFormat, "aspect:gnu_format", "", false, "format space-separated following GNU convention")
```
alexeagle, line 48: no this is actually a bazel flag, it exists on `bazel version`

docs/BUILD.bazel
```diff
@@ -9,8 +9,13 @@ _DOCS = ["aspect.md"] + [
 
 genrule(
     name = "command_list_bzl",
+    # We need to be able to start bazel so that we can get the correct flags
+    srcs = [
+        "//:WORKSPACE",
```
alexeagle, line 14: does it really need all the real file? could a minimal workspace work?

integration_tests/help_test.sh
```diff
@@ -4,6 +4,7 @@ set -o pipefail -o errexit -o nounset
 HOME="$TEST_TMPDIR"
 ASPECT="$TEST_SRCDIR/build_aspect_cli/cmd/aspect/aspect_/aspect"
 export HOME
+touch WORKSPACE
```
alexeagle, line 7: intentional?

pkg/aspect/root/flags/bazel_flags.go
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright 2022 Aspect Build Systems, Inc.
+ *
+ * Licensed under the aspect.build Commercial License (the "License");
+ * you may not use this file except in compliance with the License.
+ * Full License text is in the LICENSE file included in the root of this repository.
+ */
+
+/*
+Copyright © 2021 Aspect Build Systems Inc
```
alexeagle, line 10: doubled up headers

pkg/aspect/root/flags/bazel_flags.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"aspect.build/cli/pkg/bazel"
+)
+
+var (
+	exposedBazelFlags = []string{
```
alexeagle, line 27: doc this? I think "exposed" means "documented"? if so just rename

pkg/aspect/root/flags/interceptor.go
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2022 Aspect Build Systems, Inc.
+ *
+ * Licensed under the aspect.build Commercial License (the "License");
+ * you may not use this file except in compliance with the License.
+ * Full License text is in the LICENSE file included in the root of this repository.
+ */
+
+/*
+Copyright © 2021 Aspect Build Systems Inc
```
alexeagle, line 10: double headers

alexeagle (commented):
almost there, let's get this in! need it for RH already

BUILD.bazel
JesseTatasciore: Comment added

docs/BUILD.bazel
JesseTatasciore: Yes that worked

integration_tests/help_test.sh
JesseTatasciore: Yes. Need a WORKSPACE file there so that when the bazel server starts it doesnt throw an error. Bazel starts to generate the available flags

pkg/aspect/root/flags/bazel_flags.go
JesseTatasciore: done

cmd/aspect/main.go
f0rmiga: This is the main function. In this case, we just print the error to the user and exit non-zero.

alexeagle (approved):

PR #140, opened 2022-02-09
feat: plugin.Base

f0rmiga:
See comment on the `type Base struct{}`.

Also, see https://github.com/aspect-build/aspect-cli/pull/133#discussion_r799801199.

### DIFF
--- a/pkg/plugin/sdk/v1alpha2/plugin/interface.go
+++ b/pkg/plugin/sdk/v1alpha2/plugin/interface.go
@@ -27,3 +27,30 @@ type Plugin interface {
 		promptRunner ioutils.PromptRunner,
 	) error
 }
+
+// Base satisfies the Plugin interface. For plugins that only implement a subset
+// of the Plugin interface, using this as a base will give the advantage of not
+// needing to implement the empty methods.
+type Base struct{}
+
+var _ Plugin = (*Base)(nil)
+
+// BEPEventCallback satisfies Plugin.BEPEventCallback.
+func (*Base) BEPEventCallback(*buildeventstream.BuildEvent) error {
+	return nil
+}
+
+// PostBuildHook satisfies Plugin.PostBuildHook.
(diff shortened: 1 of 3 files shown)


plugins/fix-visibility/plugin.go
```diff
@@ -37,6 +38,8 @@ func main() {
 
 // FixVisibilityPlugin implements an aspect CLI plugin.
 type FixVisibilityPlugin struct {
+	plugin.Base
```
jbedard, line 41: This plug-in has nothing that can be deleted now?

(sorry on mobile add hard to tell atm...)

f0rmiga: Nope

pkg/plugin/sdk/v1alpha2/plugin/interface.go
```diff
@@ -27,3 +27,30 @@ type Plugin interface {
 		promptRunner ioutils.PromptRunner,
 	) error
 }
+
+// Base satisfies the Plugin interface. For plugins that only implement a subset
+// of the Plugin interface, using this as a base will give the advantage of not
+// needing to implement the empty methods.
+type Base struct{}
```
jbedard, line 34: Since this is public the name is essentially final (without a breaking change) so we should think about it...

Will this always be accessed with the package name such as `plugin.Base`? Or should the struct itself have more then just "base"

jbedard (approved):

pkg/plugin/sdk/v1alpha2/plugin/interface.go
f0rmiga: I thought about naming it also `Empty`, but it doesn't read well when consumed by the real plugin implementation.
Yes to the first question. No to the second question.

I'm happy to run alternative names with other folks. This is not as _final_ as you may think. This is SDK v1alpha2. Ideally, we wouldn't change but nothing will forbid us from doing so.

PR #141, opened 2022-02-09
fix: interceptors Run() when list of interceptors is empty

jbedard:


pkg/interceptors/run_test.go
```diff
@@ -49,6 +49,30 @@ func ExampleRun() {
 	// interceptor 1
 }
 
+func ExampleRun_nointerceptors() {
```
jbedard, line 52: Should this one be `_no_interceptors`? Or maybe `_noInterceptors` but I don't like that one :|

Have a preference for this odd naming convention?

jbedard: Currently I've done both the underscore and no-underscore so we should probably pick one and I'll update it...

f0rmiga (approved):

pkg/interceptors/run_test.go
f0rmiga: No, I'll never look at those names again.

PR #142, opened 2022-02-09
refactor: avoid redeclaring interceptor chain methods on each invocation

jbedard:
Didn't think of this last time I was looking at it... 🤷 

There's interesting differences between before/after that we could over-analyze. Unless this is invoked millions of times it's probably best to just take whichever one you think looks better.

Feel free to close if you prefer the old one or think it's just not worth changing it.


f0rmiga:
The proposed implementation is actually less efficient than the current implementation. Your proposed implementation runs the loop for all commands every time. The current implementation only runs once - for the command we ask, e.g. `build` or `test`, etc.

jbedard:
This PR makes it run once instead of every single time the returned method is invoked. So 1 instead of 0+.

But I think today only 1 of them gets invoked and only once... that's what you mean?

PR #143, opened 2022-02-09
test: add PluginSystem.Configure() tests

jbedard:
~~WIP with ugly unformulated code until #135 is merged, but let me know if you have any suggestions~~

~~This could be merged before #135 now. Either way there will be conflicts so doesn't matter which merges first 🤷~~

~~This PR currently includes #150~~

### DIFF
--- a/pkg/plugin/system/system_test.go
+++ b/pkg/plugin/system/system_test.go
@@ -10,6 +10,7 @@ package system
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"testing"
@@ -17,12 +18,15 @@ import (
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/gomega"
 	"github.com/spf13/cobra"
+	yaml "gopkg.in/yaml.v2"
 
 	rootFlags "aspect.build/cli/pkg/aspect/root/flags"
 	"aspect.build/cli/pkg/aspecterrors"
 	"aspect.build/cli/pkg/ioutils"
 	"aspect.build/cli/pkg/plugin/client"
 	client_mock "aspect.build/cli/pkg/plugin/client/mock"
+	"aspect.build/cli/pkg/plugin/loader"
+	loader_mock "aspect.build/cli/pkg/plugin/loader/mock"
 	plugin_mock "aspect.build/cli/pkg/plugin/sdk/v1alpha2/plugin/mock"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/plugin/system/system.go
```diff
@@ -217,14 +226,19 @@ type ClientProvider interface {
 	Kill()
 }
 
+type InternalPlugin struct {
```
jbedard, line 229: @JesseTatasciore I ended up doing something similar to what you did in your PR, to attach extra metadata to the plugin. I'm not sure which PR will merge first but let me know if there's anything here that doesn't align with yours or will cause annoying merge conflicts...

JesseTatasciore: I mean looking at what you changed, there will definitely be conflicts but I dont think it will be very hard to sort out

pkg/plugin/system/system.go
```diff
@@ -139,7 +130,7 @@ func (ps *pluginSystem) BESBackendInterceptor() interceptors.Interceptor {
 	return func(ctx context.Context, cmd *cobra.Command, args []string, next interceptors.RunEContextFn) error {
 		besBackend := bep.NewBESBackend()
 		for node := ps.plugins.head; node != nil; node = node.next {
-			besBackend.RegisterSubscriber(node.plugin.BEPEventCallback)
+			besBackend.RegisterSubscriber(node.plugin.plugin.BEPEventCallback)
```
f0rmiga, line 133: I don't think this change is necessary.

pkg/plugin/system/system.go
```diff
@@ -204,16 +195,35 @@ func (ps *pluginSystem) commandHooksInterceptor(methodName string, streams iouti
 	}
 }
 
-// ClientFactory hides the call to goplugin.NewClient.
+// ClientFactory hides the call to goplugin.NewClient() and conversion to InternalPlugin
```
f0rmiga, line 198:
```suggestion
// ClientFactory hides the call to goplugin.NewClient() and conversion to InternalPlugin.
```

pkg/plugin/system/system_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	goplugin "github.com/hashicorp/go-plugin"
+	. "github.com/onsi/gomega"
+
+	"aspect.build/cli/pkg/ioutils"
+	"aspect.build/cli/pkg/plugin/sdk/v1alpha2/plugin"
+)
+
+// Finder returning an error
+type errorFinder struct{}
```
f0rmiga, line 21: Use gomock instead for this and the below structs.

f0rmiga (changes requested):

pkg/plugin/system/system.go
jbedard: I'm sure there will be conflicts but I'd hope mainly just with naming? Let me know if we're doing anything fundamentally different that would make the conflicts worse...

pkg/plugin/system/system_test.go
jbedard: Done

pkg/plugin/system/system_test.go
```diff
@@ -10,10 +10,13 @@ package system
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"testing"
 
+	yaml "gopkg.in/yaml.v2"
+
```
f0rmiga, line 19:
```suggestion
	yaml "gopkg.in/yaml.v2"
```

pkg/plugin/system/system_test.go
```diff
@@ -279,3 +284,303 @@ func TestPluginSystemInterceptors(t *testing.T) {
...
+	t.Run("fails when Finder fails to find plugin config file", func(t *testing.T) {
+		g := NewGomegaWithT(t)
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		var stdout strings.Builder
+		streams := ioutils.Streams{Stdout: &stdout, Stderr: &stdout}
+
+		finder := loader_mock.NewMockFinder(ctrl)
+		finder.EXPECT().Find().Return("", errors.New("finderr"))
```
f0rmiga, line 298:
```suggestion
		finder.EXPECT().Find().Return("", errors.New("finder"))
```

pkg/plugin/system/system_test.go
```diff
@@ -279,3 +284,303 @@ func TestPluginSystemInterceptors(t *testing.T) {
...
+
+		ps := &pluginSystem{finder: finder}
+
+		err := ps.Configure(streams)
+
+		g.Expect(err).NotTo(BeNil())
```
f0rmiga, line 304: Gotta check if the error is the same as you expect. Not looking at the implementation of `Configure`, how can we be sure the error is the one from `Find`? The same applies below.

pkg/plugin/system/system_test.go
```diff
@@ -279,3 +284,303 @@ func TestPluginSystemInterceptors(t *testing.T) {
...
+		ps := &pluginSystem{
+			finder:        finder,
+			parser:        parser,
+			clientFactory: factory,
+			plugins:       &PluginList{},
+		}
+
+		err := ps.Configure(streams)
+
+		g.Expect(err).ToNot(BeNil())
```
f0rmiga, line 474: This one is critical for checking if the error matches the plugin one.

f0rmiga (changes requested):

pkg/plugin/system/system_test.go
```diff
@@ -276,6 +280,307 @@ func TestPluginSystemInterceptors(t *testing.T) {
...
+			LogLevel: "debug",
+		}
+		testPlugin2 := loader.AspectPlugin{
+			Name:     "test plugin2",
+			From:     "...",
+			LogLevel: "debug",
+		}
+
+		p1 := plugin_mock.NewMockPlugin(ctrl)
+		p1.EXPECT().Setup(gomock.Any())
```
f0rmiga, line 378: In a follow-up PR, it would be nice to be explicit about what the Setup takes. I.e. testing that it's getting called with the expected args.

pkg/plugin/system/system_test.go
```diff
@@ -276,6 +280,307 @@ func TestPluginSystemInterceptors(t *testing.T) {
...
+
+	t.Run("marshaled properties are passed to plugin.Setup", func(t *testing.T) {
+		g := NewGomegaWithT(t)
+		ctrl := gomock.NewController(t)
+		defer ctrl.Finish()
+
+		var stdout strings.Builder
+		streams := ioutils.Streams{Stdout: &stdout, Stderr: &stdout}
+
+		pluginProperties := make(map[string]interface { // foo: {},
```
f0rmiga, line 546: Remove comment?

pkg/plugin/system/system_test.go
jbedard: I've updated some of them, but this specific one accepts an array of bytes which is pretty awkward. I think I'd only want to verify the bytes are correct in a test explicitly testing that, to avoid the "data => bytes" logic being repeated everywhere.

f0rmiga (approved):

PR #145, opened 2022-02-12
feat: add outputs command

alexeagle:
Things to resolve before merge:

- [ ] common logic for verifying the label supplied by user exists, maybe help them
- [ ] when no arguments and interactive, prompt for them (maybe shared with query?)
- [x] move code to a plugin, not the core - blocked on https://github.com/aspect-build/aspect-cli/pull/133
- [ ] produce docs (but depends how plugins are documented?)
- [x] add a demo asciinema to this PR and the marketing site


pkg/bazel/analysis_v2.pb.go
```diff
@@ -0,0 +1,1331 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
f0rmiga, line 1: Why did this get checked in?

f0rmiga:
@alexeagle this is ready for a rebase since https://github.com/aspect-build/aspect-cli/pull/133 got merged.

pkg/bazel/analysis_v2.pb.go
alexeagle: same reason as flags.pb.go next to it - editor tooling is not working for codegen in bazel-out. I'll remove this from the PR for now, it's not load-bearing and also it's missing an up-to-date test

alexeagle:
okay it's rebased as a plugin that exposes the `outputs` command now.

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (plugin *Outputs) CustomCommands() ([]*aspectplugin.Command, error) {
+	return []*aspectplugin.Command{
+		aspectplugin.NewCommand(
+			"outputs",
+			"Inspect declared outputs.",
+			"Queries for the outputs declared by actions generated by the given target.",
+			func(ctx context.Context, args []string, bzl bazel.Bazel) error {
+				workspaceRoot := ctx.Value(interceptors.WorkspaceRootKey).(string)
+				bzl.SetWorkspaceRoot(workspaceRoot)
```
JesseTatasciore, line 54: This is done for you behind the scenes. `bzl` should come into the function ready to go. Happens here: 
https://github.com/aspect-build/aspect-cli/blob/main/pkg/plugin/sdk/v1alpha2/plugin/interface.go#L135

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+func main() {
+	goplugin.Serve(config.NewConfigFor(NewDefaultPlugin()))
+}
+
+type Outputs struct {
+	aspectplugin.Base
+	ioutils.Streams
+	Bzl bazel.Bazel
```
JesseTatasciore, line 32: I dont think this is actually being used? A `bzl` instance comes through with the function callback now

pkg/bazel/analysis_v2.proto
```diff
@@ -0,0 +1,226 @@
+// Copyright 2018 The Bazel Authors. All rights reserved.
```
JesseTatasciore, line 1: I noticed the v2 on this file. Note that I have basically no knowledge about the bazel proto and its versioning but it made me think: If there is a v2 then there was probably a v1 and could possibly be a v3. Could we get ourselves into a situation where the version of bazel being used causes problems here? Either with super old versions of bazel or future new versions?

plugins/outputs/plugin.go
alexeagle: oh that's great, since I'm sure plugin authors wouldn't remember to do that and would get wrong bazel versions

plugins/outputs/plugin.go
alexeagle: yeah good call, the streams wasn't used anymore either

pkg/bazel/analysis_v2.proto
alexeagle: this was a strange case, specific to aquery, where they changed the data representation entirely. There was a `aquery --noincompatible_proto_output_v2` flag but it was default-off in bazel 4 and removed in bazel 5.
I don't think this is a pattern we'll have to worry about often, no.

pkg/bazel/bazel.go
```diff
@@ -94,3 +95,31 @@ func (b *bazel) Flags() (map[string]*FlagInfo, error) {
 
 	return flags, nil
 }
+
+func (b *bazel) AQuery(query string) (*ActionGraphContainer, error) {
```
f0rmiga, line 99: Missing godoc. Actually, we need to add more docs to the public API of this package.

pkg/bazel/bazel.go
```diff
@@ -94,3 +95,31 @@ func (b *bazel) Flags() (map[string]*FlagInfo, error) {
...
+	defer close(bazelErrs)
+	go func() {
+		defer w.Close()
+		_, err := b.RunCommand([]string{"aquery", "--output=proto", query}, w)
+		bazelErrs <- err
+	}()
+
+	protoBytes, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Bazel aquery output: %w", err)
```
f0rmiga, line 113: The pattern for the error message should contain the current context. See my suggestion. This is so that we can migrate to Go 2 error handling with minimal effort. https://go.googlesource.com/proposal/+/master/design/go2draft-error-handling-overview.md

```suggestion
		return nil, fmt.Errorf("failed to run Bazel aquery: %w", err)
```

pkg/bazel/bazel.go
```diff
@@ -94,3 +95,31 @@ func (b *bazel) Flags() (map[string]*FlagInfo, error) {
...
+	}
+
+	if err := <-bazelErrs; err != nil {
+		return nil, fmt.Errorf("failed to run Bazel aquery: %w", err)
+	}
+
+	proto.Unmarshal(protoBytes, agc)
+	if err := proto.Unmarshal(protoBytes, agc); err != nil {
+		return nil, fmt.Errorf("Failed to parse ActionGraphContainer: %w", err)
```
f0rmiga, line 122: In this case, it's okay to add some extra information to the error, and it won't break the Go 2 error handling.

```suggestion
		return nil, fmt.Errorf("failed to run Bazel aquery: parsing ActionGraphContainer: %w", err)
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright 2022 Aspect Build Systems, Inc.
+ *
+ * Licensed under the aspect.build Commercial License (the "License");
+ * you may not use this file except in compliance with the License.
+ * Full License text is in the LICENSE file included in the root of this repository.
+ */
+
+package main
```
f0rmiga, line 9: I'd add some high-level descriptions for this plugin here.

```suggestion
// The output plugin ...

package main
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+	return &Outputs{}
+}
+
+func (plugin *Outputs) CustomCommands() ([]*aspectplugin.Command, error) {
+	return []*aspectplugin.Command{
+		aspectplugin.NewCommand(
+			"outputs",
+			"Inspect declared outputs.",
+			"Queries for the outputs declared by actions generated by the given target.",
+			func(ctx context.Context, args []string, bzl bazel.Bazel) error {
```
f0rmiga, line 41: It would be beneficial to have a top-level function that returns this and allows injecting some dependencies for testing (if needed, otherwise this plugin will just be a good reference for future plugins). Also will help with reducing a bit the indentation of the whole implementation.

```go
func NewOutputsCommand() aspectplugin.CustomCommandFn {
    return func(ctx context.Context, args []string, bzl bazel.Bazel) error {
        ...
    }
}
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+				if len(args) < 1 {
+					return fmt.Errorf("A label is required as the first argument to aspect outputs")
```
f0rmiga, line 43: By convention in Go, errors should never be capitalized.
```suggestion
					return fmt.Errorf("a label is required as the first argument to aspect outputs")
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+				}
+				query := args[0]
+				var mnemonicFilter string
+				if len(args) > 1 {
+					mnemonicFilter = args[1]
+				} else {
+					mnemonicFilter = ""
+				}
```
f0rmiga, line 51: The zero-value of the string in line 46 is already an empty string.
```suggestion
				}
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+				agc, err := bzl.AQuery(query)
+				if err != nil {
+					return err
+				}
+
+				// Use RAM to store lookup maps for these identifiers
+				// rather than an O(n^2) algorithm of searching on each access
```
f0rmiga, line 58:
```suggestion
				// rather than an O(n^2) algorithm of searching on each access.
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+				for _, f := range agc.PathFragments {
+					frags[f.Id] = f
+				}
+				arts := make(map[uint32]*bazel.Artifact)
+				for _, a := range agc.Artifacts {
+					arts[a.Id] = a
+				}
+
+				// The paths in the proto data are organized as a trie
+				// to make the representation more compact
```
f0rmiga, line 69:
```suggestion
				// to make the representation more compact.
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+				// https://en.wikipedia.org/wiki/Trie
+				// Make a map to store each prefix so we can memoize common paths
```
f0rmiga, line 71:
```suggestion
				// Make a map to store each prefix so we can memoize common paths.
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+				prefixes := make(map[uint32]*[]string)
+
+				// Declare a recursive function to walk up the trie to the root
```
f0rmiga, line 74:
```suggestion
				// Declare a recursive function to walk up the trie to the root.
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+				var prefix func(pathID uint32) []string
+
+				prefix = func(pathID uint32) []string {
+					if prefixes[pathID] != nil {
+						return *prefixes[pathID]
+					}
+					fragment := frags[pathID]
+					// reconstruct the path from the parent pointers
```
f0rmiga, line 82:
```suggestion
					// Reconstruct the path from the parent pointers.
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+				// The paths in the proto data are organized as a trie
+				// to make the representation more compact
+				// https://en.wikipedia.org/wiki/Trie
+				// Make a map to store each prefix so we can memoize common paths
+				prefixes := make(map[uint32]*[]string)
+
+				// Declare a recursive function to walk up the trie to the root
+				var prefix func(pathID uint32) []string
+
+				prefix = func(pathID uint32) []string {
```
f0rmiga, line 77: Interesting.

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+
+					if fragment.ParentId > 0 {
+						segments = append(segments, prefix(fragment.ParentId)...)
+					}
+					prefixes[pathID] = &segments
+					return segments
+				}
+
+				for _, a := range agc.Actions {
+					if len(mnemonicFilter) > 0 && (a.Mnemonic != mnemonicFilter) {
```
f0rmiga, line 93:
```suggestion
					if len(mnemonicFilter) > 0 && a.Mnemonic != mnemonicFilter {
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+						continue
+					}
+					for _, i := range a.OutputIds {
+						artifact := arts[i]
+						segments := prefix(artifact.PathFragmentId)
+						var path strings.Builder
+						// assemble in reverse order
```
f0rmiga, line 100:
```suggestion
						// Assemble in reverse order.
```

plugins/outputs/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+						for i := len(segments) - 1; i >= 0; i -= 1 {
```
f0rmiga, line 101:
```suggestion
						for i := len(segments) - 1; i >= 0; i-- {
```

plugins/outputs/plugin.go
alexeagle: yes I had to look this up, to create a recursive function within a scope, it must be declared separately from being defined.

plugins/outputs/plugin.go
alexeagle: I think I understand this, just means "extract function with a fancy Factory signature"

f0rmiga (approved):

PR #149, opened 2022-02-15
refactor: move plugin client into own package

jbedard:
Just the first step but I think mergable on its own...


jbedard:
Actually hold on, need to move a bit more...

jbedard:
Closing this for now, doing this before the other refactoring in the other PR is a bit more then just moving things...

PR #151, opened 2022-02-17
feat: run plugin setup concurrently 

jbedard:
Need to merge tests first: #143

I might add a few more tests involving more then 2 plugins being setup. Things like having the 1st vs 3rd fail etc.

### DIFF
--- a/pkg/plugin/system/system.go
+++ b/pkg/plugin/system/system.go
@@ -15,6 +15,8 @@ import (
 	"reflect"
 	"time"
 
+	"golang.org/x/sync/errgroup"
+
 	yaml "gopkg.in/yaml.v2"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


pkg/plugin/system/system.go
```diff
@@ -15,6 +15,8 @@ import (
 	"reflect"
 	"time"
 
+	"golang.org/x/sync/errgroup"
```
f0rmiga, line 18: Could you just fix the grouping here?

pkg/plugin/system/system.go
```diff
@@ -71,28 +71,32 @@ func (ps *pluginSystem) Configure(streams ioutils.Streams) error {
...
+			propertiesBytes, err := yaml.Marshal(p.Properties)
+			if err != nil {
+				return fmt.Errorf("failed to configure plugin system: %w", err)
+			}
 
-		if err := aspectplugin.Setup(propertiesBytes); err != nil {
-			return fmt.Errorf("failed to setup plugin: %w", err)
-		}
+			if err := aspectplugin.Setup(propertiesBytes); err != nil {
+				return fmt.Errorf("failed to configure plugin system: %w", err)
```
f0rmiga, line 91: Could you fix the test for this, please?

f0rmiga (approved):

PR #152, opened 2022-02-17
fix: apply license header to source files

alexeagle:
This is the text suggested by Aaron, our IP lawyer.


f0rmiga (approved):

pkg/bazel/flags.pb.go
```diff
@@ -1,3 +1,11 @@
+/*
```
f0rmiga, line 1: Actually, this is generated code that is not covered by our license.

alexeagle: yeah I know, but we don't assert that this matches the code generator so I think it's harmless and simpler to let the pre-commit check govern all .go files rather than try to exclude some.

Issue #94, opened 2022-01-06
Save bazel commands that will be run to a log file


JesseTatasciore:
- Everytime the aspect-cli runs a bazel command we should save it to a log file (useful for debugging, learning, etc)
- Everytime the aspect-cli run we should inform the user where they can find the log of bazel commands (should add a config option to disable this as well in case the user does not want this to happen)

alexeagle:
Use an off-the-shelf logger library. Name a directory with an invocation ID, then it contains /aspect.log, /plugin1.log, /crash.log, /command.log etc

Make it easy for us to log more things from more places.
Thulio has ideas of what API to use to expose the logger.
Plugins should each get their own log file.

alexeagle:
https://www.youtube.com/watch?v=QrGrOK8oZG8

alexeagle:
We can have log levels in the API, but let's just set the level to DEBUG always for now, with no user affordance to change it. Can revisit later. Generally we'd like more data available for us to diagnose issues rather than less.

alexeagle:
Maybe use https://pkg.go.dev/log/syslog if it meets our requirements?
logstash? humio?

https://www.humio.com/log-management/
https://www.elastic.co/logstash/

PR #154, opened 2022-02-17
Add logging capabilities to the cli core and plugins

JesseTatasciore:


pkg/bazel/bazel.go
```diff
@@ -13,12 +13,56 @@ import (
...
+		stringifyInt(t.Minute(), 2) +
+		"-" +
+		stringifyInt(t.Second(), 2) +
+		nanosecondStr[:2] +
+		"-" +
+		nanosecondStr[2:6]
+
+	superSpecialHexString := "617370656374" // I wonder what this says when converted back to a string?
+
+	invocationID = dateString + "-" + timeString + "-" + superSpecialHexString
```
jbedard, line 44: Did we think about putting the process-id in here?

JesseTatasciore: I dont think so? The discussion we had was mostly about how we should set the ID and make it based on a date so that they are ordered

jbedard: I'd say it could be useful, especially if aspect is is launched by another tool which records PIDs, or if aspect launches child processes that may log their parent ID etc. Then you can actually find the aspect cli logs...

JesseTatasciore: I took a quick look. Looks like I can get the current PID easily so shouldnt be a problem. The bigger concern is that there are rules about the invocation ID we give. I would add the PID at the end. After the second `-`. This bit needs to be 12 characters long. Based on some googling I dont think we have an issue of it being longer than that, but it will always be shorter. So if we had a PID of `12345` I would likely want our invocation ID to end with `000000012345`. Thoughts?

jbedard: I assume that's fine. Are we following a standard of some type here?

JesseTatasciore: Cool. And its just the standard UUID format. Bazel will reject it if we dont follow it correctly. The standard format is `8-4-4-4-12`. Where each number represents the expected number of hex characters.

alexeagle:
Hey, sorry I disappeared from the repo for a bit - what's next for this one?

JesseTatasciore:
I havent looked at this in a while now. I just linked https://github.com/aspect-build/aspect-cli/issues/94 as this was the issue related to this PR I think. If I recall correctly I had it working but I was not sure it satisfied what we actually wanted in a logger. As in the way that it was used / implemented. We had several discussions where we talked about it but it was hard to get a read on what we actually wanted to do (everyone had a different opinion). There were also a few things that I needed to look into still in order to see if we could use them for our logging purposes

gregmagolan:
This should go in upstream in silo

PR #157, opened 2022-02-28
chore: check in build event stream Go proto

alexeagle:
This makes it easier to write plugins which import from this folder, allowing them to get type-completion
on proto fields.
The alternative would require fancy editor setup to locate this file in bazel-out/external/aspect-cli


mattem:
Could we add a diff test to ensure the checked in file is up to date?

alexeagle:
Yeah let's start by fixing the existing issue in that folder - https://github.com/aspect-build/aspect-cli/pull/159

PR #158, opened 2022-03-02
chore: update bazel docsite URL

alexeagle:


JesseTatasciore (approved):
I think you just need to update the docs to fix the build

PR #159, opened 2022-03-02
chore: assert that checked-in proto is uptodate

alexeagle:


pkg/bazel/BUILD.bazel
```diff
@@ -34,3 +35,19 @@ go_proto_library(
...
+filegroup(
+    name = "go_generated_srcs",
+    srcs = [":bazel_go_proto"],
+    output_group = "go_generated_srcs",
+)
+
+write_source_files(
+    name = "write_flags_pb",
+    files = {
+        "flags.pb.go": ":go_generated_srcs",
```
alexeagle, line 51: Sadly this trick won't work when there are two proto files in this directory.

@kormide any other suggestions?

kormide: I don't quite follow. Do you mean that `go_generated_srcs` is a filegroup of more than one file, and so `write_source_files` doesn't know which one to write?

kormide: Or is this that you can't have the same label for the source/output file, so you're putting it in a filegroup? In that case could you just copy the file in the output dir to a different name, then reference that in the dict value?

alexeagle: Yeah rules_go will produce more than one file in that output group, and we don't control the names of the files it writes. We want to copy all the output files to the current directory rather than to a new nested subfolder.

kormide: Hm, what if we supported a wildcard in the macro like:
```
files = {
  "*": ":go_generated_srcs",
}
```
The macro would recognize it as a macro and not pass it into the rule as a label. The rule would somehow know to look at the DefaultInfo of `:go_generated_srcs` and copy the files to the current output dir.

alexeagle: Could you play with this one a bit from Paul Johnston:
https://github.com/stackb/rules_proto/blob/7c95feba87ae269d09690fcebb18c77d8b8bcf6a/rules/proto_compiled_sources.bzl
it calls through to his "gencopy" rule, and from a quick read I'm not sure if it supports this case.

mattem:
Is this ready for review? The bot keeps pining me but I'm unsure of its status. If it's not ready, can it be made a draft?

alexeagle:
Sure, this is reviewable and can land, it's an improvement. However we'll have to figure something out before we can add a second pb.go in this folder.

mattem (approved):

pkg/bazel/BUILD.bazel
gregmagolan: we have a solution for this now in bazel-lib 0.5.3 using default_output_files rule

alexeagle:
working on a new approach with Greg: https://github.com/aspect-build/bazel-examples/commit/134e49f3b7bc63510ab718f4ceeb8df535dc3069

Issue #162, opened 2022-03-07
Hitting max gRPC request size


alexeagle:
At RH we got a very large buildEvent payload and failed:

```
(22:26:11) ERROR: The Build Event Protocol upload failed: Not retrying publishBuildEvents, no more attempts left: status='Status{code=RESOURCE_EXHAUSTED, description=grpc: received message larger than max (4509322 vs. 4194304), cause=null}' RESOURCE_EXHAUSTED: RESOURCE_EXHAUSTED: grpc: received message larger than max (4509322 vs. 4194304) RESOURCE_EXHAUSTED: RESOURCE_EXHAUSTED: grpc: received message larger than max (4509322 vs. 4194304)
(22:26:11) INFO: Build completed, 1 test FAILED, 3335 total actions
Error: failed to run 'aspect test' command: rpc error: code = Unknown desc = No exitcode was reported during the Bazel execution, was the buildFinished event missing?
```

I guess there's probably a setting in the gRPC server we can adjust or expose to the .aspectrc file?

alexeagle:
Also in that error, there's no way to see which plugin reports that final error back to the cli.

PR #163, opened 2022-03-07
fix: bump BES server recv max message size

f0rmiga:
Bazel doesn't seem to set an outbound limit for gRPC message sizes. Here, since the [BES server is not exposed to the network](https://github.com/aspect-build/aspect-cli/blob/fac6baa3307651b488b9c46f371b4f0efbd46fde/pkg/plugin/system/bep/bes_backend.go#L64), it's okay to set the max value.


alexeagle (approved):
Thanks!

alexeagle:
Fixes https://github.com/aspect-build/aspect-cli/issues/162

PR #165, opened 2022-03-08
feat: plugin.Setup with structured config

f0rmiga:
This allows plugins to access some metadata from the aspectplugins.


pkg/plugin/sdk/v1alpha3/plugin/interface.go
```diff
@@ -35,7 +35,38 @@ type Plugin interface {
 		isInteractiveMode bool,
 		promptRunner ioutils.PromptRunner,
 	) error
-	Setup(properties []byte) error
+	Setup(config *SetupConfig) error
+}
+
+// SetupConfig represents a plugin configuration parsed from the aspectplugins
+// file.
+type SetupConfig struct {
```
f0rmiga, line 43: @alexeagle Is this sufficient?

plugins/error-augmentor/plugin.go
```diff
@@ -45,7 +44,6 @@ func NewDefaultPlugin() *ErrorAugmentorPlugin {
 
 func NewPlugin() *ErrorAugmentorPlugin {
 	return &ErrorAugmentorPlugin{
-		properties:          nil,
```
f0rmiga, line 48: @JesseTatasciore This didn't seem to be used. Was this just a left-over?

pkg/plugin/sdk/v1alpha3/plugin/interface.go
alexeagle: yeah this covers my use case, thanks!

alexeagle (approved):

plugins/error-augmentor/plugin.go
JesseTatasciore: Yeah it must have been

PR #173, opened 2022-03-10
Fix protobuf namespace issue caused by mirroring bazel protobufs and importing buildozer

JesseTatasciore:


alexeagle (approved):
nice find :)

Issue #167, opened 2022-03-10
Plugins load from remote or from local source


alexeagle:
Currently the plugins can only load from a path on disk. Code is here https://github.com/aspect-build/aspect-cli/blob/0f230e151e57b5b539a11b28b4ec8216f211a27b/pkg/plugin/client/client.go#L57

Users should be able to write
- `from: github.com/my-org/my-plugin@1.2.3` to install a plugin hosted remotely
- `from: //path/to/plugin-target` to use a plugin that's authored in the local repository

alexeagle:
When would we `bazel build` the plugin(s)? Let's always run the `bazel build` every time.
Need to make sure that build command doesn't thrash the analysis cache.
Maybe this is just a local dev convenience, and real users should fetch remote.

mattem:
@alexeagle Can this be closed, or is there further work to be done under this ticket?

alexeagle:
`from: github.com/my-org/my-plugin@1.2.3` is actually the most important part I think.

https://aspect.build/help/topics/plugins says
>In the future, we’ll add more ways to specify a plugin, such as with a remote URL. This would use semantic versioning ranges to constrain the versions which can be used. When aspect runs, it can then prompt you to re-lock the dependencies to exact versions if they have changed, and can verify the integrity of the plugin contents against what was first installed.

sluongng:
I took a look at this project today to investigate on how to create my own plugins, but I find the current plugin setup being a bit too complicated for my taste.

I prefer to have 1 single binary as the entry point, not `1 main binary + (1 binary for each plugin)`. I think there are value in having multiple plugins included in a self-contained binary. Some example of this include: telegraf, gazelle etc... The plugins can still be turned off/on via config file, but the distribution is a lot simpler (single binary) and the API are just Go interfaces instead of gRPC calls.

So I would suggest to consider a single binary system similar to gazelle's:

- Expose an `aspect_cli_binary` bazel rule for users to self-compile the whole binary (using bazel)
- Expose a `plugins` attribute in that rule where users can point to `go_library` that have an `AspectPlugin` struct export which implements all the interfaces required.
- The `aspect_cli_binary` rule constructs a new `plugins.go` file from Go template that import all the plugins, then re-export them as a go slice for the main library to consume.
- Final aspect binary is compiled using existing source and the generated `plugins.go` file.

This is exactly how Gazelle works with the `language.go` file today.

WDYT?

alexeagle:
[...]

PR #174, opened 2022-03-10
feat: load plugins from local target

alexeagle:
Addresses half of #167

### DIFF
--- a/docs/help/topics/plugins.md
+++ b/docs/help/topics/plugins.md
@@ -1,41 +1,51 @@
 # Aspect CLI Plugins
 
-NB: plugin support is still in design phase, the documentation below may not yet
-apply.
-
-The plugin system is inspired by the excellent system developed for HashiCorp's
-`terraform` CLI.
+The plugin system is based on the excellent system developed by HashiCorp for the `terraform` CLI.
 
 ## High-level design
 
 A plugin is any program with a gRPC server that implements our plugin protocol.
+
 We provide convenient support for writing plugins in Go, but this is not
-required. Plugins are hosted and versioned independently from the aspect CLI.
+required. You can write a plugin in any language.
+Plugins are hosted and versioned independently from the aspect CLI.
 
 The aspect CLI process spawns the plugin as a subprocess, then connects as a
 gRPC client to it. The client and server run a negotiation protocol to determine
 version compatibility and what capabilities the plugin provides.
 
+You can read more about this archecture here:
+<https://github.com/hashicorp/go-plugin/blob/master/README.md>
+
 ## Plugin configuration
 
-In your [aspect CLI config], list the plugins you'd like to install. You can use
-semver ranges to constrain the versions which can be used. When aspect runs, it
-will prompt you to re-lock the dependencies to exact versions if they have
-changed. We also verify the checksum of the plugin contents against what was
-first installed.
+In a `.aspectplugins` file at the repository root, list the plugins you'd like to install.
+
+This is a yaml file. The shortest example provides a name and a local path to the plugin binary:
+
+```yaml
+- name: some-plugin
+  from: ./path/to/plugin_binary
+```
+
+The `from` line may start with `//` in which case it is interpreted as a [Bazel Label] in the
+current workspace.
+That label must be a `*_binary` rule which builds a plugin binary. When the CLI loads this
+plugin, it first builds it from source.
+This is useful as a local development round-trip while authoring a plugin. However it is not a
+great way to deploy a plugin to users, as it causes them to perform an extra build every time
+they run `aspect`, whether they intend to use the plugin or not.
+
+In the future, we'll add more ways to specify a plugin, such as with a remote URL.
+This would use semver ranges to constrain the versions which can be used.
+When aspect runs, it can then prompt you to re-lock the dependencies to exact versions if they
+have changed, and can verify the checksum of the plugin contents against what was first installed.
 
 > The locking semantics follow the [Trust on first use] approach.
 
 [trust on first use]: https://en.wikipedia.org/wiki/Trust_on_first_use
-[aspect cli config]: TODO
-
-## Plugin discovery
-
-TODO: where we search to resolve plugins on disk or fetch them from network
-TODO: how to author a local plugin and resolve it for development
+[bazel label]: https://bazel.build/concepts/labels
 
 ## Capabilities
 
-Plugins can implement any of the following:
-
-- BuildComplete
+More documentation on the plugin API coming soon!
--- a/docs/help/topics/plugins.md
+++ b/docs/help/topics/plugins.md
@@ -21,7 +21,7 @@ You can read more about this archecture here:
 
 In a `.aspectplugins` file at the repository root, list the plugins you'd like to install.
 
-This is a yaml file. The shortest example provides a name and a local path to the plugin binary:
+This is a YAML file. The shortest example provides a name and a local path to the plugin binary:
 
 ```yaml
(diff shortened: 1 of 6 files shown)


docs/help/topics/plugins.md
```diff
@@ -1,41 +1,51 @@
...
 
-In your [aspect CLI config], list the plugins you'd like to install. You can use
-semver ranges to constrain the versions which can be used. When aspect runs, it
-will prompt you to re-lock the dependencies to exact versions if they have
-changed. We also verify the checksum of the plugin contents against what was
-first installed.
+In a `.aspectplugins` file at the repository root, list the plugins you'd like to install.
 
-> The locking semantics follow the [Trust on first use] approach.
+This is a yaml file. The shortest example provides a name and a local path to the plugin binary:
```
f0rmiga, line 24:
```suggestion
This is a YAML file. The shortest example provides a name and a local path to the plugin binary:
```

docs/help/topics/plugins.md
```diff
@@ -1,41 +1,51 @@
...
+- name: some-plugin
+  from: ./path/to/plugin_binary
+```
 
-## Plugin discovery
+The `from` line may start with `//` in which case it is interpreted as a [Bazel Label] in the
+current workspace.
+That label must be a `*_binary` rule which builds a plugin binary. When the CLI loads this
+plugin, it first builds it from source.
+This is useful as a local development round-trip while authoring a plugin. However it is not a
```
f0rmiga, line 35:
```suggestion
This is useful as a local development round-trip while authoring a plugin. However, it is not a
```

docs/help/topics/plugins.md
```diff
@@ -1,41 +1,51 @@
...
+great way to deploy a plugin to users, as it causes them to perform an extra build every time
+they run `aspect`, whether they intend to use the plugin or not.
 
-TODO: where we search to resolve plugins on disk or fetch them from network
-TODO: how to author a local plugin and resolve it for development
+In the future, we'll add more ways to specify a plugin, such as with a remote URL.
+This would use semver ranges to constrain the versions which can be used.
```
f0rmiga, line 40:
```suggestion
This would use semantic versioning ranges to constrain the versions which can be used.
```

docs/help/topics/plugins.md
```diff
@@ -1,41 +1,51 @@
...
+When aspect runs, it can then prompt you to re-lock the dependencies to exact versions if they
+have changed, and can verify the checksum of the plugin contents against what was first installed.
```
f0rmiga, line 42:
```suggestion
have changed, and can verify the integrity of the plugin contents against what was first installed.
```

pkg/plugin/client/client.go
```diff
@@ -38,10 +42,42 @@ type CustomCommandExecutor interface {
 	ExecuteCustomCommand(cmdName string, ctx context.Context, args []string) error
 }
 
-type clientFactory struct{}
+type clientFactory struct {
+	bzl bazel.Bazel
+}
+
+// buildPlugin asks bazel to build the target and returns the path to the resulting binary
```
f0rmiga, line 49:
```suggestion
// buildPlugin asks bazel to build the target and returns the path to the resulting binary.
```

pkg/plugin/client/client.go
```diff
@@ -38,10 +42,42 @@ type CustomCommandExecutor interface {
...
+func (c *clientFactory) buildPlugin(target string) (string, error) {
+	agc, err := c.bzl.AQuery(target)
+	if err != nil {
+		return "", err
+	}
+	outs := bazel.ParseOutputs(agc)
+
+	var pluginPath string
+	for _, a := range outs {
+		// TODO: don't hard-code GoLink, plugins could be written in other languages
```
f0rmiga, line 59: Could this be a GitHub issue instead so we can keep track and actually schedule work to address this?

pkg/plugin/client/client.go
```diff
@@ -38,10 +42,42 @@ type CustomCommandExecutor interface {
...
+		if a.Mnemonic == "GoLink" {
+			pluginPath = a.Path
+			break
+		}
+	}
+	if pluginPath == "" {
+		return "", fmt.Errorf("no output file from a GoLink action was found for target %v", target)
```
f0rmiga, line 66:
```suggestion
		return "", fmt.Errorf("failed to build plugin %q with Bazel: no output file from a GoLink action was found", target)
```

pkg/plugin/client/client.go
```diff
@@ -38,10 +42,42 @@ type CustomCommandExecutor interface {
...
+	}
+
+	// TODO: we should be careful to use the right flags for this build
+	// to avoid busting the analysis cache. We want to pretend to be a typical
+	// build the developer or CI would be performing.
+	if _, err := c.bzl.Spawn(append([]string{"build", target})); err != nil {
+		return "", fmt.Errorf("failed to build plugin %v with Bazel: %w", target, err)
```
f0rmiga, line 73:
```suggestion
		return "", fmt.Errorf("failed to build plugin %q with Bazel: %w", target, err)
```

pkg/plugin/client/client.go
```diff
@@ -38,10 +42,42 @@ type CustomCommandExecutor interface {
...
+		}
+	}
+	if pluginPath == "" {
+		return "", fmt.Errorf("no output file from a GoLink action was found for target %v", target)
+	}
+
+	// TODO: we should be careful to use the right flags for this build
+	// to avoid busting the analysis cache. We want to pretend to be a typical
+	// build the developer or CI would be performing.
+	if _, err := c.bzl.Spawn(append([]string{"build", target})); err != nil {
```
f0rmiga, line 72:
```suggestion
	if _, err := c.bzl.Spawn([]string{"build", target}); err != nil {
```

pkg/plugin/client/client.go
```diff
@@ -38,10 +42,42 @@ type CustomCommandExecutor interface {
...
+		if a.Mnemonic == "GoLink" {
+			pluginPath = a.Path
+			break
+		}
+	}
+	if pluginPath == "" {
+		return "", fmt.Errorf("no output file from a GoLink action was found for target %v", target)
+	}
+
+	// TODO: we should be careful to use the right flags for this build
```
f0rmiga, line 69: I don't think there's an actual `TODO` here. It should be a `WARNING` instead to never add flags here.

pkg/plugin/client/client.go
f0rmiga, line 73:
```suggestion
		return "", fmt.Errorf("failed to build plugin %q with Bazel: %w", target, err)
```

pkg/plugin/client/client.go
```diff
@@ -38,10 +42,42 @@ type CustomCommandExecutor interface {
...
+	bzl bazel.Bazel
+}
+
+// buildPlugin asks bazel to build the target and returns the path to the resulting binary
+func (c *clientFactory) buildPlugin(target string) (string, error) {
+	agc, err := c.bzl.AQuery(target)
+	if err != nil {
+		return "", err
+	}
+	outs := bazel.ParseOutputs(agc)
```
f0rmiga, line 55:
```suggestion
	queryOutput, err := c.bzl.AQuery(target)
	if err != nil {
		return "", err
	}
	outs := bazel.ParseOutputs(queryOutput)
```

f0rmiga (changes requested):

alexeagle:
thanks for thorough review, PTAL

f0rmiga (approved):

PR #186, opened 2022-03-16
refactor: re-organize bazel protos

### LINKED ISSUE #185: Still have editor red squggles when authoring a plugin
I checked in all the generated .pb.go files, however we still rely on a Bazel alias in /bazel/buildeventstream/proto which actually points into our third-party folder. So the Go tooling is unable to resolve the imports on `aspect.build/cli/bazel/buildeventstream/proto`

Actually that import statement looks pretty nice in user code. All our imports from third-party/some/long/path/from/bazel are the ones that look bad. I think we should move all the proto files into `aspect.build/cli/bazel` - there's no requirement that we have a folder named "third-party" nor that we lay the proto files out on disk the same way Bazel does.

alexeagle:
Fixes #185


BUILD.bazel
```diff
@@ -38,17 +38,22 @@ gazelle(
 #
 # Note that each target also has a corresponding test asserting that the
 # generated files stay up-to-date.
+# Re-generate this list with:
+# bazel 2>/dev/null query 'kind(go_proto_library, //...)' | xargs -I {} echo "        \"{}.update_go_pb\","
```
mattem, line 42: pipe to buildozer?

mattem (approved):

PR #190, opened 2022-03-20
fix: pass grpc server address without protocol

mattem:
5a5d8344c308384a69a56bdad764d8ce98d58436 changed the way the grpc address was passed around, but it seems that starting the server doesn't want the protocol appended, so it fails to start and results in the follow error for all `aspect build / test / run` commands:

```
$ bazel-bin/cmd/aspect/aspect_/aspect build aspect
Error: failed to run BES backend: failed to serve and wait BES backend: context deadline exceeded
```

Changes back to just the address sans protocol.


alexeagle:
We really need plugin integration tests soon..

alexeagle (approved):

PR #193, opened 2022-03-23
Allow startup flags to be passed through to bazel

JesseTatasciore:
image

Print statement was added to the code that invokes bazel in order to show the arguments we are passing through and their order


alexeagle (approved):
any kind of test we could/should add for this? what regressions do we imagine are possible?

JesseTatasciore:
@alexeagle Going to land this so it gets in but ill think about how / where we could add tests and make a follow up PR. As of right now I am not sure where we would want to test it. But if we are going to test for startup flags being added by bazel then we should test for the regular flags as well. I opened a ticket: https://github.com/aspect-build/aspect-cli/issues/195

alexeagle:
Yup that's fine, I'm not sure if there's much risk of regressions here anyway.

PR #202, opened 2022-04-06
[pkg][aspect][root][flags] wip handle more cases with config files. W…

dymart:
…rite if they don't exist


pkg/aspect/root/flags/interceptor.go
```diff
@@ -86,32 +86,68 @@ func FlagsInterceptor(streams ioutils.Streams) interceptors.Interceptor {
 		if cfgFile != "" {
 			// Use config file from the flag.
 			viper.SetConfigFile(cfgFile)
+			viper.AutomaticEnv()
+			if err := viper.ReadInConfig(); err == nil {
```
f0rmiga, line 90: What if reading the config fails?

pkg/aspect/root/flags/interceptor.go
```diff
@@ -86,32 +86,68 @@ func FlagsInterceptor(streams ioutils.Streams) interceptors.Interceptor {
...
+			// search in current directory first
+			if err := viper.ReadInConfig(); err != nil {
+				if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+					// search in home directory next
+					viper.AddConfigPath(home)
+					if err := viper.ReadInConfig(); err != nil {
+						if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+							// Config file not found;
+							// create new config file if it does not exist
+							if err := viper.WriteConfigAs(fmt.Sprintf("%s/.aspect.yaml", home)); err != nil {
```
f0rmiga, line 117: It's too aggressive to write files to the user's home directory without asking for permission first.

pkg/aspect/root/flags/interceptor.go
```diff
@@ -86,32 +86,68 @@ func FlagsInterceptor(streams ioutils.Streams) interceptors.Interceptor {
...
+								// if you can't write to home directory fail and write to the current directory
+								// Should we handle if this fails?
+								viper.WriteConfigAs(".aspect.yaml")
```
f0rmiga, line 120: Write relative to the working directory? I think you actually want to pass the absolute path here by using the workspace path.

pkg/aspect/root/flags/interceptor.go
```diff
@@ -86,32 +86,68 @@ func FlagsInterceptor(streams ioutils.Streams) interceptors.Interceptor {
...
+							}
+						} else {
```
f0rmiga, line 122: Empty blocks?

dymart: I wasn't sure how we wanted to handle errors in the cli. What is the preferred way?

pkg/aspect/root/flags/interceptor.go
dymart: Similar to the empty code blocks comment below, I wasn't sure how we wanted to deal with errors. Also why this pr is a draft as I don't want it merged without error handling

pkg/aspect/root/flags/interceptor.go
dymart: What is the way we would like to handle this? 
My thought was precedence would go to the file in the working directory and then to the file in the home directory. So I was going to write them in the opposite order. Is it preferred to just write to the working directory then?

pkg/aspect/root/flags/interceptor.go
dymart: I was trying to line up the earlier viper.AddConfigPath(".") and this write so that they would be in the same place. If there is a different way we would like to resolve the the workspace path to an absolute path we should initialize that earlier so the read and write looking in the same place, no?

dymart:
Still need to add tests before removing from draft

dymart:
Is there a more direct way to test for the configs?

JesseTatasciore:
> Is there a more direct way to test for the configs?

Not that I can think of. I would have added integration tests for it as well

JesseTatasciore:
Was just testing some things out with `examples/predefined-queries`. I noticed that when I answer the initial prompts, the answers are saved in the repo level config instead of the user level config. I think they belong in the user level config.

pkg/aspect/root/flags/interceptor.go
JesseTatasciore: Ide argue that we want to ask the user and then write the to home directory. We need somewhere to have config that is user specific (not repo specific). If we write everything to the repo one then peoples user preferences could get committed when they shouldnt be.

JesseTatasciore: Maybe we could add a field to the repo level config along the lines of `userConfig: foo.yaml` and then they could add `foo.yaml` to their `.gitignore` and we could save user specific configs to that file? Then it would never be committed and we dont need to add anything to the home directory? But if they dont have something like that set then we just ask permission to add to the home directory?

JesseTatasciore: Also, check the files in your HOME directory. Software add there all the time. Ive got a `.yarn`, `.npm`, `.vscode` in my mine (to name a few) and I didnt give any permission to write to my home folder. Maybe we add a `.aspect` folder where we can add anything we might need such as a config file, logs, etc

dymart: The most updated version will create a .aspect.yaml file in home directory if others aren't present. It will no longer try to write to the working directory.

If we want to have a user specific file I would stay with similar conventions to bazel and have a user.aspect.yaml and an .aspect.yaml file to make it clear which is which and to make it easy to gitignore for everyone. It would also make it easier for us as we would know the priority of the files beforehand.

The thought behind not asking for permission is to keep some determinism between invocations of the cli. It is also as you pointed out convention now to write config files to the HOME directory, I don't think asking for this permission really adds much and could make it more difficult to use in a CI or automated fashion later on. Also I believe @f0rmiga has changed his opinion from his initial comment but I'll let him clarify.

I don't think adding more complexity of managing different directories is required for this use case. If we want to revisit that for logs or something else that really needs it we definitely can.

Happy to discuss more :)

dymart:
> Was just testing some things out with `examples/predefined-queries`. I noticed that when I answer the initial prompts, the answers are saved in the repo level config instead of the user level config. I think they belong in the user level config. So maybe something here or where the query saves the prompt results needs to change.

I would like to keep this pr to just cleaning up how reading the config is handled and creating a config if one is not present. Cleaning up how writes to the config files are handled feels like a good follow up pr!

dymart:
> But if they both have the same stuff the one from the user level is used over the one from the repo level so thats good! That was broken before

I feel like we should have more of a discussion around this topic to make it very clear how things interact and document it to help with debugging and user understanding. We should also add testing for this once we decide on what to do.

dymart:
work has moved to design doc

PR #204, opened 2022-04-08
fix: Remove restriction on number of arguments passed to aspect info

JesseTatasciore:
When running a command such as `aspect info bazel-bin --color=no` both `bazel-bin` and `--color=no` are picked up as arguments by cobra. Given this we do not want to restrict it to 1 argument


alexeagle (approved):

alexeagle:
would be nice to add a test case

f0rmiga (approved):

PR #212, opened 2022-06-10
feat: multi-platform testing

f0rmiga:


f0rmiga:
Before merging this, we want to set up the remote cache.

alexeagle:
Looks like this went stale, sorry!

f0rmiga:
I guess we never set up the remote cache?

PR #213, opened 2022-06-10
Create osutils for performing OS specific operations

JesseTatasciore:


pkg/osutils/filesystem/darwin.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"time"
+)
+
+const ChmodPath = "/bin/chmod"
```
f0rmiga, line 24: I'm thinking that we want to use `chmod` instead of the absolute path. If you call just `chmod`, then we take advantage of how it resolves the `PATH` environment variable. I can see how in this case it's rarely the case that the user would want to override it via `PATH` but there's no reason why we would make such a choice of enforcing `/bin/chmod`. This is a comment about a generalized approach to this kind of situation.

pkg/osutils/filesystem/filesystem.go
```diff
@@ -0,0 +1,83 @@
...
+
+func osExecCommand(name string, arg ...string) ExecCmdRunner {
+	return exec.Command(name, arg...)
+}
+
+func osStat(name string) (fs.FileInfo, error) {
+	return os.Stat(name)
+}
+
+type Filesystem struct {
```
f0rmiga, line 51: Could you comment on the exported types based on the godoc good practices, please?

pkg/osutils/filesystem/linux.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"time"
+)
+
+const ChmodPath = "/usr/bin/chmod"
```
f0rmiga, line 24: Same

f0rmiga (approved):

PR #217, opened 2022-07-05
Pass a context struct to RunCommand

shs96c:
This makes it possible to set the location of the workspace
directory, and pass in environment variables if needed.

The Target Determinator in `bazel-contrib` needs us to be
able to set the workspace root, and repinning deps when
using `rules_jvm_external` requires us to set an environment
variable.

By passing a context object in, we can add additional
parameters should we find these necessary.


shs96c:
The need for this was highlighted when we tried to hook the Target Determinator into the Aspect CLI. We ended up building [an abstraction for executing Bazel itself](https://github.com/bazel-contrib/target-determinator/commit/8fb39d63b864759662a71fab429ec259363360c6) but found that there was no way of changing the workspace directory.

shs96c:
Closing this, as my employer requests that all CLAs are approved by legal, and getting that approval is a time-consuming process. Will file an issue.

PR #221, opened 2022-07-13
feat: avoid re-entrance

alexeagle:
If the aspect-build/bazel fork is chosen by the first line in .bazelversion, then aspect-cli will try to call itself.
Instead it can provide the second line of the file in this case when calling our nested bazelisk

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -78,9 +78,17 @@ Companies with more than 250 employees can contact us at sales@aspect.dev for a 
 
 # Installation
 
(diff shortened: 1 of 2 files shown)


dymart (approved):

alexeagle:
Manually tested that this works, by publishing a fake release at https://github.com/aspect-build/bazel/releases/tag/0.5.1
and then using .bazelversion file 

```
aspect-build/0.5.1
5.0.0
```

PR #232, opened 2022-09-06
bump: go deps

f0rmiga:


alexeagle (approved):
lgtm when not red :)

f0rmiga:
> lgtm when not red :)

I prefer that too. :)

alexeagle (approved):

PR #247, opened 2022-09-13
Implement `bootstrap version`

### LINKED ISSUE #235: Implement `version` command for `aspect-bootstrap`
Goal: Print the current version of `aspect-bootstrap`.

cgrindel:
Closes #235.

- Update `buildinfo` to provide `BuildInfo` struct with the stamped values.
  - Providing the information as a struct makes it much easier to test certain conditions in downstream code.
- Refactor common version writing code to `pkg/versionwriter`.
- Update `aspect version` code to use `pkg/versionwriter`.


cgrindel:
I converted it back to a Draft so that it is not accidentally merged before #246.

buildinfo/buildinfo.go
```diff
@@ -30,3 +34,45 @@ var Release = "no release"
...
+		BuildTime,
+		HostName,
+		GitCommit,
+		GitStatus,
+		Release,
+	)
+}
+
+func (bi BuildInfo) HasRelease() bool {
+	return bi.Release != ""
```
alexeagle, line 73: nit: the conditional logic in this function and the one below are so trivial, and not expected to grow. It doesn't feel worth it to me to write a unit test, and then have to add 40 lines of code to make this file mockable

cmd/bootstrap/main.go
```diff
@@ -1,7 +1,16 @@
...
+	"os"
+
+	"aspect.build/cli/cmd/bootstrap/root"
+)
 
 func main() {
-	fmt.Println("Hello, World!")
+	rootCmd := root.NewRootCmd()
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Fprintf(os.Stderr, "Whoops. There was an error while executing your CLI '%s'", err)
```
alexeagle, line 13: is this a "you did it wrong" or a "file a bug" message? I think the latter, this should never happen regardless of user input?

pkg/versionwriter/version_writer.go
```diff
@@ -0,0 +1,63 @@
...
+
+const (
+	GNU Format = iota
+	Conventional
+
+	NotCleanVersionSuffix = " (with local changes)"
+	NoReleaseVersion      = "unknown [not built with --stamp]"
+)
+
+type VersionWriter struct {
```
alexeagle, line 21: feels to me like this whole package is just a fancy "toString()" for a BuildInfo - maybe these functions could live there instead?

buildinfo/buildinfo.go
cgrindel: I created `HasRelease()`, because I think that we may have two conditions to check. (I should have added the other. My bad.). One condition is not empty. The other is not equal to our default (i.e., unstamped) value, `no release`. 

For `IsClean()`, I think it is more compact and pretty clear. For example, a client would write the following to check for the clean state:

```go
if bi.IsClean() {
  // awesome code
}
```

If we do not have it, then they write the following:

```go
if bi.GitStatus == buildinfo.CleanGitStatus {
  // awesome code
}
```

Also, in the off chance that we change how we are tracking the state, the client code does not need to be touched.

I am not sure when we/if we will ever need to mock this struct. Test code should be able to create an instance without too much trouble using the `buildinfo.New()`. Do you have a test scenario in mind?

alexeagle: I don't mean to pick on this one too much, I'm more trying to give guidance on the whole, that we're a scrappy unfunded startup so I'm more worried about the failure more where we don't make it to market than the one where some conditional logic doesn't have a unit test.

What you have here is fine, just wanted to make this point :)

cmd/bootstrap/main.go
cgrindel: I agree. Do we have standard language for these errors? I will update this accordingly.

pkg/versionwriter/version_writer.go
cgrindel: I struggled with this. I ended up with a separate package because we were carrying around the `Format` state that is coming from a flag. Let me see what the code looks like if I move it to `buildinfo.BuildInfo`.

cmd/bootstrap/main.go
alexeagle: Something about it being non-user-caused and what to do about it. Maybe sth like "Internal Error, please file an issue, %v"

Looking around the root of the aspect command, I don't see a similar error message to copy from though, it just prints ERROR and exits 1

cgrindel: Ok. I will add a template for this and use it.

alexeagle: I'm sure our Go expert @f0rmiga could point to some idiom for this, seems like aspecterrors/errors.go could provide some InternalError or something.

cgrindel: That is what I was thinking.

pkg/versionwriter/version_writer.go
cgrindel: I moved the logic to `buildinfo`.

cmd/bootstrap/version/version.go
```diff
@@ -0,0 +1,54 @@
...
+
+func NewVersionCmd() *cobra.Command {
+	gnuFormat := false
+	cmd := &cobra.Command{
+		Use:   "version",
+		Short: "Print the version of aspect CLI as well as tools it invokes.",
+		Long:  `Prints version info on colon-separated lines, just like bazel does`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			bi := buildinfo.Current()
+			// format := versionwriter.Conventional
```
alexeagle, line 26: remove the commented code?

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
 	github.com/spf13/cobra v1.5.0
 	github.com/spf13/viper v1.13.0
+	github.com/stretchr/testify v1.8.0
```
alexeagle, line 19: I think this is overlapping a lot with gomega? I think we want to avoid extra 3p deps where possible. Is it easy to just use that instead like existing tests?

buildinfo/build_info.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (bi BuildInfo) HasRelease() bool {
+	return bi.Release != "" && bi.Release != PreStampRelease
+}
+
+func (bi BuildInfo) IsClean() bool {
+	return bi.GitStatus == CleanGitStatus
+}
+
+func (bi BuildInfo) Version() string {
```
alexeagle, line 78: I'm a little confused by naming, what does "Utility" mean - seems like this function is more like Release() and the one below is Version()

alexeagle (commented):
thanks, looks close :)

buildinfo/build_info.go
cgrindel: There is already a field called `Release`. It represents the stamped release. The `Version()` function applies logic that decorates the release info. I went ahead and renamed `UtilityVersion` to `CommandVersion`. It outputs the command-line application's version which includes the name of the application along with the `Version()`.

go.mod
cgrindel: I should have noticed that we were using a different assertion library. Done.

cmd/bootstrap/main.go
cgrindel: To avoid blowing up this PR, I add #251. In the meantime, I added `aspecterrors.HandleError()` and updated `aspect` and `bootstrap` to use it for their top-level error handling.

cgrindel:
@alexeagle I believe that I have addressed all of your comments. When you get a chance, please take a look. Thanks!

buildinfo/stamp.go
```diff
@@ -1,17 +1,9 @@
 /*
  * Copyright 2022 Aspect Build Systems, Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the aspect.build Commercial License (the "License");
```
alexeagle, line 4: oops, merge conflict

cmd/bootstrap/version/version.go
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2022 Aspect Build Systems, Inc.
+ *
+ * Licensed under the aspect.build Commercial License (the "License");
```
alexeagle, line 4: could you fix this to match other files - we just relicensed

buildinfo/stamp.go
cgrindel: Done

cmd/bootstrap/version/version.go
cgrindel: Done.

alexeagle (approved):

Issue #252, opened 2022-09-14
Update configuration file search to look in the proper locations


cgrindel:
Design document: https://hackmd.io/GKLazSQ3RU-aagWQoJJicg?view

## Configuration File Locations

There will only be four places to set the configuration of the Aspect CLI. In order of precedence from highest to lowest (highest precedence overriding lowest).

- Individual flags pass to the CLI `--aspect:some_flag=some_value`
- Pass a custom config file location directly to the CLI ex. `--aspectconfig=.aspect/config.yaml`
- The users home directory `$HOME/.aspect/config.yaml`
- Base of the workspace `$WORKSPACE/.aspect/config.yaml`

PR #254, opened 2022-09-15
Fix error when running a command that attempts to save config file when it does not exist

### LINKED ISSUE #245: Error when running `bazel clean` and choosing to remember my choice
## Steps
1. Be sure not to have a `$HOME/.aspect` file.
2. Change to a directory with the `aspect-cli` repository.
3. Execute `bazel clean`.

The following is the output:

```sh
✔ Reclaim disk space for this workspace (same as bazel clean)
You can skip this prompt to make 'aspect clean' behave the same as 'bazel clean'
Remember this choice and skip the prompt in the future: y
Error: failed to update config file: Config File ".aspect" Not Found in "[/Users/chuck]"
```

Without having looked at the code, I suspect that we need to create the file if it does not exist.

cgrindel:
Closes #245.


pkg/configutils/Write.go
```diff
@@ -0,0 +1,31 @@
+/*
```
alexeagle, line 1: we don't have any other go sources where the filename is title case like this

pkg/aspect/root/flags/interceptor.go
```diff
@@ -102,10 +102,12 @@ func FlagsInterceptor(streams ioutils.Streams) interceptors.Interceptor {
 			// Search for config in home directory with name ".aspect" (without extension).
```
alexeagle, line 102: comment now out-of-date?

pkg/configutils/Write.go
cgrindel: Doh.

pkg/aspect/root/flags/interceptor.go
cgrindel: I am not fixing that in this PR. We will fix with #252.

alexeagle (approved):

PR #258, opened 2022-09-16
refactor: load plugins config from under .aspect folder

alexeagle:
We now have other files in that location and want to keep them organized together.

### DIFF
--- a/pkg/plugin/loader/loader.go
+++ b/pkg/plugin/loader/loader.go
@@ -29,7 +29,7 @@ import (
 
 const (
 	workspaceFilename     = "WORKSPACE"
-	aspectpluginsFilename = ".aspectplugins"
+	aspectpluginsFilename = ".aspect/cli/plugins.yaml"
 )
 
 // AspectPlugin represents a plugin entry in the plugins file.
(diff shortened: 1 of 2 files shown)


pkg/plugin/loader/loader.go
```diff
@@ -75,15 +75,15 @@ func (f *finder) Find() (string, error) {
...
+				return "", fmt.Errorf("failed to find Bazel workspace: %w", err)
 			}
 			cwd = filepath.Dir(cwd)
 			continue
 		}
 		aspectpluginsPath := path.Join(cwd, aspectpluginsFilename)
 		if _, err := f.osStat(aspectpluginsPath); err != nil {
 			if !os.IsNotExist(err) {
-				return "", fmt.Errorf("failed to find .aspectplugins: %w", err)
+				return "", fmt.Errorf("failed to find .aspect/cli/plugins.yaml: %w", err)
```
gregmagolan, line 86: nit: you could use the `aspectpluginsFilename` constant here instead of hard-coding the filename in the error msg

gregmagolan: but we're planning to do another pass of this and consolidate cli/plugins.yaml and cli/config.yaml anyway so no point in perfecting it

gregmagolan (approved):

PR #260, opened 2022-09-16
refactor: downloaded plugins are always versioned

alexeagle:
Note, we'll probably want to support `version: latest` as well. This will need
- a way for us to keep an index.json up-to-date in the S3 bucket
- logic here to look for the index file to know what versions exist
- probably also matching logic to find latest github release
- write back to the plugins config file to pin to what we discovered

### DIFF
--- a/.aspectplugins
+++ b/.aspectplugins
@@ -2,3 +2,8 @@
 #- name: hello-world
 #  version: v0.2.0
 #  from: github.com/aspect-build/aspect-cli-plugin-template
+
+# Example: Aspect's pro plugin
+# - name: plugin-aspect-pro
+#   from: https://static.aspect.build/cli
+#   version: v0.8.0
--- a/pkg/plugin/client/client.go
+++ b/pkg/plugin/client/client.go
@@ -112,16 +112,23 @@ func (c *clientFactory) New(aspectplugin loader.AspectPlugin, streams ioutils.St
 			aspectplugin.From = built
 		}
 	} else if strings.Contains(aspectplugin.From, "/") {
-		if strings.HasPrefix(aspectplugin.From, "github.com/") {
-			if len(aspectplugin.Version) < 1 {
-				return nil, fmt.Errorf("when using a plugin released on GitHub, the version field is required")
-			}
-			// Example release URL:
-			// https://github.com/aspect-build/aspect-cli-plugin-template/releases/download/v0.1.0/plugin-plugin-linux_amd64
-			aspectplugin.From = fmt.Sprintf("https://%s/releases/download/%s/%s", aspectplugin.From, aspectplugin.Version, aspectplugin.Name)
+		if len(aspectplugin.Version) < 1 {
+			return nil, fmt.Errorf("failed to download plugin '%s': the version field is required", aspectplugin.Name)
 		}
+		// Syntax sugar:
+		//   from: github.com/org/repo
+		// is the same as
+		//   from: https://github.com/org/repo/releases/download
+		// Example release URL:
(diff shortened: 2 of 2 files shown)


f0rmiga:
I don't think we should ever have the "latest" feature. See Docker, it's a mess. It brings more problems than it solves.

.aspectplugins
```diff
@@ -2,3 +2,8 @@
 #- name: hello-world
 #  version: v0.2.0
 #  from: github.com/aspect-build/aspect-cli-plugin-template
+
+# Example: Aspect's pro plugin
+# - name: plugin-aspect-pro
+#   from: http://static.aspect.build
```
f0rmiga, line 8:
```suggestion
#   from: https://static.aspect.build
```

f0rmiga: I actually think we should throw a warning message if it's http. It's super easy to impersonate a domain and inject malicious code.

pkg/plugin/client/client.go
```diff
@@ -112,16 +112,23 @@ func (c *clientFactory) New(aspectplugin loader.AspectPlugin, streams ioutils.St
...
-			// Example release URL:
-			// https://github.com/aspect-build/aspect-cli-plugin-template/releases/download/v0.1.0/plugin-plugin-linux_amd64
-			aspectplugin.From = fmt.Sprintf("https://%s/releases/download/%s/%s", aspectplugin.From, aspectplugin.Version, aspectplugin.Name)
+			aspectplugin.From = fmt.Sprintf("https://%s/releases/download", aspectplugin.From)
 		}
+		// Example release URL:
+		// http://static.aspect.build/v0.1.0/plugin-aspect-pro-darwin_amd64
 		if strings.HasPrefix(aspectplugin.From, "http://") || strings.HasPrefix(aspectplugin.From, "https://") {
-			downloaded, err := DownloadPlugin(aspectplugin.From, aspectplugin.Name)
+			versionedUrl := fmt.Sprintf("%s/%s/%s", aspectplugin.From, aspectplugin.Version, aspectplugin.Name)
```
f0rmiga, line 130: I just noticed that we are not using https://pkg.go.dev/net/url to construct the urls. This package does some better handling than constructing by hand with Sprintf.

f0rmiga (changes requested):

gregmagolan:
> I don't think we should ever have the "latest" feature. See Docker, it's a mess. It brings more problems than it solves.

Better than latest is a prompt in the CLI that looks something like "there is a newer version available of this plugin, run `command` to upgrade"?

f0rmiga:
That's one option.

.aspectplugins
alexeagle: We don't have SSL setup for the S3 bucket. Adding one looks quite non-trivial, as with anything AWS there are several other services to configure perhaps? Or we could have cloudflare as the SSL termination point and have it proxy the traffic but of course that adds another point of failure and latency

alexeagle: > The reason for all of this seemingly unnecessary complexity is that CloudFront can handle SSL connections with a custom domain, but S3 cannot. (Perhaps, someday, AWS will get around to building a simpler solution and this entire post will feel quaint.)

gregmagolan:
Actually, thinking about this some more. We should really just prompt for "There is a newer version of the CLI available. Run `foobar` to upgrade"

And for every release we specify which versions of pro plugins are used

.aspectplugins
f0rmiga: Yeah, let's do this in another PR.

f0rmiga (approved):

gregmagolan (approved):

PR #262, opened 2022-09-18
feat: add analyze profile pass-through command"

gregmagolan:
- chore: sort plugins list in root.go
- feat: add analyze-profile pass-through command

Follow up to https://github.com/aspect-build/aspect-cli/pull/261. Look there for more context.

### DIFF
--- a/cmd/aspect/analyzeprofile/analyzeprofile.go
+++ b/cmd/aspect/analyzeprofile/analyzeprofile.go
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2022 Aspect Build Systems, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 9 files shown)


cmd/aspect/analyzeprofile/analyzeprofile.go
```diff
@@ -0,0 +1,54 @@
...
+	cmd := &cobra.Command{
+		Use:   "analyze-profile",
+		Short: "Analyzes build profile data.",
+		Long: `Analyzes build profile data for the given profile data files.
+
+Analyzes each specified profile data file and prints the results.  The
+input files must have been produced by the 'bazel build
+--profile=file' command.
+
+By default, a summary of the analysis is printed.  For post-processing
```
alexeagle, line 42: while we are here - I could imagine `analyze-profile` being amazingly more helpful. This is where we should insert a bunch of our "we understand what the problem is". Maybe worth brainstorming that into some issues

alexeagle (approved):

alexeagle:
Can we amend the documentation to also fill in the "how do I get one of these profiles"?

With no arguments, could we analyze the "previous profile" - whatever is in `$(bazel info output_base)/command.profile.gz` ?

gregmagolan:
> Can we amend the documentation to also fill in the "how do I get one of these profiles"?

sgtm. I can add a note to the docstring
 
> With no arguments, could we analyze the "previous profile" - whatever is in `$(bazel info output_base)/command.profile.gz` ?

lets leave anything that isn't pass-through behaviour to follow-up; will keep those changes isolated and be easier to review

gregmagolan:
Hmmm. The current docstring from Bazel already has a note that a profile can be generated with `bazel build
--profile=file`. In the interest of time, I'll leave that and we can improve on it in a follow-up.

PR #263, opened 2022-09-19
feat: add canonicalize-flags pass-through command

gregmagolan:


cmd/aspect/canonicalizeflags/canonicalizeflags.go
```diff
@@ -0,0 +1,48 @@
...
+	return NewCanonicalizeFlagsCmd(ioutils.DefaultStreams)
+}
+
+func NewCanonicalizeFlagsCmd(streams ioutils.Streams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "canonicalize-flags",
+		Short: "Canonicalizes a list of bazel options.",
+		Long: `This command canonicalizes a list of bazel options.
+
+Don't forget to prepend  '--' to end option parsing before the flags to canonicalize..`,
```
alexeagle, line 38: ends with double period

alexeagle (approved):

cmd/aspect/canonicalizeflags/canonicalizeflags.go
gregmagolan: Nice catch.

PR #265, opened 2022-09-19
feat: add fetch pass-through command

gregmagolan:


pkg/aspect/fetch/fetch.go
```diff
@@ -0,0 +1,53 @@
...
+
+package fetch
+
+import (
+	"context"
+
+	"aspect.build/cli/pkg/bazel"
+	"aspect.build/cli/pkg/ioutils"
+	"github.com/spf13/cobra"
+
```
cgrindel, line 25: Is this blank line purposeful?

cgrindel (approved):

pkg/aspect/fetch/fetch.go
gregmagolan: Not intentional. That blank line can be removed. Looks like the go formatter wants to separate the built-in imports from external ones but it doesn't automatically collapse empty lines between external ones, treating each block a independently sorted from the looks of it.

PR #266, opened 2022-09-19
feat: add modquery pass-through command

gregmagolan:


alexeagle (approved):
wow what a crazy one. We get to learn the corners of the Bazel CLI for sure

gregmagolan:
> wow what a crazy one. We get to learn the corners of the Bazel CLI for sure

we'll probably need to gate on the underlying bazel version as well for commands like this; will leave that for a follow-up

pkg/aspect/modquery/modquery.go
```diff
@@ -0,0 +1,53 @@
...
+
+package modquery
+
+import (
+	"context"
+
+	"aspect.build/cli/pkg/bazel"
+	"aspect.build/cli/pkg/ioutils"
+	"github.com/spf13/cobra"
+
```
cgrindel, line 25: nit: Blank line

cgrindel (approved):

pkg/aspect/modquery/modquery.go
gregmagolan: 👍

PR #270, opened 2022-09-19
feat: add license pass-through command

gregmagolan:
This one needs thought and design since we probably want to include our license in the list as well and maybe not the bazel license at all since bazel is a tool that the aspect CLI uses under the hood. Will leave in DRAFT for now.


gregmagolan:
Will not do. `license` will be a first-class command.

PR #279, opened 2022-09-21
feat: add support for specifying default plugins sources in PluginSystem

gregmagolan:
Unclear if we'll need this so leaving it as DRAFT for now


alexeagle (approved):

gregmagolan:
Will no do

PR #280, opened 2022-09-21
refactor: bazel

cgrindel:
- Refactor `bazel.bazel`
  - Remove `workspace.Finder`.
  - Add `workspaceRoot` to store the path to the root of the Bazel workspace.
- Update `bazel.New()` to take a `workspaceRoot` parameter.
- Add `bazel.Find()` and `bazel.FindFromWd()` to find a workspace root and return an initialized `Bazel` instance.
- Refactor `Bazel` interface
  - Remove `WithOverrideWorkspaceRoot()`. Instances can now be created using a path.
  - Remove `Spawn()`. Appears to be an alias for `RunCommand()`.
  - Update `WithEnv()` to not create a new instance when applying the environment variables.

### DIFF
--- a/pkg/bazel/bazel.go
+++ b/pkg/bazel/bazel.go
@@ -90,23 +90,6 @@ func (b *bazel) WithEnv(env []string) Bazel {
 	return b
 }
 
-// // WithOverrideWorkspaceRoot returns a new instance of Bazel, overriding the
-// // value for the workspace root.
-// func (b *bazel) WithOverrideWorkspaceRoot(workspaceRoot string) Bazel {
-// 	newBazel := *b
-// 	newBazel.workspaceFinder = nil
-// 	newBazel.overrideWorkspaceRoot = workspaceRoot
-// 	return &newBazel
-// }
-
-// // WithEnv returns a new instance of Bazel, setting the provided environment
-// // variables.
-// func (b *bazel) WithEnv(env []string) Bazel {
-// 	newBazel := *b
-// 	newBazel.env = env
-// 	return &newBazel
-// }
-
 // AvailableStartupFlags will return the full list of startup flags available for
 // the current version of bazel. This is NOT the list of startup flags that have been
 // set for the current run via SetStartupFlags.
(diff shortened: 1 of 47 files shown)


pkg/bazel/bazel.go
```diff
@@ -28,7 +28,6 @@ import (
 	"aspect.build/cli/bazel/flags"
 	"aspect.build/cli/pkg/bazel/workspace"
 	"aspect.build/cli/pkg/ioutils"
-
```
f0rmiga, line 31: Unnecessary change. We actually like that first-party imports are a separate block from third-party block.

pkg/bazel/bazel.go
```diff
@@ -110,29 +114,13 @@ func (*bazel) createRepositories() *core.Repositories {
 	return core.CreateRepositories(gcs, gcs, gitHub, gcs, gcs, true)
 }
 
-// Spawn is similar to the main() function of bazelisk
-// see https://github.com/bazelbuild/bazelisk/blob/7c3d9d5/bazelisk.go
-func (b *bazel) Spawn(command []string, streams ioutils.Streams) (int, error) {
```
f0rmiga, line 115: Interesting that we had this. :)
Thanks for cleaning up.

pkg/bazel/workspace/not_found_error.go
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package workspace
+
+import (
+	"fmt"
+)
+
+type NotFoundError struct {
```
f0rmiga, line 23: Would you mind adding godoc comments to these?

f0rmiga (approved):
It's a good improvement.

pkg/bazel/bazel.go
cgrindel: Reverted.

pkg/bazel/workspace/not_found_error.go
cgrindel: Done.

PR #292, opened 2022-10-05
fix: gather and set start-up flags in before configuring the plugin system

gregmagolan:
Also store all flags as a global so we only ever call `bazel help flags-as-proto` once per aspect cli invocation

### DIFF
--- a/pkg/bazel/bazel.go
+++ b/pkg/bazel/bazel.go
@@ -37,7 +37,7 @@ import (
 // This is for performance, avoiding a lookup of the possible startup flags for every
 // instance of a bazel struct.
 // We know the flags will be constant for the lifetime of an `aspect` cli execution.
-var availableStartupFlags []string
+var allFlags map[string]*flags.FlagInfo
 
 // Global mutable state!
 // This is for performance, avoiding needing to set the specified startup flags for every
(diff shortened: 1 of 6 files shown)


pkg/bazel/bazel.go
```diff
@@ -156,20 +145,40 @@ func (b *bazel) Flags() (map[string]*flags.FlagInfo, error) {
...
+		if isStartup {
+			startupFlags = append(startupFlags, arg)
+		} else {
+			argsWithoutStartupFlags = append(argsWithoutStartupFlags, arg)
+		}
+	}
+
+	os.Args = argsWithoutStartupFlags
+
+	return allFlags, nil
```
f0rmiga, line 181: If you return 3 arguments, you don't need this hack with `os.Args` from this hidden place, instead, move the assignment to the `main` function, which is the right place for controversial things to leave.
```suggestion
	return allFlags, argsWithoutStartupFlags, nil
```

gregmagolan: Did some refactoring to make this cleaner

pkg/bazel/bazel.go
```diff
@@ -156,20 +145,40 @@ func (b *bazel) Flags() (map[string]*flags.FlagInfo, error) {
...
+				if allFlags[*flagCollection.FlagInfos[i].Name].GetHasNegativeFlag() {
 					availableStartupFlags = append(availableStartupFlags, "no"+*flagCollection.FlagInfos[i].Name)
 				}
 			}
 		}
 	}
 
-	return flags, nil
+	argsWithoutStartupFlags := []string{os.Args[0]}
+
```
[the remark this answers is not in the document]
gregmagolan: Did some refactoring to make this cleaner

f0rmiga (approved):

PR #293, opened 2022-10-05
fix: use String flags for bazel boolean flags

gregmagolan:


f0rmiga (approved):
IIRC, it doesn't even accept `--bool-flag=false`.

PR #296, opened 2022-10-07
Add support for Bazel boolean flags.

cgrindel:
Closes aspect-build/silo#434.

### DIFF
--- a/pkg/aspect/root/flags/noable_bool.go
+++ b/pkg/aspect/root/flags/noable_bool.go
@@ -19,6 +19,7 @@ package flags
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/spf13/pflag"
 )
@@ -26,26 +27,40 @@ import (
 func RegisterNoableBool(flags *pflag.FlagSet, name string, value bool, usage string) *bool {
 	result := value
 
-	trueNB := noableBool{
-		value:        &result,
-		valueWhenSet: true,
-	}
-	flags.Var(&trueNB, name, usage)
+	nb := &noableBool{value: &result}
 
-	falseNB := noableBool{
-		value:        &result,
-		valueWhenSet: false,
+	flag := &pflag.Flag{
+		Name:      name,
+		Shorthand: "",
+		Usage:     usage,
+		Value:     nb,
+		DefValue:  nb.String(),
+		// The value that will be passed to Set() if no other values are specified.
+		NoOptDefVal: "true",
 	}
-	flags.Var(&falseNB, "no"+name, usage)
+	flags.AddFlag(flag)
+
+	noFlag := &pflag.Flag{
+		Name:      "no" + name,
+		Shorthand: "",
+		Usage:     usage,
+		Value:     nb,
+		DefValue:  nb.String(),
+		// The value that will be passed to Set() if no other values are specified.
+		NoOptDefVal: "false",
+	}
+	flags.AddFlag(noFlag)
 
 	return &result
+}
+
+func boolStr(value bool) string {
+	return fmt.Sprintf("%t", value)
 }
 
 type noableBool struct {
 	// The address of the actual value.
 	value *bool
-	// The value that should be set when the Set() function is called.
-	valueWhenSet bool
 }
 
 func (nb *noableBool) Type() string {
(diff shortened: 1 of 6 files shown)


pkg/aspect/root/flags/noable_bool_test.go
```diff
@@ -0,0 +1,50 @@
...
+	g := NewWithT(t)
+	// From Bazel doc
+	// https://bazel.build/reference/command-line-reference#option-syntax
+	doBoolFlagTest(g, false, true, "--foo")
+	doBoolFlagTest(g, true, false, "--nofoo")
+	doBoolFlagTest(g, false, true, "--foo=yes")
+	doBoolFlagTest(g, true, false, "--foo=no")
+	doBoolFlagTest(g, false, true, "--foo=1")
+	doBoolFlagTest(g, true, false, "--foo=0")
+	doBoolFlagTest(g, false, true, "-f")
```
f0rmiga, line 49: Missing failing test for invalid value.

pkg/aspect/root/flags/noable_bool.go
```diff
@@ -0,0 +1,102 @@
...
+	flags.AddFlag(noFlag)
+
+	return &result
+}
+
+func boolStr(value bool) string {
+	return fmt.Sprintf("%t", value)
+}
+
+type noableBool struct {
```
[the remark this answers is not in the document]
cgrindel: The name of the flag is stored in the `pflag.Flag` struct. The `noableBool` is just the `Value` implementation.

pkg/aspect/root/flags/noable_bool.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (nb *noableBool) Set(value string) error {
+	switch strings.ToLower(value) {
+	case "true", "yes", "1":
+		*nb.value = true
+	case "false", "no", "0":
+		*nb.value = false
+	default:
+		return fmt.Errorf("invalid bool value %s", value)
```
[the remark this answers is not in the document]
cgrindel: The framework handles it.

pkg/aspect/root/flags/noable_bool_test.go
cgrindel: Good catch. I added negative tests. Also, I reworked the logic to ensure that noXXX flags can only accept true.

pkg/aspect/root/flags/noable_bool.go
```diff
@@ -0,0 +1,133 @@
...
+	BoolFlagNo    = "no"
+	BoolFlag1     = "1"
+	BoolFlag0     = "0"
+)
+
+func RegisterNoableBool(flags *pflag.FlagSet, name string, value bool, usage string) *bool {
+	return RegisterNoableBoolP(flags, name, "", value, usage)
+}
+
+// Register a boolean flag that supports the Bazel option parsing.
```
gregmagolan, line 39: nit: mention the implementation is case-insensitive so values such as `--foo=YES` and `--foo=True` are valid

pkg/aspect/root/flags/noable_bool_test.go
gregmagolan: Some case insensitive tests would be good as well: `--foo=YES` `--foo=True` etc...

cgrindel: Done

pkg/aspect/root/flags/noable_bool.go
cgrindel: Done

cgrindel:
@gregmagolan @f0rmiga I believe that I have made all of the updates to the PR. Could you take another look?

pkg/aspect/root/flags/noable_bool.go
```diff
@@ -0,0 +1,137 @@
...
+	BoolFlagNo    = "no"
+	BoolFlag1     = "1"
+	BoolFlag0     = "0"
+)
+
+func RegisterNoableBool(flags *pflag.FlagSet, name string, value bool, usage string) *bool {
+	return RegisterNoableBoolP(flags, name, "", value, usage)
+}
+
+// Register a boolean flag that supports Bazel option parsing.
```
f0rmiga, line 39: Nit: not a valid godoc comment.

f0rmiga (approved):

pkg/aspect/root/flags/noable_bool.go
cgrindel: The automerge happened. However, I am willing to fix the issue. I don't see what the problem is with the comment. What would you recommend that I change?

f0rmiga: ```// RegisterNoableBoolP registers a boolean flag ...```

I wonder why our current linters are not catching those. Maybe I missed some configuration.

f0rmiga: I also changed the merge rules to require that comments are resolved before auto-merge happens.

PR #298, opened 2022-10-08
feat: add `//install` target for installing the Aspect CLI on the system

cgrindel:
Related to aspect-build/silo#329

This target will be used to build and install the Aspect CLI for the Homebrew formula.

### DIFF
--- a/install/BUILD.bazel
+++ b/install/BUILD.bazel
@@ -1,0 +1,8 @@
+sh_binary(
+    name = "install",
+    srcs = ["install.sh"],
+    data = [
+        "//cmd/aspect",
+        "@bazel_tools//tools/bash/runfiles",
+    ],
+)
--- a/install/BUILD.bazel
+++ b/install/BUILD.bazel
@@ -6,3 +6,15 @@ sh_binary(
         "@bazel_tools//tools/bash/runfiles",
     ],
 )
+
+sh_test(
+    name = "install_test",
+    srcs = ["install_test.sh"],
+    data = [
+        ":install",
+    ],
+    deps = [
+        "@aspect_bazel_lib//shlib/lib:assertions",
+        "@bazel_tools//tools/bash/runfiles",
+    ],
+)
--- a/install/install.sh
+++ b/install/install.sh
@@ -94,7 +94,7 @@ dest_path="${bin_dir}/aspect"
 
 # Remove an existing file
 if [[ -e "${dest_path}" ]]; then
-  warn "Removing existing file ${dest_path}."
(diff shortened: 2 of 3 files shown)


alexeagle (commented):
I'm confused, isn't it the job of the homebrew formula to install the program? why would a program need to know how to install itself? Is there some pattern you're following here, and could you link to that?
For example bazelisk is on homebrew but doesn't have an install command, why are we different?

cgrindel:
Homebrew wants you to describe how to build and install your application. Bazelisk does it by using `go build` [formula line](https://github.com/Homebrew/homebrew-core/blob/master/Formula/bazelisk.rb#L29). I opted to use Bazel to do the build. So, I needed a way to get the binary to a specified location. The formula will provide it to us. 

In terms of other patterns, [Bazel](https://github.com/Homebrew/homebrew-core/blob/master/Formula/bazel.rb) has shell scripts that build and install artifacts.

cgrindel:
One thing that may not be clear from the PR is that the install script when run by Homebrew will copy the files to the Homebrew location not /use/local/bin. I added the default so that one could test the install locally.

cgrindel:
FYI. [This](https://github.com/aspect-build/homebrew-aspect/pull/1) is what the Homebrew formula looks like, right now.

alexeagle (approved):
I see, I'm sorry, I misunderstood that install was a new command under aspect, but I see that it's just a standalone bash program.

This will be a good place to write an initial .aspect/cli/plugins.yaml file as well, if we convince Greg that the OSS CLI should be brew install'ed and the user experience isn't harmed by prompting and downloading the pro plugin later.

install/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
+sh_binary(
```
alexeagle, line 1: I don't think it should be a new "install" package at the top-level of silo, it's specific to the CLI right? I'd expect it to be in /cli/install

cgrindel: It should. I started this PR, before we moved the source of truth. Perhaps I will close this PR and open it in silo.

cgrindel:
Moving this PR to `silo`.

alexeagle:
I'm sorry, I was in a hurry with that comment as well and didn't check what repo this PR is against.

I think it's great to have the PR here actually, because we can use it to rehearse how the copybara process works when external contributors send us fixes.

install/BUILD.bazel
alexeagle: My mistake, this isn't a PR against the silo repo

PR #305, opened 2022-10-11

alexeagle:
Commit range https://github.com/aspect-build/silo/compare/824241bec6913b4742db2160f15bd90e3bed425a..e2b313f6037ad1bb10151f077df468c4c096c5f0

### DIFF
--- a/cmd/aspect/clean/BUILD.bazel
+++ b/cmd/aspect/clean/BUILD.bazel
@@ -11,7 +11,6 @@ go_library(
         "//pkg/bazel",
         "//pkg/interceptors",
         "//pkg/ioutils",
-        "@com_github_mattn_go_isatty//:go-isatty",
         "@com_github_spf13_cobra//:cobra",
     ],
 )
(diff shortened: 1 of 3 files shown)


alexeagle:
This PR created by running Copybara from the silo repo, using the copybara config added in https://github.com/aspect-build/silo/pull/451

cgrindel (approved):

alexeagle:
I think we should have copybara produce a PR description more like "sync code from silo" rather than just copying from the first commit, WDYT?

cgrindel:
Could it be `chore(copybara): ` followed by the original PR title?

alexeagle:
There is no "original PR" since this brings across a chunk of the commit history to fast-forward the OSS repo. It comes from many PRs in silo.

(seems like we have a difference in mental model where you generally think in terms of squashes where 1 PR always lands as 1 commit)

alexeagle:
I re-ran copybara with a fixed title similar to your suggestion, and the commit range in the body to make it easier for us to navigate between the two histories.

cgrindel:
> There is no "original PR" since this brings across a chunk of the commit history to fast-forward the OSS repo. It comes from many PRs in silo.
Yeah. I was thinking about it backwards, as if the sync was from `aspect-cli` to `silo`. Please disregard.

PR #307, opened 2022-10-11
chore: revert "feat: add `//install` target for installing the Aspect CLI on the system"

cgrindel:
Reverts aspect-build/aspect-cli#298


alexeagle:
Do we need this install tool to be in the silo repo?

cgrindel:
It is for CLI only. I thought the source of truth is silo. Also, this PR may be moot given our decision to just do a binary bottle formula for Homebrew.

alexeagle:
Yeah it seems like the install target is a false start, since we distribute a binary

alexeagle (approved):

PR #316, opened 2022-11-10
Install CLI for ourselves

alexeagle:


gregmagolan:
Already landed in another PR

PR #357, opened 2022-11-25
feat: introduce a plugins catalog

alexeagle:
This is just the MVP of the schema file and adds a single plugin. In the next PR I'll add more fields so that we can render something useful at docs.aspect.build/plugins

### DIFF
--- a/docs/plugins/plugins.json
+++ b/docs/plugins/plugins.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "./plugins.schema.json",
+  "plugins": [
+    {
+      "ghrepo": "aspect-build/plugin-fix-visibility"
+    }
+  ]
+}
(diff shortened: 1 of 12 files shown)


docs/plugins/plugins.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "./plugins.schema.json",
+  "plugins": [
+    {
+      "ghrepo": "aspect-build/plugin-fix-visibility"
```
gregmagolan, line 5: "ghrepo"?

alexeagle: That's what we used in the docs repo in the json schema for rulesets. Agree that doesn't look so good for a public API

kormide (approved):
:+1:

PR #360, opened 2022-11-26
docs: extend plugin catalog

alexeagle:


docs/plugins/plugins.schema.json
```diff
@@ -12,10 +12,40 @@
...
+          },
+          "homepage": {
+            "description": "A website describing the plugin. If not present, we'll use the GitHub UI",
+            "type": "string"
+          },
+          "name": {
+            "description": "The value users of the plugin put in the 'name' field of their .aspect/cli/plugins.yaml when using this plugin",
+            "type": "string"
+          },
+          "maintainers": {
```
alexeagle, line 24: note, I made this match the bazel-central-registry metadata.json schema just to reduce our overall cognitive load, may as well match the other place we publish stuff.

docs/plugins/plugins.d.ts
```diff
@@ -0,0 +1,41 @@
...
+ * and run json-schema-to-typescript to regenerate this file.
+ */
+
+export interface AspectCLIPluginsCatalog {
+  $schema?: string;
+  plugins: {
+    /**
+     * GitHub repository
+     */
+    "github-repo": string;
```
kormide, line 14: On the registry I think they use `repository`.

alexeagle: yeah, but they support more than just github, and we don't yet. Hm. I was going to just barrel ahead but I think you're right, it's better to have `repository = github:some/repo` and then do the error handling/reporting

docs/plugins/plugins.d.ts
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Source code repository, starting with the host, e.g. github:my-org/some-repo
+     */
+    repository: string;
+    /**
+     * A website describing the plugin. If not present, we'll use the GitHub UI
+     */
+    homepage?: string;
+    /**
+     * The value users of the plugin put in the 'name' field of their .aspect/cli/plugins.yaml when using this plugin
```
gregmagolan, line 20: "value users"?

docs/plugins/plugins.schema.json
gregmagolan: nice; looks good

gregmagolan (approved):

PR #361, opened 2022-11-27
chore: add release instructions

gregmagolan:


alexeagle (commented):
we also need to push the `v(N-4).x.x` tag right?

gregmagolan:
> we also need to push the `v(N-4).x.x` tag right?

Right. Will add that.

alexeagle (approved):
🙏🏻

PR #401, opened 2023-03-12
Use mapped kind for collision detection

### LINKED ISSUE #394: [Bug]: Gazelle plugin for JS does not handle map_kind correctly
### What happened?

Gazelle has many directives. One of them is map_kind which allows you to specify that you want to use a different build rule in place of a default 

`# gazelle:map_kind ts_project my_ts_project //:index.bzl`

The issue is that if you do this then the build file generator twice it will yell at you and say that there is already a target called `my_ts_project` with the name that `ts_project` wanted to use

### Version

aspect 5.2.1

### How to reproduce

```shell
Use `# gazelle:map_kind ts_project my_ts_project //:index.bzl` and run `aspect configure` twice. The first time should succeed and the second time should fail
```

### Any other information?

_No response_

### Fund our work

- [ ] Sponsor our open source work by donating a [bug bounty](https://opencollective.com/aspect-build/)

mrmeku:
Fixed #394


jbedard:
Was fixed in another commit, sorry to merge the other one :/

Issue #402, opened 2023-03-13
[FR]: Have js_validate_import_statements produce warnings instead of errors


mrmeku:
### What is the current behavior?

Currently, js_validate_import_statements is set to a boolean state. Either you get errors when import statements can't be validated or you get silence. 

### Describe the feature

For go resolutions, warnings are produced such that users can be alerted to problems without gazelle breaking entirely

I'd like to replicate that functionality for JS as well. js_validate_import_statements would be set to an enum set of states.
`# gazelle:js_validate_import_statements warning|error|off`

PR #411, opened 2023-03-25
Change validate_import_statements to a ternary

mrmeku:
Values are now
 - "error": same as true from before
 - "warn": produce warnings, but do not error
 - "off": same as false from before

Fixed: #402

### DIFF
--- a/docs/help/topics/directives.md
+++ b/docs/help/topics/directives.md
@@ -25,7 +25,7 @@ JavaScript directives for generating BUILD files follow the same format as [gaze
 | Imports matching the glob will be ignored when generating BUILD files in the specifying directory and descendants. |
 | `# gazelle:js_resolve _glob_ _target_`                  |                             |
 | Imports matching the glob will be resolved to the specified target within the specifying directory and descendants.<br />This directive is an extension of the standard `resolve` directive with added glob support and only applying to JavaScript rules. |
-| `# gazelle:js_validate_import_statements true\|false`   | `true`                      | 
+| `# gazelle:js_validate_import_statements error\|warn\|off`   | `error`                      | 
 | Ensure all import statements map to a known dependency. |
 | `# gazelle:js_project_naming_convention _name_`         | `{dirname}`                 |
 | The format used to generate the name of the main `ts_project` rule. |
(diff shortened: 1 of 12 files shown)


gazelle/js/resolve.go
```diff
@@ -166,21 +166,27 @@ func (ts *TypeScript) resolveModuleDeps(
...
 						"\t2. Ignore the dependency using the '# gazelle:%[3]s %[1]s' directive.\n"+
-						"\t3. Disable Gazelle resolution validation using '# gazelle:%[4]s false'",
+						"\t3. Disable Gazelle resolution validation using '# gazelle:%[4]s off'",
 					mod.ImportPath, mod.SourcePath, Directive_IgnoreImports, Directive_ValidateImportStatements,
 				)
-
+				resolutionErrors = append(resolutionErrors, notFound)
+			case ValidationWarn:
+				notFound := fmt.Errorf(
+					"Import %[1]q from %[2]q is an unknown dependency",
```
jbedard, line 187: Why not use the same message as the error case? The user might still need to fix it or want to ignore/resolve it to not see the message again...

docs/help/topics/directives.md
```diff
@@ -25,7 +25,7 @@ JavaScript directives for generating BUILD files follow the same format as [gaze
 | Imports matching the glob will be ignored when generating BUILD files in the specifying directory and descendants. |
 | `# gazelle:js_resolve _glob_ _target_`                  |                             |
 | Imports matching the glob will be resolved to the specified target within the specifying directory and descendants.<br />This directive is an extension of the standard `resolve` directive with added glob support and only applying to JavaScript rules. |
-| `# gazelle:js_validate_import_statements true\|false`   | `true`                      | 
+| `# gazelle:js_validate_import_statements "error"\|"warn"\|"off"`   | `error`                      |
```
jbedard, line 28: I don't think we want the quotes...?

gazelle/js/resolve.go
mrmeku: Fair point... it just seemed verbose for a warning, but consistency is good too

docs/help/topics/directives.md
mrmeku: Good catch

jbedard (approved):

jbedard:
Will be synced/merged in https://github.com/aspect-build/aspect-cli/pull/414

PR #429, opened 2023-04-25
chore(ci): enable aspect workflows

alexeagle:
Gives us a "staging" environment for workflows releases that has actual public traffic, and we don't use it as a "dev" environment.


gregmagolan (approved):
🐰

PR #457, opened 2023-06-06
chore: enable warming task on Workflows config

gregmagolan:


.aspect/workflows/config.yaml
```diff
@@ -2,6 +2,7 @@
 ---
 queue: aspect-cli
 tasks:
+  warming:
```
mattem, line 5: This is fine if you wanted to be explicit, but this is added for free, 
https://github.com/aspect-build/silo/blob/bfee9cfe3bfe82a212d9b9589545561f00529349/rosetta/src/configuration/helper.ts#L150-L159

mattem (approved):

Issue #499, opened 2023-08-04
[Bug]: tsconfig paths are not expanded correctly for relative paths


joelMuehlena:
### What happened?

Relative paths in nested (not root) tsconfig.json files are not expanded and resolved correctly, but are missing the actual relative path to the project. The below explanation will provide a little bit more context.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: aspect 5.5.2 / custom built from main

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

Language(s) and/or frameworks involved:

typescript

### Any other information?

I encountered some issues while trying to use aspect cli with `configure`. The problem occurs within a nested file structure with different tsconfig.json files e.g. at `<project-root>/frontend/app1/tsconfig.json` and relative paths configs.

Firstly the config file will be read and interpreted correctly like described in #396. But when I have the paths configured like so (`"@/*": ["./src/*"]`):

```json
{
  "compilerOptions": {
    "target": "es5",
    "lib": ["dom", "dom.iterable", "esnext"],
    "allowJs": true,
    "skipLibCheck": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "esModuleInterop": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsx": "preserve",
    "incremental": true,
    "plugins": [
      {
        "name": "next"
      }
    ],
    "paths": {
      "@/*": ["./src/*"]
    }
  },
  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
  "exclude": ["node_modules"]
}
```

Like shown in the previous snippet, the paths config is relative. When using the cli those imports using `@` cannot be resolved. Concretely the path will be expanded like the following: `@/app/test -> src/app/test` while normal dependencies should be formed like `frontend/app1/src/app/test`  to be resolved correctly.

I got it working with the following hack, which simply returns the actual relative path of the tsconfig file and expands all paths found to be relative to it with the passed path:

```diff
diff --git a/gazelle/js/typescript/config.go b/gazelle/js/typescript/config.go
index 04d1b24..3678a98 100644
--- a/gazelle/js/typescript/config.go
+++ b/gazelle/js/typescript/config.go
[...]

PR #500, opened 2023-08-05
fix: tsconfig paths are not expanded correctly for tsconfigs and libs not placed at workspace root

joelMuehlena:
### Type of change

- Bug fix (#499)

### Test plan

- New test cases added

### General

This fix ensures, that if an expanded import path like specified in the corresponding tsconfig.json (of a file which for example lays in the following space <root>/projects/p1) is not absolute e.g. `@org/liba/* -> ./*: @org/liba/test` will be expanded to `projects/p1/test`. Otherwise gazelle has issues resolving the correct imports.


xc2:
Hello @joelMuehlena, 

Thanks for your help with relative imports in Clover. They're working great with `baseUrl` and `paths`, but I'm running into an issue where they don't work with just `baseUrl` alone. Here's a codesandbox with the issue (basic ts setup without bazel): https://codesandbox.io/s/practical-architecture-wlft6z?file=/src/App.tsx:0-31

Could you please add support for `baseUrl`-alone as well? That would be a huge help!

Thanks again

joelMuehlena:
Hi @xc2,
I looked into your codesandbox & request and made some changes. Maybe you can have a look if those changes are reflecting your request.

joelMuehlena:
@jbedard this should be everything. Maybe you can have a look at it, when you have the time?

gazelle/js/typescript/tsconfig.go
```diff
@@ -99,6 +118,8 @@ func parseTsConfigJSON(cm *TsConfigMap, root, configDir string, tsconfigContent
 	var BaseUrl string
 	if c.CompilerOptions.BaseUrl != nil {
 		BaseUrl = path.Clean(*c.CompilerOptions.BaseUrl)
+	} else if baseConfig != nil && baseConfig.BaseUrl != "" {
```
jbedard, line 121: Does one of your tests fail without this?

gazelle/js/typescript/tsconfig_test.go
```diff
@@ -31,6 +47,45 @@ func assertExpand(t *testing.T, options *TsConfig, p string, expected ...string)
 	}
 }
 
+func TestBaseUrlRegex(t *testing.T) {
+	t.Parallel()
+
+	t.Run("valid strings", func(t *testing.T) {
+		t.Parallel()
```
jbedard, line 54: I don't think we use this in any of the other tests atm. Does it have to be in each test *and* above? I'd vote to just remove it for now, can think about it more in the future 🤷

gazelle/js/typescript/tsconfig.go
```diff
@@ -44,7 +61,9 @@ var DefaultConfigPaths = TsConfigPaths{
 	Map: &map[string][]string{},
 }
 
-var Log = logrus.New()
+// Matches strings starting with ../, ./, @ or .
+// Used to detect if a path could possibly expanded using baseUrl
+var baseurl_validation_regex = regexp.MustCompile(`^(\.\.?\/|@|\.)`)
```
jbedard, line 66: Do you have an example where the `@` case is required? I thought that would be considered a named import so it wouldn't be relative?

gazelle/js/typescript/tsconfig.go
```diff
@@ -140,21 +161,31 @@ func parseTsConfigJSON(cm *TsConfigMap, root, configDir string, tsconfigContent
...
 }
 
+// Returns the path from the bazel-root to the active tsconfig.json file
+// if the passed path is not absolute.
+// Or an empty string if the path is absolute
+func (c TsConfig) getRelativeExpansionIfLocal(importPath string) string {
+	// Absolute paths must never be expanded but everything else must be relative to the bazel-root
+	// and therefore expanded with the path to the current active tsconfig.json
+	if !path.IsAbs(importPath) {
+		BazelLog.Debugf("Found local path %s in tsconfig.json. Should be expanded with tsconfig dir: %s", importPath, c.ConfigDir)
```
jbedard, line 171: Maybe `Tracef`? I think this would be called very frequently?

gazelle/js/typescript/tsconfig.go
```diff
@@ -183,12 +214,19 @@ func (c TsConfig) ExpandPaths(from, p string) []string {
 			matchedText := p[len(m.prefix) : len(p)-len(m.suffix)]
 			mappedPath := strings.Replace(originalPath, "*", matchedText, 1)
 
-			mappedPath = path.Clean(mappedPath)
```
jbedard, line 186: I guess if tests pass this wasn't needed? 🤔 

All the other `possible` paths get cleaned though so maybe we should do it (but after the `Join` below).

gazelle/js/typescript/tsconfig.go
```diff
@@ -140,21 +161,31 @@ func parseTsConfigJSON(cm *TsConfigMap, root, configDir string, tsconfigContent
 	return &config, nil
 }
 
+// Returns the path from the bazel-root to the active tsconfig.json file
```
jbedard, line 164: What do you mean by "bazel-root"? Isn't it the directory of the tsconfig and not even related to the fact that we're working within a bazel repo?

gazelle/js/typescript/tsconfig.go
```diff
@@ -183,12 +214,19 @@ func (c TsConfig) ExpandPaths(from, p string) []string {
...
+	// Expand paths from baseUrl
+	// Must not to be absolute or relative to be expanded
+	// https://www.typescriptlang.org/tsconfig#baseUrl
+	if !baseurl_validation_regex.MatchString(p) && !path.IsAbs(p) {
+		possible = append(possible, path.Join(c.getRelativeExpansionIfLocal(p), c.BaseUrl, p))
+	}
+
+	BazelLog.Tracef("Found %d possible paths for %s: %v", len(possible), p, possible)
+
 	// Add 'rootDirs' as alternate directories for relative imports
```
jbedard, line 230: This same doesn't need to be done to `rootDir`s?

jbedard:
A bunch of minor comments but overall looks great! Could merge as-is but if you have time to address or reply to the comments that would be great 👍

joelMuehlena:
Yep no problem, I'll take a look on the review this week when I have some time.

xc2:
> Hi @xc2, I looked into your codesandbox & request and made some changes. Maybe you can have a look if those changes are reflecting your request.

Thank you, it works as expected!

gazelle/js/typescript/tsconfig_test.go
joelMuehlena: If the nesting cases with t.Run should also be parallel the t.Parallel must also be nested. Otherwise the wrapping test function will be parallel to other parallel tests and the nested ones are sequential inside the parallel wrapper func. But totally fair to remove it if not used otherwhere.

gazelle/js/typescript/tsconfig.go
joelMuehlena: Yes might be better. Just set this to debug for dev reasons (trace was more output and the most was not required for this fix). Will change to trace

gazelle/js/typescript/tsconfig.go
joelMuehlena: Join should do a clean anyways don't it?

gazelle/js/typescript/tsconfig.go
joelMuehlena: I slightly altered the comment to be more general. Otherwise one could just omit it

gazelle/js/typescript/tsconfig.go
joelMuehlena: Yep it should moved it down

gazelle/js/typescript/tsconfig.go
joelMuehlena: Actually you are right the @ should be omitted. My thought was the @ pattern is mostly used by path mappings which are handled before so we could save us a case here, but there also could be folders starting with @ which are not directly relative, where this regex is the condition guard.

If I remove the @ case a regex will be might be overkill wouldn't it? The cases are only `.,./,../` which also could be handled with an if and strings.HasPrefix. Any opinions on what would be the preferred way?

gazelle/js/typescript/tsconfig.go
joelMuehlena: No, but it handles the case, if a config was extended with a baseUrl and the local one not having one the extended one will be used. It's quite difficult to test this case without "a real file", because the parsing function loads the extension of a file directly. So we can remove this but then the case with a baseDir inherited from an extension file will not be covered.

gazelle/js/typescript/tsconfig.go
jbedard: I'm not sure. If it does we should remove all the `Clean`s in a followup commit.

gazelle/js/typescript/tsconfig.go
jbedard: I'd say put it in a util method and then do whatever you want, can always change the method later without the caller caring. Note that I think the `^\.` will also catch basic `.dotfile` cases? Maybe:
```
func isRelative(s string) {
   return s == "." or strings.HasPrefix(s, "./") or strings.HasPrefix(s, "../")
}
```
?

jbedard: Could also put the `path.IsAbs` within the util method, although I think that isn't actually needed?

gazelle/js/typescript/tsconfig.go
```diff
@@ -140,21 +159,33 @@ func parseTsConfigJSON(cm *TsConfigMap, root, configDir string, tsconfigContent
 	return &config, nil
 }
 
+// Returns the path from the project base to the active tsconfig.json file
+// This is used to build the path from the project base to the file being imported
+// because gazelle seems to resolve files relative to the project base
+// if the passed path is not absolute.
+// Or an empty string if the path is absolute
+func (c TsConfig) getRelativeExpansionIfLocal(importPath string) string {
```
jbedard, line 167: Can you think of another name without the "if local"? That seems like an implementation detail. `expandPath` or something like that?

gazelle/js/typescript/tsconfig.go
joelMuehlena: > Join joins any number of path elements into a single path, separating them with slashes. Empty elements are ignored. The result is Cleaned. ~ https://pkg.go.dev/path#Join with https://cs.opensource.google/go/go/+/master:src/path/path.go;l=179

So the Clean is performed by Join already.

gazelle/js/typescript/tsconfig.go
joelMuehlena: I put it with inside, because I think technically it is part of the check and so it is all tied together

gazelle/js/typescript/tsconfig.go
joelMuehlena: Renamed

gazelle/js/typescript/tsconfig.go
joelMuehlena: How should I handle this? Just leave it as is?

jbedard (approved):
Awesome, thanks @joelMuehlena

alexeagle:
Thanks, we brought the PR into our monorepo and exporting again in https://github.com/aspect-build/aspect-cli/pull/502

PR #539, opened 2023-10-18
chore: upgrade to Aspect Workflows 5.8.1

gregmagolan:
Also switch queue names to the default Workflows values of `aspect-default` and `aspect-default-warming`

---

### Type of change

- Chore (any other change that doesn't affect source or test files, such as configuration)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,6 @@ workflows:
     jobs:
       - legacy
       - aspect-workflows/setup:
-          resource_class: aspect-build/aspect-cli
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -35,6 +35,6 @@ workflows:
     jobs:
       - legacy
       - aspect-workflows/setup:
-          resource_class: aspect-build/aspect-cli
+          resource_class: aspect-build/aspect-default
```
kormide, line 38: Just to verify, you also created this resource class in CCI?

kormide (approved):

PR #594, opened 2024-01-04
chore: fix build without the bytes with Bazel 7

gregmagolan:
Tracked down the issue that caused delivery with build without the bytes to fail with Bazel 7. This will green up main.

---

### Type of change

- Chore (any other change that doesn't affect source or test files, such as configuration)

### Test plan

CI

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -13,6 +13,12 @@ common --workspace_status_command "${PWD}/workspace_status.sh"
 
 common:release -c opt --stamp
 
+# --noexperimental_check_output_files setting from .aspect/bazelrc/performance.bazelrc breaks "build without the bytes" with Bazel 7
+# NB: these can't be common since since performance.bazel uses {build,fetch,query} and these will stop over any common settings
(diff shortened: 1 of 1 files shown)


jbedard (approved):

.bazelrc
```diff
@@ -13,6 +13,12 @@ common --workspace_status_command "${PWD}/workspace_status.sh"
 
 common:release -c opt --stamp
 
+# --noexperimental_check_output_files setting from .aspect/bazelrc/performance.bazelrc breaks "build without the bytes" with Bazel 7
```
alexeagle, line 16: lots of users will need this change in their own repo as well, right?

gregmagolan: If they are using "build without the bytes" with bazel and our recommended settings and setting `remote_download_outputs=toplevel` then yes. I don't think that is a very common intersection however

alexeagle: ha, happened to come back here from the thread with @pyrocat101

PR #597, opened 2024-01-05
chore: switch to a pre-generated CircleCI configuration for Aspect Workflows

gregmagolan:
Switch to a simpler way to bring in the Aspect Workflows generated CircleCI configuration that doesn't require the `aspect-build/workflows@5.9.0` continuation Orb. The Orb does not work in many used cases when adopting Aspect Workflows since CircleCI restricts the circleci.yml to a single workflow when using the continuation Orb and many CircleCI users have many CircleCI workflows defined in their CircleCI config.

To adopt this pattern on a new Aspect Workflows CircleCI deployment:

1. Rename your current `.circleci/config.yml` file to make room for the merged configuration. For example, this repository uses `.circleci/user-config.yml`.
2. Add a `.aspect/workflows/BUILD.bazel` file and copy the contents from this commit
3. Add a `.aspect/workflows/deps.bzl` file and copy the contents from this commit. We'll reduce this boilerplate in the future by create a ruleset that contains it.
4. Add a `.circleci/BUILD.bazel` file and copy the contents from this commit. Set the `CIRCLECI_ORG` to your CircleCI organization name and set `CIRCLECI_USER_CONFIG_FILE` to the name of the your renamed configuration file from step 1.
5. Add the following two lines to your WORKSPACE file:

```
load("//.aspect/workflows:deps.bzl", "fetch_workflows_deps")
fetch_workflows_deps()
```
6. If you're using a tool such as prettier to format yaml, you may need to add `.circleci/config.yml` to your `.prettierignore` (or equivalent) file so that the generated yml file is not auto-formatted.


.circleci/BUILD.bazel
```diff
@@ -0,0 +1,50 @@
...
+    name = "rosetta",
+    actual = select({
+        "@bazel_tools//src/conditions:darwin_arm64": "@rosetta_darwin_aarch64//file:rosetta",
+        "@bazel_tools//src/conditions:darwin_x86_64": "@rosetta_darwin_x86_64//file:rosetta",
+        "@bazel_tools//src/conditions:linux_aarch64": "@rosetta_linux_aarch64//file:rosetta",
+        "@bazel_tools//src/conditions:linux_x86_64": "@rosetta_linux_x86_64//file:rosetta",
+    }),
+)
+
+alias(
```
alexeagle, line 17: Ah, I studied https://github.com/theoremlp/rules_multitool at one point and I think this may be a good opportunity to use it. Reduces all this boilerplate nicely without us needing a ruleset

.aspect/workflows/deps.bzl
```diff
@@ -0,0 +1,91 @@
...
+
+def http_archive(name, **kwargs):
+    maybe(_http_archive, name = name, **kwargs)
+
+def http_file(name, **kwargs):
+    maybe(_http_file, name = name, **kwargs)
+
+# buildifier: disable=function-docstring
+def fetch_workflows_deps():
+    http_file(
```
alexeagle, line 33: shouldn't this be in a `for platform, integrity in xx_integrity.keys()` ?

.circleci/BUILD.bazel
gregmagolan: There is still the sets of hashes in deps.bzl and the genrule in the BUILD file that really belong in a rule set so the DX is cleaner for Workflows deployments. I'd rather dep on a aspect_rules_workflows or aspect_rules_cicd and remove all of the complexity than dep on a rules_multitool and remove only a portion of it.

.aspect/workflows/deps.bzl
gregmagolan: yeah, I'll reduce this before landing

.circleci/BUILD.bazel
gregmagolan: I'm also thinking this pattern also makes sense to use for GitHub Actions to reduce the complexity we have right now with the setup step generating json that feeds into the re-usable workflow matrix. Stamping out the generated GHA workflows would be much much simpler. That also seems better than a generic GitHub Actions that runs the setup step using rosetta which generates json and feeds into the re-usable workflow.

.aspect/workflows/deps.bzl
gregmagolan: ✅

.circleci/BUILD.bazel
alexeagle: I'd still like to push on the "intended" way to use GHA. This is one of those cases where we spent a lot of time evaluating all the options for the way we are doing it now. We can escalate to someone about the inability to use a reusable workflow from another GH org.

alexeagle: Also I'm still not sold on us publishing another ruleset, that's signing us up for more overhead and maintenance, we have too many already for the amount of automation we have. I'd want to understand all the other things that will go in there, not just this one thing.
I know, by objecting I'm volunteering to propose something that works just as well...

.circleci/BUILD.bazel
```diff
@@ -0,0 +1,50 @@
...
+        "@bazel_tools//src/conditions:darwin_arm64": "@circleci_config_merge_darwin_aarch64//:circleci-config-merge",
+        "@bazel_tools//src/conditions:darwin_x86_64": "@circleci_config_merge_darwin_x86_64//:circleci-config-merge",
+        "@bazel_tools//src/conditions:linux_aarch64": "@circleci_config_merge_linux_aarch64//:circleci-config-merge",
+        "@bazel_tools//src/conditions:linux_x86_64": "@circleci_config_merge_linux_x86_64//:circleci-config-merge",
+    }),
+)
+
+genrule(
+    name = "aspect_workflows_config",
+    srcs = ["//.aspect/workflows:config.yaml"],
```
alexeagle, line 29: this file is missing, can you track that down? forgot to `git add`?

alexeagle (approved):
Okay I am caught up with the problem and possible solutions, agree this is the right start.

PR #638, opened 2024-02-27

gregmagolan:
Commit range https://github.com/aspect-build/silo/compare/1670a5925b744903859e773b6ee1360c24cb1b92..709281da44f53657fc84b7e939c774086e3c10af


bazel/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+load("@aspect_bazel_lib//lib:write_source_files.bzl", "write_source_files")
```
jbedard, line 1: @gregmagolan I think these all need to be dropped, the copybara config isn't dropping them atm

alexeagle: I can take another pass on getting this "copybara-clean".

We had a colliding vision on how to do `.pb.go` file generation between the two repos :(

alexeagle: I'll take over this PR...

alexeagle (approved):

PR #647, opened 2024-03-07

gregmagolan:
Bazel Open Source Support customer waiting on the 
```
fix: handle BEP lint results concurrently
```
change in this sync.

Commit range https://github.com/aspect-build/silo/compare/b2e57a7be83f5e3a055c1f37287c647c0b363c59..650e805ea86d684dc89bbbdcd74b359b19b7dcf2


cmd/aspect/root/BUILD.bazel
```diff
@@ -35,12 +35,12 @@ go_library(
         "//cmd/aspect/sync",
         "//cmd/aspect/test",
         "//cmd/aspect/version",
-        "//docs/help/topics",
```
alexeagle, line 38: sorry these changes are still spurious since copybara isn't configured to match 100%. Have to be manually reverted after the sync

jbedard:
The "tidy" commit is the extra commit (not from copybara) to get things green? I guess that's all things like lint that fail (because /cli is dropped form the path) and generated files?

jbedard (approved):
I think the commits look good, as long as you've taken a pass as well to make sure commits aren't pulling in extras that should be in a separate cleanup commit.

PR #658, opened 2024-03-26
Cache bazel flags proto

dzbarsky:
It takes around 65ms to execute the `bazel help` commands in the new workspace to grab the flags. We can cache the flags per-version to avoid that.

### Type of change
- Performance (a code change that improves performance)

```
➜  aspect-cli git:(zbarsky/version) hyperfine -i 'bazel build //:no-such-target'                                                                                                                                                                                                 ~/aspect-cli
Benchmark 1: bazel build //:no-such-target
  Time (mean ± σ):     285.6 ms ±   4.2 ms    [User: 56.3 ms, System: 31.2 ms]
  Range (min … max):   280.3 ms … 292.4 ms    10 runs

➜  aspect-cli git:(zbarsky/version) ✗ hyperfine -i 'bazel-bin/cmd/aspect/aspect_/aspect build //:no-such-target'                                                                                                                                                                 ~/aspect-cli
Benchmark 1: bazel-bin/cmd/aspect/aspect_/aspect build //:no-such-target
  Time (mean ± σ):     244.4 ms ±   4.4 ms    [User: 49.1 ms, System: 17.9 ms]
  Range (min … max):   238.2 ms … 251.1 ms    12 runs
```

### Test plan
The first commit adds timings so you can see how the cache helps. The second commit removes them to keep the code clean. (We should squash before merging)

### DIFF
--- a/cmd/aspect/main.go
+++ b/cmd/aspect/main.go
@@ -18,9 +18,7 @@ package main
 
 import (
 	"context"
-	"fmt"
 	"os"
-	"time"
 
 	"aspect.build/cli/cmd/aspect/root"
 	"aspect.build/cli/pkg/aspect/root/config"
@@ -32,7 +30,6 @@ import (
 )
 
 func main() {
-	start := time.Now()
 	// Convenience for local development: under `bazel run <aspect binary target>` respect the
 	// users working directory, don't run in the execroot
 	if wd, exists := os.LookupEnv("BUILD_WORKING_DIRECTORY"); exists {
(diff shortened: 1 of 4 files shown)


cmd/aspect/main.go
```diff
@@ -50,15 +50,15 @@ func main() {
 	root.HandleVersionFlags(streams, os.Args[1:], bzl)
 
 	// Re-enter another aspect if version running is not the configured version
-	reentered, err := bzl.HandleReenteringAspect(streams, os.Args[1:], root.CheckAspectLockVersionFlag(os.Args[1:]))
+	bazelVersion, reentered, err := bzl.HandleReenteringAspect(streams, os.Args[1:], root.CheckAspectLockVersionFlag(os.Args[1:]))
```
dzbarsky, line 53: might make sense to expose this on the bazel interface instead. Up to yall, I don't feel strongly how this is spelled. This reentrancy bit is very complex FWIW :)

gregmagolan:
Replaced by https://github.com/aspect-build/aspect-cli/commit/540b9ff8d2731f6246cdcb0b1ba1cadb2de45db9

Issue #602, opened 2024-01-13
[Bug]: bazel --query_file goes to interactive


justhecuke:
### What happened?

```
$ bazel --version
aspect 5.8.20

$ echo '//:all' > queryfile.txt
$ bazel query --query_file queryfile.txt
```
Results in an interactive prompt:
```
Use the arrow keys to navigate: ↓ ↑ → ← 
? Select a preset query: 
  ▸ why: Determine why targetA depends on targetB
    deps: Get the deps of a target
```

This should not happen. The query should execute what is in the query_file and not start an interactive prompt.

### Version

Development (host) and target OS/architectures:
22.04.1-Ubuntu SMP Fri Nov 17 21:07:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

Output of `bazel --version`:
aspect 5.8.20

Our `.bazelversion` is `6.4.0`

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
N/A

Language(s) and/or frameworks involved:
N/A

### How to reproduce

```shell
echo '//:all' > queryfile.txt
bazel query --query_file queryfile.txt
```

gregmagolan:
Calling
```
bazel query --aspect:interactive=false --query_file queryfile.txt
```
should work-around the issue.

We should fix this in the CLI, however, to not go into interactive mode if the user specifies a `--query_file`

hjellek:
> Calling
> 
> ```
> bazel query --aspect:interactive=false --query_file queryfile.txt
> ```
> 
> should work-around the issue.
> 
> We should fix this in the CLI, however, to not go into interactive mode if the user specifies a `--query_file`

This does not seem to work correctly for me, I get interactive even with `--aspect:interactive=false`:

```
bazel query --aspect:interactive=false --query_file /tmp/queryfile.txt
✗ Include predefined aquery's and cquery's when calling query:
Created $HOME/.aspect/cli/config.yaml
```

After `$HOME/.aspect/cli/config.yaml` is created, it does not happen.
It's a problem in CI for us whenever a new runner executes for the first time.

PR #659, opened 2024-03-27

alexeagle:
Commit range https://github.com/aspect-build/silo/compare/407c86388bd42af767b828c758c56ed159b33870..6ce527b9f4427b7c55fdeea134c456441cb0b6c2


mattem:
I don't think this commit will fix #602 , this is explicitly enabling the use of `--query_file` on the `outputs` verb.

alexeagle:
I see, removed the "fixes" tag in that case.

mattem (approved):

PR #690, opened 2024-05-13
build: update and patch com_grail_bazel_toolchain

jbedard:
### Test plan

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -32,10 +32,14 @@ http_archive(
 http_archive(
     name = "com_grail_bazel_toolchain",
     patch_args = ["-p1"],
-    patches = ["//patches:com_grail_bazel_toolchain.patch"],
-    sha256 = "7d847d4f807797269a6389f1072233b58e66c7786477ce7b8c28b6f7293251d5",
-    strip_prefix = "toolchains_llvm-ceeedcc4464322e05fe5b8df3749cc02273ee083",
-    urls = ["https://github.com/bazel-contrib/toolchains_llvm/archive/ceeedcc4464322e05fe5b8df3749cc02273ee083.tar.gz"],
+    # Note: these commits are on the silo branch of aspect-forks/bazel-toolchain
+    patches = [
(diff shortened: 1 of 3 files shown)


WORKSPACE
```diff
@@ -32,10 +32,14 @@ http_archive(
...
-    strip_prefix = "toolchains_llvm-ceeedcc4464322e05fe5b8df3749cc02273ee083",
-    urls = ["https://github.com/bazel-contrib/toolchains_llvm/archive/ceeedcc4464322e05fe5b8df3749cc02273ee083.tar.gz"],
+    # Note: these commits are on the silo branch of aspect-forks/bazel-toolchain
+    patches = [
+        "//patches:com_grail_bazel_toolchain.patch",
+        "//patches:com_grail_bazel_toolchain.001.patch",
+    ],
+    sha256 = "a9fc7cf01d0ea0a935bd9e3674dd3103766db77dfc6aafcb447a7ddd6ca24a78",
+    strip_prefix = "toolchains_llvm-c65ef7a45907016a754e5bf5bfabac76eb702fd3",
+    urls = ["https://github.com/bazel-contrib/toolchains_llvm/archive/c65ef7a45907016a754e5bf5bfabac76eb702fd3.tar.gz"],
```
alexeagle, line 42: nit: we should remove "grail" from the name, since it was donated to bazel-contrib that's purely legacy

jbedard: Should I do it in this aspect-cli PR though? Where I'm just copying this from silo...

alexeagle: I'm not sure, Maybe it should be made in both CLI and silo, I agree that lining them up is most likely to unwedge the CLI release

jbedard: It seems like rules_go still depends on the old name, so we either keep it as-is or add an alias to keep rules_go happy. I'd prefer just merging this in the cli repo and trying the alias in silo another time...

alexeagle (approved):

PR #710, opened 2024-06-07
process: sync commits from Aspect-internal silo

gregmagolan:
Commit range https://github.com/aspect-build/silo/compare/580f85e4ccbdb3124c75d39c220c0e686667acb2..e27563aedb24f14a85ed2072f23c0f995877997d


alexeagle:
I think I'll clobber this with #712

PR #730, opened 2024-07-29
process: sync commits from Aspect-internal silo

jbedard:
Commit range https://github.com/aspect-build/silo/compare/a03515edba43f9546cb8ee8990838114dcc7a484..4525e80cbee246850bdb2325ca202fea0d60d0a9


.github/workflows/tidy_up_repository.sh
```diff
@@ -4,26 +4,26 @@ set -o errexit -o nounset -o pipefail
 
 # First verify if the repository is clean.
 if ! git diff --exit-code; then
-	error_msg="ERROR: The repository is not clean - please verify the changes and fix them."
```
jbedard, line 7: I don't know why stuff like this changed. I worry it's config files or versions out of sync? :/

alexeagle (approved):

.github/workflows/tidy_up_repository.sh
alexeagle: Maybe we should just stop formatting files in this repo

PR #743, opened 2024-09-10
chore: remove homebrew from release artifacts and switch to monorepo versioning

gregmagolan:
Aspect CLI OSS open-core will no longer be available at the `aspect-build/aspect/aspect` homebrew formula. Instead the Aspect Community Licensed Aspect CLI which includes closed source code on top of the OSS open-core will be shipped via that homebrew formula.

In the future, a from-source Homebrew formula could be created (similar to what `bazelisk` does) to build the Aspect CLI OSS from source and have it available at `brew install aspect`.

This change also switch the versioning and stamping for Aspect CLI OSS releases to use our monorepo versioning scheme `<year>.<week>.<commits_in_week>-<commit_hash>`. A GitHub Action on a weekly cron to add the weekly tag is added here.


README.md
```diff
@@ -19,15 +19,19 @@ See more on our product webpage: <https://aspect.build/cli>
 
 # Installation
 
-## Homebrew (MacOS)
+## Bazelisk (MacOS / Linux)
```
alexeagle, line 22: I think this repo should have both sections, or else some users will navigate here, install OSS, and wonder why starzelle isn't working.

Standard install
brew ...

OSS Code only
releases page ...

alexeagle (approved):

README.md
gregmagolan: Done

PR #753, opened 2024-10-05
gazelle/kotln: Fix import resolution to use ancestor packages

gonzojive:
When a Kotlin file imports "xyz.abc.foo.Bar.Baz", gazelle will only resolve the import successfully if "xyz.abc.foo.Bar" can be resolved to a Bazel label.

The plugin should also check if "xyz.abc.foo", "xyz.abc", or "xyz" resolve to a bazel label.

---

- Breaking change (forces users to change their own code or config): no? (bug fix to match user expectations)
- Suggested release notes appear below: yes/no

### Test plan

- New test cases added: `gazelle/kotlin/tests/deep_import/` gazelle test.


alexeagle:
This is merged in our internal monorepo, will sync back to OSS shortly!

Issue #766, opened 2024-10-29
[Bug]: Looks like arm64 was dropped?


tradergt:
### What happened?

Release for arm64 seems to be dropped: 
```> bazel test //...
2024/10/29 13:37:16 Downloading https://github.com/aspect-build/aspect-cli/releases/download/2024.41.29/bazel-2024.41.29-linux-arm64...
2024/10/29 13:37:16 could not download Bazel: failed to download bazel: failed to download bazel: HTTP GET https://github.com/aspect-build/aspect-cli/releases/download/2024.41.29/bazel-2024.41.29-linux-arm64 failed with error 404

```

### Version

Development (host) and target OS/architectures:
Linux workstation.smelser.org 6.6.47+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.6.47-1+rpt1 (2024-09-02) aarch64 GNU/Linux

Output of `bazel --version`:
> bazel --version
aspect 5.10.15

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

bazel_dep(name = "aspect_rules_py", version = "1.0.0-rc0")
bazel_dep(name = "rules_python", version = "0.37.2", dev_dependency = True)

Language(s) and/or frameworks involved:
python/go

### How to reproduce

```shell
If your on an arm64 instance, if you use the latest release, it will fail
```

jbedard:
Yeah it looks like the binary version is different then the release version. If you use the version in the file URLs does it work for you?

I'll have to look into why the wrong filenames are being used though 🤔

tradergt:
Its all good, I will just stay on the version I am on and get the next version or the fixed version. Mainly just wanted to report the bug.

jbedard:
I probably would not have noticed if you didn't, thanks! 👍

jbedard:
I think this was just me doing a release without fetching tags first, see https://github.com/aspect-build/aspect-cli/pull/767

I also kicked off a new release that should have the right numbers.

tradergt:
Nice, I just tried it and it all works. Thanks!

PR #767, opened 2024-10-30
docs: add note about fetching tags during release

jbedard:
So I don't forget next time...

---

### Test plan

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,12 +44,14 @@ After this, you should be able to merge your changes without any conflicts in th
 
 ## Releasing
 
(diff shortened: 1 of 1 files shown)


jbedard:
FWIW the issue prompting this was me tagging + pushing a release without first doing `git fetch --tags` first. So the generated locally was not up to date, but the tag generated by CI for the binaries (but not the release notes) was correct (see https://github.com/aspect-build/aspect-cli/issues/766).

Why does CI generate the tag again? I'm not sure if that's worth investigating or fixing now, but fyi.

alexeagle (approved):

PR #785, opened 2024-12-03
Handle non `@types` imports from tsconfig `compilerOptions.types`

walkerburgin:
The TypeScript docs for `compilerOptions.types` make it sound like `compilerOptions.types` only applies to `@types` packages, but in practice it can also be used to include types from non `@types` in the global scope. For example, `cypress` recommends this in their [docs](https://docs.cypress.io/app/tooling/typescript-support#Configure-tsconfigjson): 

```json
{
  "compilerOptions": {
    "target": "es5",
    "lib": ["es5", "dom"],
    "types": ["cypress", "node"]
  },
  "include": ["**/*.ts"]
}
```

From what I can tell, if you don't specify `compilerOptions.types`, TypeScript will **not** automatically include types from a visible non `@types` package like `cypress` (`describe`,  `beforeEach`, etc). But you can include non `@types` packages in `compilerOptions.types`  to allow list them into the global scope. 

---

- Suggested release notes appear below: yes/no

### Test plan
I updated the existing test case for `compilerOptions.types`. 

- New test cases added

### DIFF
--- a/gazelle/js/tests/tsconfig_deps/types/tsconfig.json
+++ b/gazelle/js/tests/tsconfig_deps/types/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "types": ["jquery", "@testing-library/jest-dom", "ignored", "@test/ignored"]
+    "types": ["jquery", "@testing-library/jest-dom", "cypress"]
   }
 }
--- a/gazelle/js/tests/tsconfig_deps/types/tsconfig.json
+++ b/gazelle/js/tests/tsconfig_deps/types/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "types": ["jquery", "@testing-library/jest-dom", "cypress"]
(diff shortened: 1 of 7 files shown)


gazelle/js/tests/tsconfig_deps/types/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "types": ["jquery", "@testing-library/jest-dom", "ignored", "@test/ignored"]
```
jbedard, line 4: Why drop the ignored ones?

walkerburgin: I think it was originally just a mistake on my part.

Looking at it again, I'm not sure if these should be silently ignored or if they should fail. They're essentially unknown dependencies, right?

jbedard: Yeah, basically unknown dependencies.

I'd prefer keeping them to ensure we aren't changing existing behaviour. We can debate if it should error or silently ignore another time.

jbedard:
This has been merged into the internal repo and will be synced back to this repo soon.

walkerburgin:
:+1: thanks!

PR #786, opened 2024-12-05
Handle `/// <reference lib="..." />` comments

walkerburgin:
According to the TypeScript docs ([here](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-lib-)): 

> This directive allows a file to explicitly include an existing built-in lib file.
>
> Built-in lib files are referenced in the same fashion as the [lib](https://www.typescriptlang.org/tsconfig#lib) compiler option in tsconfig.json (e.g. use lib="es2015" and not lib="lib.es2015.d.ts", etc.).

I think `bazel configure` is currently attempting to interpret them as types imports, leading to errors like this: 

```
Import "webworker" from "pkgs/foo-lib/src/worker.ts" is an unknown dependency. Possible solutions:
	1. Instruct Gazelle to resolve to a known dependency using a directive:
		# gazelle:resolve [src-lang] js import-string label
		   or
		# gazelle:js_resolve import-string-glob label
	2. Ignore the dependency using the '# gazelle:js_ignore_imports webworker' directive.
	3. Disable Gazelle resolution validation using '# gazelle:js_validate_import_statements off'
```

`pkgs/foo-lib/src/worker.ts`:

```ts
/// <reference lib="webworker" />

postMessage({ message: "Hello, world" });
```

---

- Breaking change (forces users to change their own code or config): yes (I guess if somebody was accidentally using `lib` instead of `types`?)

    `aspect configure` handles `/// <reference lib="..." />` comments correctly

### Test plan

- Covered by existing test cases (I updated the existing test)

### DIFF
--- a/gazelle/js/parser/parser.go
+++ b/gazelle/js/parser/parser.go
@@ -53,7 +53,7 @@ var importQueries = map[string]string{
 	`,
 }
 
-var tripleSlashRe = regexp.MustCompile(`^///\s*<reference\s+(?:lib|path|types)\s*=\s*"(?P<lib>[^"]+)"`)
+var tripleSlashRe = regexp.MustCompile(`^///\s*<reference\s+(?P<attr>lib|path|types)\s*=\s*"(?P<lib>[^"]+)"`)
 
 func ParseSource(filePath string, sourceCode []byte) (ParseResult, []error) {
 	imports := make([]string, 0, 5)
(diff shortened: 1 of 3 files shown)


jbedard (approved):

gazelle/js/parser/parser.go
```diff
@@ -53,7 +53,7 @@ var importQueries = map[string]string{
 	`,
 }
 
-var tripleSlashRe = regexp.MustCompile(`^///\s*<reference\s+(?:lib|path|types)\s*=\s*"(?P<lib>[^"]+)"`)
+var tripleSlashRe = regexp.MustCompile(`^///\s*<reference\s+(?P<attr>lib|path|types)\s*=\s*"(?P<lib>[^"]+)"`)
```
jbedard, line 56: Can we just remove `lib|` from here?

walkerburgin: I think so. I wasn't sure whether you'd prefer to parse it and then explicitly ignore it or just pretend like we never saw it. Defer to you!

walkerburgin: Updated!

Issue #799, opened 2025-01-22
[Bug]: Runfiles manifest entries with escapes cause issues with bazel outputs


ewhauser:
### What happened?

I'm seeing the following error from `aspect outputs`:

```
(19:31:01) [ASPECT] [XTRACE] [RUN] (/mnt/ephemeral/workdir/monorepo/monorepo) /etc/aspect/workflows/bin/h2o --root-start-time 1737401430 --trace-id e1ee01275a2ee972a7656e39948c46d3 --span-id adda15fd87db3b50 --profile /mnt/ephemeral/output/monorepo/__main__/delivery_manifest.profile.gz
(19:31:01) [ASPECT] [XTRACE] [EXEC] (.) bazel --nohome_rc --bazelrc=.aspect/workflows/bazelrc --bazelrc=.bazelrc --output_user_root=/mnt/ephemeral/bazel/monorepo/__main__ --output_base=/mnt/ephemeral/output/monorepo/__main__ outputs --aspect:lock_version --profile=/mnt/ephemeral/output/monorepo/__main__/delivery_manifest.profile.gz --query_file=/tmp/rosetta-qdmSgR/delivery_manifest.outputs.txt --announce_rc ExecutableHash
(19:31:02) [ASPECT] [XTRACE] [RUN] (/mnt/ephemeral/workdir/monorepo/monorepo) /etc/aspect/workflows/bin/h2o --root-start-time 1737401461 --trace-id e1ee01275a2ee972a7656e39948c46d3 --span-id 410c11101f1fb088 --profile /mnt/ephemeral/output/monorepo/__main__/delivery_manifest.profile.gz
╭─
 Error: 
  x Failed to query for deliverable targets
  ├▶ • Error running bazel command: 'bazel outputs (invocation 1c41abcf-f3e8-4072-ae76-c496b24e0ef1)
  ╰▶ • A build or parsing failure occurred.
  
  ╭[exit code: 1]
  │
  ├[stdout]
  · 
  ·
  ├[stderr]
  · Error: failed to stat runfiles manifest entry rules_python~~python~python_3_11_4_x86_64-unknown-linux-gnu/lib/python3.11/site-packages/setuptools/command/launcher\smanifest.xml: stat rules_python~~python~python_3_11_4_x86_64-unknown-linux-gnu/lib/python3.11/site-packages/setuptools/command/launcher\smanifest.xml: no such file or directory


  ╰
╰─
```

[...]

PR #800, opened 2025-01-22
fix(cli): unescape runfiles entries from manifest #799

ewhauser:
Fix for #799

### DIFF
--- a/pkg/bazel/runfiles/runfiles.go
+++ b/pkg/bazel/runfiles/runfiles.go
@@ -1,0 +1,33 @@
+package runfiles
+
+// Unescape the behavior of the C++ implementation
+// https://github.com/bazelbuild/bazel/blob/release-8.0.1/src/main/tools/build-runfiles.cc#L107
+func Unescape(path string) string {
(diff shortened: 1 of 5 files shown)


pkg/bazel/runfiles/runfiles.go
```diff
@@ -0,0 +1,33 @@
+package runfiles
+
+// Unescape the behavior of the C++ implementation
+// https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/main/tools/build-runfiles/build-runfiles.cc#L76
```
jbedard, line 4: This link is a 404...?

jbedard:
Was there any progress on this? What was the reason for closing?

ewhauser:
Left a comment on the ticket. This was a bug in `rules_multirun`; there is no reason that an escaped key should actually be in the value entry which is what is being hashed.

jbedard:
Ok, thanks for the info 👍

PR #810, opened 2025-02-06
Do not log a warning on `null` subpath exports

walkerburgin:
Small quality of life improvement to stop logging a warning on `null` subpath exports in `package.json`: 

```json
{
  "exports": {
    "./features/*.js": "./src/features/*.js",
    "./features/private-internal/*": null
  }
}
```

---

### Test plan

### DIFF
--- a/gazelle/js/tests/npm_package_deps/exports-ts/package.json
+++ b/gazelle/js/tests/npm_package_deps/exports-ts/package.json
@@ -3,6 +3,7 @@
   "private": true,
   "exports": {
     ".": "./src/main.js",
-    "./stars/*": "./other/*.js"
+    "./stars/*": "./other/*.js",
+    "./stars/internal/*": null
   }
 }
--- a/gazelle/js/node/package.go
+++ b/gazelle/js/node/package.go
@@ -59,6 +59,16 @@ func ParsePackageJsonImportsFile(rootDir, packageJsonPath string) ([]string, err
 				case string:
 					// Regular subpath export
 					imports = append(imports, path.Clean(e))
+				case nil:
+					// According to https://nodejs.org/api/packages.html#subpath-patterns, to exclude
+					// private subfolders from patterns, null targets can be used:
+					// {
+					//   "exports": {
+					// 	   "./features/*.js": "./src/features/*.js",
+					// 	   "./features/private-internal/*": null
+					// 	 }
+					// }
+					break
 				case map[string]interface{}:
 					// Conditional subpath export
 					for subEKey, subE := range e {


jbedard:
Was there stdout spam previously? Can we edit a package.json `exports` in one of the tests to test that? The gazelle tests record stdout and assert it doesn't change so I think this can be reprod in a test npm...

walkerburgin:
Yeah, here's what prompted this: 

```
➜   bazel configure
Updating BUILD files for go, javascript, aspect-configure
2025/02/05 16:53:45 unknown exports../internal/* type: <nil>
653 BUILD files visited
0 BUILD files updated
```

With a `package.json` that has this in its `exports`: 

```json5
"exports": {
    // ...
    "./*": {
        "types": "./dist/*.d.ts",
        "default": "./dist/*.js"
    },
    "./internal/*": null
}
```

I'm still trying to figure out how to repro this behavior in a test

jbedard:
If you search for `"exports":` in a test and add the `null` entry does it cause that test to fail? The stdout is part of the test assertions, I think by default with no opt-in. See some `expectedStdout.txt` files in there.

gazelle/js/tests/npm_package_deps/exports-ts/package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "exports": {
     ".": "./src/main.js",
-    "./stars/*": "./other/*.js"
+    "./stars/*": "./other/*.js",
+    "./stars/internal/*": null
```
walkerburgin, line 7: If I do this on `main` and then change [this](https://github.com/aspect-build/aspect-cli/blob/1e6308b7b1799c1b8716630efad35629aabe9d36/gazelle/js/node/package.go#L73) from `BazelLog.Warnf("unknown exports.%s type: %T", exportKey, export)` to `BazelLog.Fatalf("unknown exports.%s type: %T", exportKey, export)` I can get `bazel test --test_output=all //gazelle/js:npm_package_deps_test` to fail. 

But I think [this](https://github.com/aspect-build/aspect-cli/blob/1e6308b7b1799c1b8716630efad35629aabe9d36/pkg/logger/logger.go#L35-L55) is interfering with `expectedStdout.txt` / `expectedStderr.txt` 🤔

jbedard: Oh, was it only logging and not spamming stdout? Then with tests the logging wasn't even visible...?

If so then just adding this line to this test is enough 👍

walkerburgin: What I'm seeing in our codebase is that `bazel configure` logs a message like this for each `null` in `"exports"`:

```
2025/02/06 07:51:30 unknown exports../internal/* type: <nil>
```

I'm not sure where the line is between logging and spamming -- I was mostly just trying to pre-empt questions from devs in our repo since having a `null` in `"exports"` in this case is intentional.

I didn't dig in to how the `epectedStdout.txt`/`expectedStderr.txt` is implemented, but I confirmed that on `main` adding `"./stars/internal/*": null` here causes us to hit the `BazelLog.Warnf("unknown exports.%s type: %T", exportKey, export)` statement. During tests it looks like the logger writes to a file so it never makes it to `stdout`/`stderr`. If I comment out [this](https://github.com/aspect-build/aspect-cli/blob/1e6308b7b1799c1b8716630efad35629aabe9d36/pkg/logger/logger.go#L35-L55) I can get a failing test: 

image

PR #864, opened 2025-08-22
fix: allow config file to silence successful linters

alexeagle:
Makes it possible to have `--quiet` behavior in a repo by default, rather than setting via individual command-line flag

### DIFF
--- a/pkg/aspect/lint/lint.go
+++ b/pkg/aspect/lint/lint.go
@@ -186,6 +186,9 @@ lint:
 	requestReports, _ := cmd.Flags().GetBool("report")
 	machineReports, _ := cmd.Flags().GetBool("machine")
 	hideSuccess, _ := cmd.Flags().GetBool("quiet")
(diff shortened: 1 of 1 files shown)


pkg/aspect/lint/lint.go
```diff
@@ -186,6 +186,9 @@ lint:
 	requestReports, _ := cmd.Flags().GetBool("report")
 	machineReports, _ := cmd.Flags().GetBool("machine")
 	hideSuccess, _ := cmd.Flags().GetBool("quiet")
+	if viper.GetBool("lint.quiet") {
```
jbedard, line 189: Isn't this backwards? The CLI flag should override the config file...?

alexeagle: In both cases I can't tell how to get a tri-value "set to true, set to false, not set" which is why I did it this way

jbedard: Don't we need to support `--quiet=false` though?

PR #876, opened 2025-10-08
fix: fail run--watch startup when watch protocol explicitly supported but fails to connect

jbedard:
Close https://github.com/aspect-build/aspect-cli-legacy/issues/4

### Test plan

- Manual testing:
  1. run `aspect run --watch` on a random target and ensure the "fallback to restart" logic occurs
  2. add the `supports_incremental_build_protocol` tag and ensure the run stops with a fatal error

### DIFF
--- a/pkg/aspect/run/aspect_watch.bzl
+++ b/pkg/aspect/run/aspect_watch.bzl
@@ -31,17 +31,17 @@ cp {info_file} {aspect_watch_watch_manifest}
         arguments = [
             default.files_to_run.executable.path,
             str(ctx.label),
-            " ".join(ctx.rule.attr.tags),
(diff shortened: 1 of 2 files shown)


pkg/aspect/run/aspect_watch.bzl
```diff
@@ -31,17 +31,17 @@ cp {info_file} {aspect_watch_watch_manifest}
         arguments = [
             default.files_to_run.executable.path,
             str(ctx.label),
-            " ".join(ctx.rule.attr.tags),
+            ",".join(ctx.rule.attr.tags),
```
jbedard, line 34: I'm not sure when this bug was introduced, I think at some point we moved the fetching of this tag data into this aspect and it must have happened when we moved it?

PR #885, opened 2025-10-15
refactor(llvm): get a smaller distribution

alexeagle:
This is a @dzbarsky special :)

### DIFF
--- a/bazel/include/llvm.MODULE.bazel
+++ b/bazel/include/llvm.MODULE.bazel
@@ -26,24 +26,32 @@ execution_cpus = [
 # From https://github.com/dzbarsky/static-clang/releases/tag/v20.1.1-4
 dzbarsky_static_clang = {
     "macos": {
-        "aarch64": struct(
-            sha256 = "a4e65db49d01e4b37abcbc36a03e4f1121fbea63a7c7b58f9d49e3848a4cbdfc",  # darwin_arm64.tar.zst
+        "aarch64": {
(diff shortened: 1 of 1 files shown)


bazel/include/llvm.MODULE.bazel
```diff
@@ -21,13 +21,42 @@ execution_cpus = [
...
 [
     [
         llvm.toolchain(
             name = "llvm_toolchain_{}_{}".format(exec_os, exec_cpu),
             exec_arch = exec_cpu,
             exec_os = exec_os if exec_os != "macos" else "darwin",
-            llvm_version = "20.1.2"
+            sha256 = dzbarsky_static_clang[exec_os][exec_arch].sha256,
+            strip_prefix = dzbarsky_static_clang[exec_os][exec_arch].strip_prefix,
+            urls = dzbarsky_static_clang[exec_os][exec_arch].urls,
```
dzbarsky: Maybe I'm misreading from phone but I don't see where urls or strip_prefix are defined?

bazel/include/llvm.MODULE.bazel
```diff
@@ -21,13 +21,42 @@ execution_cpus = [
...
+            llvm_version = "20.1.1"
```
dzbarsky, line 67: You can drop the version now, I think it will be ignored

dzbarsky:
I'll send a PR soon. We've been cooking upstream

PR #895, opened 2025-10-31
feat: config, tools

arrdem:
- Start processing `config.axl`
- Start setting up for registering tools (bazelisk, ...) via `config.axl`.

### Test plan

N/A.


arrdem:
@gregmagolan should we just close this?

PR #897, opened 2025-11-06
feat: make streams more reliable

thesayyn:
BREAKING CHANGES

- `ctx.bazel.build` arg rename: `bazel_verb` to `command`, enforced that it can only have `test` or `build` as its value
- `ctx.bazel.build` arg rename: `bazel_flags` to `flags`
- `ctx.bazel.build` now supports `workspace_events` argument and `workspace_events()` stream. 
- `build_events|execution_logs`: the `poll` function has been renamed to `try_pop`
- `build_events`, `execution_logs`, `workspace_events` now have a `done()` function which returns `true` if the stream is done, and all the events are received either via `for` loop or calling of `try_pop`.
- move `UNIX Pipe` logic into `galvanize` crate for code share, it now checks if Bazel closes the writing end of pipe for reliably determining pipe status.

### DIFF
--- a/crates/galvanize/src/lib.rs
+++ b/crates/galvanize/src/lib.rs
@@ -4,7 +4,6 @@ use std::path::PathBuf;
 
 use nix::sys::stat::Mode;
 use nix::unistd::mkfifo;
-use proc_pidinfo::*;
 
 pub struct Pipe {
     path: PathBuf,
@@ -36,11 +35,11 @@ impl Pipe {
         use procfs::process::{FDTarget, Process};
         let proc = Process::new(pid as i32).map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
-        for fd in proc.fd()? {
-            let fd = fd?;
+        for fd in proc.fd().map_err(|err| io::Error::other(err))? {
+            let fd = fd.map_err(|err| io::Error::other(err))?;
             if let FDTarget::Path(fd_path) = fd.target {
                 // Resolve the path exactly as the kernel reports it
-                if fd_path == *path {
+                if fd_path == self.path {
                     return Ok(true);
                 }
             }
@@ -50,6 +49,7 @@ impl Pipe {
 
     #[cfg(target_os = "macos")]
     pub fn is_path_open(&self, pid: u32) -> io::Result<bool> {
+        use proc_pidinfo::*;
         let pid = Pid(pid);
         for fd in proc_pidinfo_list::<ProcFDInfo>(pid)? {
             match proc_pidfdinfo::<VnodeFdInfoWithPath>(pid, fd.proc_fd)? {


crates/axl-runtime/src/builtins/aspect/build.axl
```diff
@@ -27,8 +27,16 @@ def impl(ctx: TaskContext) -> int:
...
     build = ctx.bazel.build(
-        events = events,
-        bazel_flags = ["--isatty=" + str(int(stdout.is_tty))]
+        build_events = True,
+        workspace_events = True,
+        execution_logs = True,
+        flags = [
+            "--isatty=" + str(int(stdout.is_tty)),
+            "--bes_outerr_buffer_size=1024",
+            "--bes_outerr_chunk_size=1024",
```
arrdem: Why only 1k? 4k is a page and the smallest typical buffer size.

crates/axl-runtime/src/engine/bazel/iter/build_event.rs
```diff
@@ -0,0 +1,94 @@
...
+impl<'v> AllocValue<'v> for BuildEventIterator {
+    fn alloc_value(self, heap: &'v Heap) -> values::Value<'v> {
+        heap.alloc_complex_no_freeze(self)
+    }
+}
+
+#[starlark_module]
+pub(crate) fn build_event_methods(registry: &mut MethodsBuilder) {
+    /// Returns `BuildEvent` if event buffer is not empty.
+    /// Maximum `1000` events is buffered at once.
```
arrdem, line 48: What happens if the buffer is full? BuildEvents are being delivered to a file, so we just all behind? Or is there backpresure?

arrdem (approved):

crates/axl-runtime/src/builtins/aspect/build.axl
thesayyn: This change can be reverted, i only made it smaller for frequent flushes.

crates/axl-runtime/src/engine/bazel/iter/build_event.rs
thesayyn: in this case, its fifo so we block writes into the pipe by bazel, the thread reading & processing the events blocks until the buffer is empty. 
https://docs.rs/fibre/latest/fibre/spmc/index.html.  

Using a file in this case is slow since it involves IO.

thesayyn: blocking of the writes to pipe does not prevent Bazel from doing work because of  `--build_event_binary_file_upload_mode=fully_async`

thesayyn: I can't say the same for execlog and workspace events.

crates/axl-runtime/src/builtins/aspect/build.axl
thesayyn: reverted

PR #902, opened 2025-11-21
Add support for interactive child processes

Silic0nS0ldier:
Adds `status` function to `Command` instances to allow running of child processes with `stdin` connected, such as a shell like `bash`. Unlike `spawn` this implicitly waits for the child process to exit.

The function's name `status` is borrowed from the Rust standard library.

---

  Support for interactive child processes via `ctx.std.process.command("APP_NAME").status()`.

### Test plan

- ~Covered by existing test cases~ New API surface.
- ~New test cases added~ No test cases for `spawn()`, so nothing to build upon.
  ```starlark
  def _impl(ctx):
      shell = ctx.std.env.var("SHELL") or "sh"
      print("Running " + shell)
      cmd = ctx.std.process.command(shell)
      result = cmd.status()
      print(shell + " finished with exit code " + str(result.code))
  test_task = task(
     name = "test-task",
     implementation = _impl,
  )
  ```
  Then integrate the task with `MODULE.aspect` and run `aspect test-task`.
  When in the interactive shell use `exit` (or whatever applies to your shell) to resume the task.

### DIFF
--- a/MODULE.aspect
+++ b/MODULE.aspect
@@ -13,7 +13,7 @@ axl_local_dep(
 axl_archive_dep(
     name = "aspect_rules_lint",
     urls = ["https://github.com/aspect-build/rules_lint/archive/65525d871f677071877d3ea1ec096499ff7dd147.tar.gz"],
-    integrity = "sha512-IKsrPMgaNuzHAkHzyKbJexzt8ObepFhtzyurirppI1XNHgMKrLMuPXXVFeOpoVws5o2EMUSY4ZisCz9SLFKbhg==",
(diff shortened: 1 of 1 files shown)


Silic0nS0ldier:
Motivation for this contribution is to allow porting of an edit patches workflow over to Aspect CLI.

An interactive shell (running in a temporary directory) is used to allow users a chance to edit and create new commits, on exit (provided there are no uncommitted changes) patches are generated from commits on top of the baseline (a branch, tag, or an initial generated commit when working on source archives).

alexeagle:
@thesayyn what's the next step here?

PR #911, opened 2025-12-10
Binding

thesayyn:
Adds support for `binding` in config.axl


thesayyn:
Updated content v2

thesayyn:
Updated content v2

PR #992, opened 2026-04-10
feat: add task name & id

gregmagolan:


crates/aspect-cli/src/builtins/aspect/feature/defaults.axl
```diff
@@ -13,6 +13,12 @@ def _default_bazel_behavior(ctx: FeatureContext):
     health_check = ctx.traits[HealthCheckTrait]
     bazel_fragment = ctx.traits[BazelTrait]
 
+    def _buildkite_bazel_section(ctx, state):
+        if ctx.std.env.var("BUILDKITE"):
```
thesayyn: for this you can check the `environment` variable above which tells you which CI you are on.

thesayyn (approved):

PR #995, opened 2026-04-11
Don't create a shared state in the BazelTrait

gregmagolan:


crates/aspect-cli/src/builtins/aspect/feature/artifacts.axl
```diff
@@ -102,7 +102,7 @@ def _artifact_upload_impl(ctx: FeatureContext):
 
     Uses ctx.task.name and ctx.task.key to identify the current task invocation.
     """
-    fragment = ctx.traits[BazelTrait]
+    bazel = ctx.traits[BazelTrait]
```
thesayyn: using name bazel overrides the global symbol bazel. lets use bazel_fragment

thesayyn (approved):

gregmagolan:
search-computer

PR #998, opened 2026-04-12
Unify args & attrs + AXL naming conventions (snake_case tasks, CamelCase features/traits, kebab-case CLI)

gregmagolan:
## Summary

This PR consolidates the task/feature argument system under a single `args` namespace and establishes clear boundaries between the three types of definition objects.

### `args` everywhere for tasks and features

Tasks and features now use `args = {}` exclusively. CLI-typed args use `args.string/int/bool/…`; config-only args use the new `args.custom()`. Both are accessed uniformly via `ctx.args.name` in implementations.

```python
greet = task(
    group = ["utils"],
    summary = "Say hello to someone",
    implementation = _impl,
    args = {
        # CLI arg — set via --recipient on the command line or in config.axl; CLI wins
        "recipient": args.string(default = "world", description = "Who to greet"),
        # Config-only arg — not shown in --help; configurable in config.axl only
        "greeting":  args.custom(str, default = "Hello", description = "Greeting word"),
    },
)

def _impl(ctx: TaskContext) -> int:
    ctx.std.io.stdout.write(ctx.args.greeting + ", " + ctx.args.recipient + "\n")
    return 0
```

Features work the same way — the separate `attrs` (config-only) and `args` (CLI) parameters are merged into a single `args` dict:

```python
ArtifactUpload = feature(
    implementation = _impl,
    args = {
        # Config-only arg — not shown in --help; configurable in config.axl
        "bucket": args.custom(str | None, default = None, description = "S3 bucket name"),
        # CLI arg — injected as --artifact-upload-mode on every task subcommand
        "mode":   args.string(default = "auto", description = "Upload mode"),
    },
)
```

### `attr()` is traits-only

`attr()` is now restricted to `trait()` field definitions and no longer used by Features.

```python
BazelTrait = trait(
    extra_flags    = attr(list[str], default = [],    description = "Extra Bazel flags for every build"),
    profile_upload = attr(bool,      default = False, description = "Upload Bazel profile after build"),
)
```

### `args.custom()` — new config-only arg

`args.custom(type, default=..., description=...)` is the tasks/features equivalent of `attr()`. It accepts any Starlark type expression, is never surfaced as a CLI flag, and can only be set via `config.axl`. In the future, we plan to support setting custom de-marshalers so that users can add custom flags types to their CLI.

### Configuring tasks from `config.axl`

`ctx.tasks` in `config.axl` is a map keyed by task path. The path is the task's group and name joined with `/`:

| Task definition | Key |
|---|---|
| `foo = task(...)` (root) | `"foo"` |
| `bar = task(group = ["grp"], ...)` | `"grp/bar"` |
| `baz = task(group = ["a", "b"], ...)` | `"a/b/baz"` |

```python
def config(ctx: ConfigContext):
    t = ctx.tasks["user/nested/greet"]
    t.args.greeting = "Bonjour"     # override a custom (config-only) arg
    t.args.recipient = "monde"      # override a CLI arg's effective default

    for t in ctx.tasks:
        if t.key.startswith("ci/"):
            t.args.upload_bucket = "ci-artifacts"
```

**Precedence at execution time** (highest wins):

```
explicit --flag on CLI  >  config.axl override  >  task definition default
```

`--help` shows effective defaults *after* `config.axl` runs — overrides are reflected in the help output.

### `long` override for CLI args

All `args.string/int/bool/…` functions accept an optional `long = "override_name"` parameter to control the `--flag-name` shown on the CLI, bypassing the default snake_case → kebab-case derivation.

### Feature CLI arg constraints

Feature CLI args are injected into **every** task subcommand:

- **Positional args** are not allowed — they collide with task-level positional args.
- **`required = true` is not allowed** — it would make every task fail unless the flag is always provided.

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/feature/gh_status_checks.axl
+++ b/crates/aspect-cli/src/builtins/aspect/feature/gh_status_checks.axl
@@ -41,5 +41,10 @@ def _github_status_checks(ctx: FeatureContext):
 
 GithubStatusChecks = feature(
     implementation = _github_status_checks,
-    attrs = {}
+    args = {
+        "mode": args.string(
+            default = "ci-only",
+            description = "When to create GitHub Status Checks",
+        ),
+    },
 )
--- a/crates/aspect-cli/src/builtins/aspect/traits.axl
+++ b/crates/aspect-cli/src/builtins/aspect/traits.axl
@@ -4,31 +4,31 @@ load("./bazel.axl", "default_retry")
 
 BazelTrait = trait(
     # Declarative data — composable, zero-cost reads
-    extra_flags = attr(list[str], []),
-    extra_startup_flags = attr(list[str], []),
-    build_event_sinks = attr(list[bazel.build.BuildEventSink], []),
-    execution_log_sinks = attr(list[bazel.execution_log.ExecLogSink], []),
+    extra_flags = attr(list[str], default = [], description = "Additional Bazel flags appended to every build/test invocation"),
+    extra_startup_flags = attr(list[str], default = [], description = "Additional Bazel startup flags prepended before the command"),
+    build_event_sinks = attr(list[bazel.build.BuildEventSink], default = [], description = "Build Event Protocol sinks to stream events to during the build"),
+    execution_log_sinks = attr(list[bazel.execution_log.ExecLogSink], default = [], description = "Execution log sinks to receive the Bazel execution log after the build"),
 
     # Optional transforms — only called when set
-    flags = attr(typing.Callable[[list[str]], list[str]] | None, None),
-    startup_flags = attr(typing.Callable[[list[str]], list[str]] | None, None),
+    flags = attr(typing.Callable[[list[str]], list[str]] | None, default = None, description = "Transform function called with the full flag list; return value replaces the list"),
+    startup_flags = attr(typing.Callable[[list[str]], list[str]] | None, default = None, description = "Transform function called with the full startup flag list; return value replaces the list"),
 
     # Lifecycle hooks — lists of callables; callers close over their own state
-    build_start = attr(list[typing.Callable[[TaskContext], None]], []),
-    build_event = attr(list[typing.Callable[[TaskContext, dict], None]], []),
-    build_retry = attr(typing.Callable[[int], bool], default_retry),
-    build_end = attr(list[typing.Callable[[TaskContext, int], None]], []),
+    build_start = attr(list[typing.Callable[[TaskContext], None]], default = [], description = "Hooks called once before the Bazel invocation starts"),
+    build_event = attr(list[typing.Callable[[TaskContext, dict], None]], default = [], description = "Hooks called for each Build Event Protocol event received during the build"),
+    build_retry = attr(typing.Callable[[int], bool], default = default_retry, description = "Predicate called with the exit code; return True to retry the build"),
+    build_end = attr(list[typing.Callable[[TaskContext, int], None]], default = [], description = "Hooks called once after the Bazel invocation completes, with the exit code"),
 )
 
 
 HealthCheckTrait = trait(
-    pre_health_check = attr(list[typing.Callable[[TaskContext], None]], []),
-    post_health_check = attr(list[typing.Callable[[TaskContext], str | None]], []),
+    pre_health_check = attr(list[typing.Callable[[TaskContext], None]], default = [], description = "Hooks called before the build to validate preconditions (e.g. auth, connectivity)"),
+    post_health_check = attr(list[typing.Callable[[TaskContext], str | None]], default = [], description = "Hooks called after the build; return a non-None string to report a warning"),
 )
 
 
 DeliveryTrait = trait(
-    delivery_start = attr(typing.Callable[[], None], lambda: None),
-    delivery_end = attr(typing.Callable[[], None], lambda: None),
-    deliver_target = attr(typing.Callable[[str, bool], None], lambda label, is_forced: None),
+    delivery_start = attr(typing.Callable[[], None], default = lambda: None, description = "Hook called once at the start of a delivery run, before any targets are delivered"),
+    delivery_end = attr(typing.Callable[[], None], default = lambda: None, description = "Hook called once at the end of a delivery run, after all targets are delivered"),
+    deliver_target = attr(typing.Callable[[str, bool], None], default = lambda label, is_forced: None, description = "Hook called for each target being delivered; receives the label and a forced-delivery flag"),
 )
--- a/crates/aspect-cli/src/cmd_tree.rs
+++ b/crates/aspect-cli/src/cmd_tree.rs
@@ -1,7 +1,10 @@
 use std::collections::HashMap;
 
+use axl_runtime::engine::arg::Arg;
 use axl_runtime::engine::task::{MAX_TASK_GROUPS, TaskLike};
-use clap::{Arg, ArgMatches, Command, value_parser};
+use axl_runtime::engine::types::feature::to_display_name;
+use clap::{Arg as ClapArg, ArgMatches, Command, value_parser};
+use starlark::collections::SmallMap;
 use thiserror::Error;
 
 const TASK_ID: &'static str = "@@@$'__AXL_TASK_ID__'$@@@";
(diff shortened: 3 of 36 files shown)


crates/aspect-cli/src/builtins/aspect/feature/gh_status_checks.axl
```diff
@@ -41,5 +41,10 @@ def _github_status_checks(ctx: FeatureContext):
 
 GithubStatusChecks = feature(
     implementation = _github_status_checks,
-    attrs = {}
+    attrs = {
+        "mode": args.string(
```
gregmagolan, line 45: This is just proving we can add an arg to a Feature. Full impl of GitHub Status Checks is in flight in another PR

crates/axl-runtime/src/engine/types/trait.rs
```diff
@@ -17,194 +15,18 @@ use starlark::values::{
...
+
 static TRAIT_TYPE_ID: AtomicU64 = AtomicU64::new(0);
 
 fn next_trait_type_id() -> u64 {
     TRAIT_TYPE_ID.fetch_add(1, Ordering::SeqCst)
 }
 
-/// A field definition for a trait, containing a type and optional default value.
-#[derive(Debug, Clone, ProvidesStaticType, Allocative)]
-pub struct Field<'v> {
```
gregmagolan, line 28: renamed to ConfigAttr moved to its own attr.rs

crates/axl-runtime/src/engine/cli_arg.rs
```diff
@@ -81,32 +81,52 @@ pub enum TaskArg {
...
-impl<'v> StarlarkValue<'v> for TaskArg {}
+/// A CLI argument definition — the result of calling `args.string(...)`, `args.int(...)`, etc.
+#[starlark_value(type = "args.CliArg")]
+impl<'v> StarlarkValue<'v> for CliArg {}
 
-starlark_simple_value!(TaskArg);
+starlark_simple_value!(CliArg);
 
-impl Display for TaskArg {
+impl CliArg {
```
gregmagolan: Args are used by tasks and features now so renamed to CliArg

crates/aspect-cli/src/cmd_tree.rs
```diff
@@ -200,9 +245,48 @@ pub fn make_command_from_task(
...
 
-    for (name, arg) in task.args() {
-        let arg = crate::flags::convert_arg(&name, arg);
-        cmd = cmd.arg(arg);
+    for (arg_name, arg) in task.cli_attrs() {
+        let heading = format!("{} Options", task_display);
+        let long_name = arg_name.replace('_', "-");
+        let mut clap_arg = crate::flags::convert_arg(arg_name, &long_name, arg).help_heading(heading);
+        if let Some(effective) = effective_defaults.get(arg_name) {
+            clap_arg = clap_arg.default_value(effective.clone());
```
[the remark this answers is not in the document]
gregmagolan: Resolved

crates/axl-runtime/src/engine/config/tasks/configured_task.rs
```diff
@@ -244,18 +315,200 @@ impl<'v> values::StarlarkValue<'v> for FrozenConfiguredTask {
...
+            .downcast_ref::<FrozenTask>()
+            .and_then(|t| t.attrs().get(attribute))
+            .is_some();
+        if !attr_exists {
+            return ValueError::unsupported(self, &format!(".{}=", attribute));
+        }
+
+        let frozen = freeze_value(value)
+            .map_err(|e| starlark::Error::new_other(anyhow!("{:?}", e)))?;
+        ct.config_overrides.borrow_mut().insert(attribute.to_owned(), frozen);
```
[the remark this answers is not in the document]
gregmagolan: Resolved

crates/axl-runtime/src/engine/cli_arg.rs
thesayyn: we should name it to Arg i think

crates/axl-runtime/src/engine/types/trait.rs
gregmagolan: Out of date

crates/axl-runtime/src/engine/cli_arg.rs
gregmagolan: Sure.. sgtm

gregmagolan: Done

crates/aspect-cli/src/builtins/aspect/traits.axl
```diff
@@ -4,31 +4,31 @@ load("./bazel.axl", "default_retry")
...
+    extra_flags = attr(list[str], default = [], description = "Additional Bazel flags appended to every build/test invocation"),
+    extra_startup_flags = attr(list[str], default = [], description = "Additional Bazel startup flags prepended before the command"),
+    build_event_sinks = attr(list[bazel.build.BuildEventSink], default = [], description = "Build Event Protocol sinks to stream events to during the build"),
+    execution_log_sinks = attr(list[bazel.execution_log.ExecLogSink], default = [], description = "Execution log sinks to receive the Bazel execution log after the build"),
 
     # Optional transforms — only called when set
-    flags = attr(typing.Callable[[list[str]], list[str]] | None, None),
-    startup_flags = attr(typing.Callable[[list[str]], list[str]] | None, None),
+    flags = attr(typing.Callable[[list[str]], list[str]] | None, default = None, description = "Transform function called with the full flag list; return value replaces the list"),
+    startup_flags = attr(typing.Callable[[list[str]], list[str]] | None, default = None, description = "Transform function called with the full startup flag list; return value replaces the list"),
```
thesayyn, line 14: ah attrs have description now?  i thought trait/attr system would be unchanged by this. i don't how this is needed. also is the default (named) parameter change needed?

crates/axl-runtime/src/engine/config/tasks/configured_task.rs
```diff
@@ -244,18 +316,232 @@ impl<'v> values::StarlarkValue<'v> for FrozenConfiguredTask {
...
+                }
+            }
+        }
+        attrs
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Default-value materialisation for Arg
+// ---------------------------------------------------------------------------
```
thesayyn, line 354: only if we could get claude to stop putting these banner style comments

crates/axl-runtime/src/engine/config/tasks/configured_task.rs
```diff
@@ -244,18 +316,232 @@ impl<'v> values::StarlarkValue<'v> for FrozenConfiguredTask {
...
+                .collect();
+            heap.alloc(AllocList(items))
+        }
+        Arg::TrailingVarArgs { .. } => heap.alloc(AllocList(Vec::<&str>::new())),
+        Arg::StringList { default, .. } => heap.alloc(AllocList(
+            default.iter().map(|s| s.as_str()).collect::<Vec<_>>(),
+        )),
+        Arg::BooleanList { default, .. } => heap.alloc(AllocList(default.iter().copied())),
+        Arg::IntList { default, .. } => heap.alloc(AllocList(default.iter().copied())),
+        Arg::UIntList { default, .. } => heap.alloc(AllocList(default.iter().copied())),
```
thesayyn, line 383: .copied here probably extra allocation

crates/aspect-cli/src/builtins/aspect/traits.axl
gregmagolan: I prefer to have it named yes. It reads much better and is more consistent with what users are already doing with args where the default is named.

We already have breaking changes since the last release by renaming fragment => trait so we're free to change the API of trait here without extra churn for users.

gregmagolan: I also added summary and description named params so we can have nice generated docs for Traits in the future.

crates/axl-runtime/src/engine/config/tasks/configured_task.rs
gregmagolan: Claude did some things and improved it

crates/axl-runtime/src/engine/config/tasks/configured_task.rs
gregmagolan: 🤷‍♂️  there are worse things

thesayyn (approved):

PR #999, opened 2026-04-13
feat: synthetic @std// module

thesayyn:

New @std module that implements a few useful standard library functions that does not need to be in the global scope or dangle off of `ctx`.  

`@std//base64.axl` for working with base64 data

`@std//hash.axl` for working with `sha*` and `blake2b` and `blake2s` hash algorithms. 

`@std//time.axl` for `time.sleep()` and `time.sleep_iter()`

### Test plan

- New test cases added


thesayyn:
@codex review

PR #1012, opened 2026-04-21
Check sha256 of tar tool before use

gregmagolan:


crates/aspect-cli/src/builtins/aspect/lib/tar.axl
```diff
@@ -41,13 +41,17 @@ def bsdtar(ctx):
 
     os, arch = _platform_key(ctx)
     bin_path = cache + "/tar"
+    sha256 = _SHA256.get(os + "-" + arch)
 
     if ctx.std.fs.exists(bin_path):
-        return bin_path
+        actual = ctx.std.fs.sha256(bin_path)
```
thesayyn: there is a `load("@std//hash.axl", "sha256")` we can use here

gregmagolan: there is... but then we'd need to load the file into memory and pass the bytes to the hash function

gregmagolan: an fs.sha255 is more efficient

thesayyn: hmm we should extend it then to handle streams. i don't really want to have two different ways to do this.

thesayyn: i mean also we can do the thing that fs.sha256 does with a stream loop. we have partially load it onto the memory anyway.

thesayyn: we also had ctx.http.download(integrity) which should be handling this...

thesayyn: maybe it can redownload if the output exists but does not match the content.

gregmagolan: sure. sounds good to me. please update the API to how you'd like it. we should make this fix now tho as the broken tar binary will be a real issue on persistent runners if we don't fix it.

gregmagolan: I don't think the `ctx.http.download(integrity)` impl was the problem.. that correctly checks the integrity after download. I think the issue is the a job can be cancelled mid-download and the file is half downloaded which will break the next job.

thesayyn: okay, all good if you want to land as is.

gregmagolan: I'll give Claude a shot at a streaming API in hash.axl

gregmagolan:
```
def _file_sha256(ctx, path):
    h = sha256()
    for chunk in ctx.std.fs.open(path):
        h.update(chunk)
    return h.hexdigest()
```
should be fast enough for a small file. For larger files, we'd be disk bound anyway

PR #1045, opened 2026-04-28
gazelle: add 'aspect gazelle' task + prebuilt aspect-gazelle wiring + CI integration

gregmagolan:
Adds `aspect gazelle` as a built-in task — the BUILD-file equivalent of `aspect format`.

Also wires the task into all four CI providers in this repo.


jbedard:
Is this for devs or for CI? On CI this should be using `--mode=diff` and checking the exit code + stdout

crates/aspect-cli/src/builtins/aspect/gazelle.axl
```diff
@@ -0,0 +1,200 @@
...
+    if pre_snap == None:
+        out.write("\x1b[0;33mWARNING\x1b[0m: cannot resolve HEAD (no initial git commit, or not in a git repository); gazelle-required detection is skipped. Gazelle ran, but we cannot tell whether it changed anything.\n")
+        out.write("Make at least one commit (or run inside a git repo) and re-run if you want `aspect gazelle` to surface gazelle-induced changes.\n")
+        return 0
+
+    # Diff the post-gazelle tree against the snapshot. Non-empty = gazelle
+    # updated BUILD files = "BUILD files out of date".
+    gazelle_diff = _diff_against(ctx, pre_snap)
+    if not gazelle_diff:
+        return 0
```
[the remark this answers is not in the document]
gregmagolan: Nice catch. Resolved.

gregmagolan:
> Is this for devs or for CI? On CI this should be using `--mode=diff` and checking the exit code + stdout

For both

crates/aspect-cli/src/builtins/aspect/gazelle.axl
```diff
@@ -0,0 +1,312 @@
...
+        "check": args.boolean(
+            default = False,
+            description = "When true, only report the diff — don't apply it. Recommended in CI: faster (skips the apply step), leaves the tree clean (no risk of a half-formatted state on mid-run interrupt), and the verdict comes from gazelle directly without depending on the workspace being a git working tree. Local devs typically leave this off so `aspect gazelle` updates BUILD files in place.",
+        ),
+        "upload_gazelle_diff": args.boolean(
+            default = False,
+            description = "Upload gazelle's diff as a CI artifact (named `gazelle.patch`) so reviewers can apply it locally without re-running gazelle. No-op when the working tree is clean (no diff to upload), or when not running on a recognized CI host (GitHub Actions, Buildkite, CircleCI, GitLab). The download URL (where the CI provider returns one) is surfaced via ArtifactsTrait.artifact_urls['gazelle_diff']. Off by default.",
+        ),
+        "gazelle_target": args.string(
+            default = "//tools/gazelle:gazelle",
```
jbedard, line 353: The standard is still `//:gazelle` in the root, as recommended by gazelle itself and the aspect-gazelle docs do the same...

gregmagolan: Hmm. I don't love that we're diverging from what rules_lint suggests and I'd rather be consistent within our own rule sets. I think we should switch the standard for this task. There is good reason to since load statements in the root BUILD file can easily leak into longer fetch & analysis times. This is why the format task ended up in //tools/format and lint in //tools/lint

gregmagolan: This new implementation of the gazelle task gives us a chance to make that breaking change.

gregmagolan: Users are going to call `aspect gazelle` locally anyway so no one will ever have to type `bazel run //tools/gazelle`

jbedard:
If this is designed to replace `bazel run //:gazelle -- *` then it needs to support the `*` in there which can be anything. The `--mode` is one CLI arg but you can also pass other generic gazelle CLI args, aspect-gazelle enhancements (`--cache=X`, `--progress` etc) or an arbitrary list of directories to update.

Should `--mode` be special like you've done here or should it just be handled in the `*`?

crates/aspect-cli/src/builtins/aspect/gazelle.axl
jbedard: I'd be open to changing the aspect-gazelle recommendations+examples due to the fetch & analysis perf reasons, but that's not what exists today.

crates/aspect-cli/src/builtins/aspect/format.axl
```diff
@@ -146,7 +146,6 @@ def _impl(ctx: TaskContext) -> int:
         "--remote_download_outputs=toplevel",
         "--build_runfile_links",
         "--experimental_build_event_upload_strategy=local",
-        "--noallow_analysis_cache_discard",
```
jbedard: This seems unrelated

bazel/platforms/BUILD.bazel
```diff
@@ -22,8 +22,8 @@ alias(
 
 platform(
     name = "linux_host_platform",
-    parents = ["@platforms//host"],
```
jbedard: This PR has a mess of unrelated changes?

jbedard:
Also note the gazelle cli is `gazelle [fix|update|update-repos] [--mode=fix] [--other-args] ...dirs to update...`.

I think the `update-repos` command unnecessary on the CLI, and I've made that assumption in the `aspect_gazelle_prebuilt`, but the `fix|update` is is still possible.

crates/aspect-cli/src/builtins/aspect/gazelle.axl
```diff
@@ -0,0 +1,312 @@
...
+        gazelle = gazelle[:-len("-runner.bash")]
+
+    # Diff phase: run with -mode=diff to get the verdict and the patch
+    # without modifying the tree. The standard aspect_gazelle wrapper
+    # hardcodes `-mode fix`; Gazelle's flag parser treats this second
+    # `-mode` as a last-wins override (BEHAVIOR switches to diff —
+    # tree is not modified, unified diff prints on stdout). Exit code
+    # is 0 in this duplicate-flag config even when the diff is
+    # non-empty, so we drive the verdict off stdout content.
+    diff_proc = r.spawn(gazelle, ["-mode=diff"], capture = True)
```
jbedard: This means we are no longer invoking via `bazel run`? As long as we are using the generated shell script that _should_ be fine, I just want to make sure we don't cause the same issues as the `configure` command in CLI 1 where we lost all bazel setup such as toolchains.

See https://github.com/bazel-contrib/bazel-gazelle/blob/493e9d9072eace799138d5cc1188f9e4dcb22061/internal/gazelle.bash.in#L28-L38

crates/aspect-cli/src/builtins/aspect/gazelle.axl
```diff
@@ -0,0 +1,312 @@
...
+    # the default `-p1` would strip the leading directory and look
+    # for `gazelle/testdata/BUILD.bazel`, which doesn't exist.
+    #
+    # `--check` skips this entirely — CI mode doesn't want a dirtied
+    # tree.
+    if not ctx.args.check:
+        tmp = ctx.std.env.temp_dir()
+        patch_path = tmp + "/gazelle-" + ctx.task.key + ".patch"
+        ctx.std.fs.write(patch_path, diff_stdout)
+        apply_result = ctx.std.process.command("git").args(["apply", "-p0", patch_path]) \
```
jbedard, line 313: We're invoking `git` on CI? 🫣

bazel/platforms/BUILD.bazel
gregmagolan: It ran gazelle :)

crates/aspect-cli/src/builtins/aspect/gazelle.axl
gregmagolan: `git` is a command that we can assume will be available :)

gregmagolan: We already rely on it in other tasks to find changed files in the case we are not in a PR or don't have access to the GitHub API to query PR changed files

crates/aspect-cli/src/builtins/aspect/format.axl
gregmagolan: Was a build failure on CI. I can pre-factor

crates/aspect-cli/src/builtins/aspect/gazelle.axl
gregmagolan: Good call. I'll look into it. In general our CLI runs binaries directly instead of using `bazel run` so I'd like to keep that pattern here if we can. It is a simpler mental model to build the binary first and then spawn it directly.

gregmagolan:
> If this is designed to replace `bazel run //:gazelle -- *` then it needs to support the `*` in there which can be anything. The `--mode` is one CLI arg but you can also pass other generic gazelle CLI args, aspect-gazelle enhancements (`--cache=X`, `--progress` etc) or an arbitrary list of directories to update.
> 
> Should `--mode` be special like you've done here or should it just be handled in the `*`?

We want to be a little more opinionated so if there are gazelle flags that users want to pass then I think we should mirror them in the task flags or add a generic --gazelle_flag that mirrors the --bazel_flag flag we have in build & test tasks.

crates/aspect-cli/src/builtins/aspect/gazelle.axl
jbedard: If it is running that ugly `gazelle.bash` script then I think you're good?

As opposed to CLI1 where we were just invoking gazelle via the go API and zero bazel was involved, so no `GOTOOL` from the `rules_go` toolchains or `REPO_CONFIG_PATH` from the `gazelle` module etc.

crates/aspect-cli/src/builtins/aspect/gazelle.axl
jbedard: If you're already doing it then this comment is me just being curious... 👍

jbedard:
We need the ability to pass a list of directories to gazelle, that is essential. If you want to restrict all other args they can be hidden behind task args (basically duplicating the gazelle cli args), or unsupported for less used ones.

Some args are part of the DX and should be easily accessible (all aspect-gazelle enhancements):
* `--progress`
* `--cache[=type]`
* `--watch` should be a shortcut for `aspect run --watch //tools/gazelle` (or whatever we decide for watch mode in cli2)

Some args you can hide behind task definitions, or even tell people to put it into their `gazelle()` target definition:
* `--mode`
* `--index[=lazy]` (if all your languages support "lazy indexing" then you can normally hardcode this in the `gazelle()` def)
* `command` - the core command being run, normally `update` but some gazelle languages do additional changes if you use `fix` instead. I actually don't know if users would toggle this or can hardcode it in their BUILD.

Some args are odd/rare and we can just tell users to manually `run //tools/gazelle` for them:
* `-cpuprofile|memprofile` etc for testing
* `-repo_config` (normally the default/automatic is enough or gets hardcoded in the `gazelle()` target)

Things that might change depending on who is invoking it under which circumstances:
* `-r[=true|false]`: to recurse or not (into the varargs list of directories)
* the varargs of which directories to update

crates/aspect-cli/src/builtins/aspect/gazelle.axl
gregmagolan: Lets change it then :) No time like the present

crates/aspect-cli/src/builtins/aspect/gazelle.axl
gregmagolan: Yea, its running the ugly gazelle.bash script.

bazel/platforms/BUILD.bazel
gregmagolan: Pre-factored

crates/aspect-cli/src/builtins/aspect/gazelle.axl
gregmagolan: I added the rationale for the default target switch to the migration docs : https://docs.staging.aspect.build/cli/migration/gazelle#gazelle-target

gregmagolan: https://docs.aspect.build/cli/migration/gazelle#gazelle-target (on https://github.com/aspect-build/docs/pull/206 lands)

crates/aspect-cli/src/builtins/aspect/gazelle.axl
gregmagolan: To be specific:

```
    gazelle = r.determine_entrypoint(ctx.args.gazelle_target)
    if not gazelle:
        fail("Failed to determine the gazelle target's entrypoint.")

    # Upstream `gazelle()` (bazel-gazelle, used by `gazelle_from_source`-
    # style targets) produces both a `<name>` sh_binary and an inner
    # `<name>-runner.bash` script — and `determine_entrypoint` may
    # surface either depending on BES ordering. The inner runner
    # script's `$0`-relative runfiles lookup expects to find a sibling
    # `<name>-runner.bash.runfiles` dir, but bazel actually creates
    # the runfiles tree as `<name>.runfiles` (matching the sh_binary).
    # Stripping the suffix points us at the sh_binary symlink instead,
    # whose runfiles dir resolves correctly.
    if gazelle.endswith("-runner.bash"):
        gazelle = gazelle[:-len("-runner.bash")]
```

gregmagolan:
> We need the ability to pass a list of directories to gazelle, that is essential. If you want to restrict all other args they can be hidden behind task args (basically duplicating the gazelle cli args), or unsupported for less used ones. ...

Addressed with https://github.com/aspect-build/aspect-cli/pull/1045/changes/2acd342a9591519d09d4d4463cadb718e404e252.

- **`--gazelle-flag`** (repeatable, mirrors `--bazel-flag`) — covers `--progress`, `--cache`, `-r`, and anything else. Flags specific to `aspect_gazelle_prebuilt` work fine here; from-source users just don't pass them.
- **Directories** — positional varargs, forwarded as trailing args to gazelle. `aspect gazelle tools/go services/api` works.
- **`--gazelle-command`** — `fix` or `update` subcommand (default: none, gazelle's own default applies).
- **`--mode`** stays internal — any `-mode`/`--mode` passed via `--gazelle-flag` is ignored, since the task controls mode and applies changes via `git apply` rather than re-invoking gazelle.
- **`--watch`** deferred — it's a fundamentally different execution model (long-lived daemon). Leaving it for a dedicated feature later.
- **`update-repos`** not supported here — its interaction with `-mode=diff` is untested. Documented to use `bazel run` directly.

PR #1056, opened 2026-04-30
feat: support trace api

thesayyn:

Telemetry endpoints now can be configured via config.axl or features and by default the Telemetry feature adds the standard otel env.

### Test plan
- New test cases added


thesayyn:
@codex review

crates/aspect-cli/src/builtins/aspect/feature/telemetry.axl
```diff
@@ -0,0 +1,220 @@
+"""Default telemetry feature.
+
+Registers exporters from three sources:
+
+  0. ASPECT_DEBUG=1 — always (regardless of `discover_endpoints`) wires up
```
gregmagolan: lol. 0 based list :)

gregmagolan: a fourth source would be `ctx.telemetry.exporters.add` like we set in config.axl?

gregmagolan: ahh. you give it a footnote in the 3rd method : "callers needing per-signal routing or resource_attributes should call `ctx.telemetry.exporters.add(...)` directly from config.axl or a custom feature". 

should it gets its own mention?

crates/aspect-cli/src/builtins/aspect/feature/telemetry.axl
```diff
@@ -0,0 +1,220 @@
...
+     Endpoints registered this way use the default protocol (grpc) and all
+     signals; callers needing per-signal routing or resource_attributes
+     should call `ctx.telemetry.exporters.add(...)` directly from
+     config.axl or a custom feature.
+
+Future versions of this feature may layer additional default behaviors on
+top (detected CI-host endpoints, sampling defaults, etc.).
+
+Environment variables read when `discover_endpoints = True`:
+  OTEL_EXPORTER_OTLP_ENDPOINT          — single endpoint for all signals
```
gregmagolan, line 58: is this env var pattern a standard approach or are we inventing something new here?

crates/aspect-cli/src/builtins/aspect/feature/telemetry.axl
```diff
@@ -0,0 +1,220 @@
...
+  OTEL_EXPORTER_OTLP_TRACES_ENDPOINT   — traces-only override
+  OTEL_EXPORTER_OTLP_LOGS_ENDPOINT     — logs-only override
+  OTEL_EXPORTER_OTLP_METRICS_ENDPOINT  — metrics-only override
```
gregmagolan, line 61: what happens when you mix them such as setting both OTEL_EXPORTER_OTLP_ENDPOINT & OTEL_EXPORTER_OTLP_TRACES_ENDPOINT?

crates/aspect-cli/src/builtins/aspect/feature/telemetry.axl
thesayyn: these are standard envs, we just recognize them

thesayyn: https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/

crates/aspect-cli/src/builtins/aspect/feature/telemetry.axl
thesayyn: fixed, this was wrong. OTEL_EXPORTER_OTLP_TRACES_ENDPOINT override catch all ones

crates/aspect-cli/src/builtins/aspect/feature/telemetry.axl
thesayyn: fixed

crates/aspect-cli/src/trace_buffer.rs
```diff
@@ -0,0 +1,232 @@
+//! In-memory buffering OTLP exporters with late attachment.
```
gregmagolan, line 1: Does aspect-cli process wait to drain this buffer before exiting? Can we output something if we're waiting after task is done so user does not think process is hanged? Should we add a configurable timeout so it doesn't wait forever?

gregmagolan (approved):

crates/aspect-cli/src/trace_buffer.rs
thesayyn: specifically for this buffer, its pre feature/config.axl eval buffer which we flush immediately after knowing where to send the data, as for process end flushing there is 5 seconds timeout at which point all the unsent traces will be dropped

thesayyn: i did make an addition to surface those timeout errors since there is no way to know them from axl side

PR #1082, opened 2026-05-14
feat(gazelle): parity with format — --scope, --soft-fail, --check 3-state, multi-dir

gregmagolan:
## Summary

Brings `aspect gazelle` up to feature parity with `aspect format`, plus two gazelle-specific knobs that generalize the patterns we saw in the wild.

### Parity with `aspect format`

| Capability | `aspect format` | `aspect gazelle` before | after |
|---|---|---|---|
| Multi-value positional | `files` accepts up to 4096 | `dirs` accepted **1** (declaration bug) | up to 4096 |
| `--scope=all\|changed` | ✓ default `changed` | ✗ | ✓ default `all` |
| `--base-ref` / `--merge-base` | ✓ | ✗ | ✓ |
| `--soft-fail` | ✓ | ✗ | ✓ |
| `--check=auto\|true\|false` | ✓ | partial (2-state boolean) | ✓ |

**`--scope=all` default** (vs format's `changed`): the typical gazelle usage is whole repo. On larger repos users may set "changed" to optimize performance.

### Multi-dir positional bug

`args.positional(...)` defaults `maximum` to **1** in the AXL runtime. Every other task that takes a list of positionals overrides it (build/test/lint → 512, format → 4096, delivery → 1B); `gazelle.axl` didn't. So `aspect gazelle dir1 dir2` failed CLI parsing — even though the downstream body (line 489: `diff_args.extend(ctx.args.dirs)`) was already list-correct.

### New gazelle-specific flags

- **`--scope-all-on-change`** (defaults `["BUILD.bazel", "MODULE.bazel"]`): when `--scope=changed` and a changed file matches one of these globs, escalate to a whole-repo update.
- **`--changed-file-pattern`**: include-style glob filter on the changed files that contribute dirs. Empty default.

### Adjacent robustness in `--scope=changed`

- Non-existent dirs (e.g. from a deletion-only change) are warned and skipped, not passed to gazelle (which would error).
- Overlapping subtrees (`services` + `services/api`) collapse to the root.

---

- Searched for relevant documentation and updated as needed: no (in-code only)

**BREAKING CHANGE**

In the built-in `gazelle` task, the `--check` changed from boolean to 3-state. `aspect gazelle --check` (bare flag, previously meaning `--check=true`) no longer parses; users now write `--check=true` explicitly or omit it (`--check=auto` will enforce on CI). Also: `--check=false` (apply mode) previously exited 1 on diff; now exits 0 ("we just did the work — review and commit"). `aspect gazelle` was added recently in #1045 before public stability commitment, so we're revising its surface.

#### Suggested release notes

- `aspect gazelle` now accepts multiple directory positionals (`aspect gazelle dir1 dir2 dir3`). Previously rejected as a CLI parse error.
- New `--scope=changed` mode restricts updates to directories that contain files changed against the PR's base branch. Prefers the GitHub PR Files API when authenticated and in a PR context (works on any CI host), falling back to local `git diff <merge-base>` otherwise — `--base-ref` (default `origin/main`) and `--merge-base` drive the fallback. Combine with positional dirs to intersect — `aspect gazelle --scope=changed apps/web/` only updates changed subtrees under `apps/web/`. Default stays at `--scope=all`.
- New `--soft-fail` exits 0 with a WARNING instead of failing the task when BUILD files are out of date — useful for incremental adoption.
- New `--scope-all-on-change` (defaults `BUILD.bazel` and `MODULE.bazel`) escalates `--scope=changed` to a whole-repo update when key config files change.
- New `--changed-file-pattern` filters which changed files contribute to the derived dir set — set to mirror your gazelle language plugins.
- **Breaking**: `--check` is now a 3-state string (`auto` (default) / `true` / `false`) instead of a boolean. Bare `--check` no longer parses; use `--check=true` (or omit it on CI — `auto` enforces there). `--check=false` (apply mode) now exits 0 on diff instead of 1.
- **Caution on `--scope=changed`**: this and its dependent flags (`--dirs`, `--scope-all-on-change`, `--changed-file-pattern`) are an incremental-CI optimization that *narrows the work* gazelle does — a misconfiguration can silently MISS BUILD-file updates a full `--scope=all` run would have produced. Concretely: narrow `--dirs` drops changed dirs the dirs listed; `--changed-file-pattern` must list every source-file extension your gazelle plugins care about; `--scope-all-on-change` must list every gazelle-config-style file whose changes affect output repo-wide (`gazelle_python.yaml`, `tsconfig.json`, `package.json`, top-level `BUILD.bazel`, …). Default to `--scope=all`; reach for `--scope=changed` only when slow gazelle performance justifies the configuration burden, and periodically re-run with `--scope=all` to check for drift. The per-flag help text and the `gazelle.axl` file docstring enumerate each failure mode.
- **Performance note on `--scope=changed`**: under gazelle's default `-index=all`, passing positional dirs (which is how `--scope=changed` invokes gazelle under the hood) only skips the BUILD-regeneration step for non-listed dirs. Gazelle still traverses the workspace, parses every BUILD, and re-determines each existing target's exports (per language; often re-parsing source files in `srcs`) repo-wide before regeneration runs. On large repos those steps dominate wall time, so the `--scope=changed` speedup against a stock gazelle invocation is much smaller than it looks. For the full incremental win, pair with gazelle's `-r=false -index=lazy` and add `# gazelle:<lang>_search` directives in BUILD files for every external code root your project imports from (third_party trees, Go module replacements, vendored proto, …): `aspect gazelle --check=true --scope=changed --gazelle-flag=-r=false --gazelle-flag=-index=lazy`. Caveat: `-index=lazy` requires every active gazelle language extension to implement the lazy-indexing API; extensions that pre-date the API will under-index and may resolve imports incorrectly. This is why `-index=all` is still gazelle's default — verify your active plugins support it before flipping this on in CI. See [the gazelle docs](https://github.com/bazel-contrib/bazel-gazelle#lazy-indexing-in-fix-and-update) for details.

### Test plan

- Covered by existing test cases: 729 existing AXL cases still pass.
- New test cases added: 32 in `test_gazelle_dir_helpers` covering `extract_changed_dirs`, `dir_at_or_below`, and `dedupe_subtrees` (including `./` / `.` normalization and `_make_scope_changed*` snapshot fixtures); total AXL count now 761.
- All snapshot suites green: PR-comment, template, gazelle-template (15 scenarios, +3 new for `--scope=changed`), format-template, lint-template, BK annotation, delivery-template.
- Manual: build with `cargo build -p aspect-cli`, run `./target/debug/aspect-cli tests axl` and the gazelle / pr-comment snapshot suites.


crates/aspect-cli/src/builtins/aspect/gazelle.axl
```diff
@@ -606,17 +770,43 @@ gazelle = task(
...
+        ),
+        "scope": args.string(
+            default = "all",
+            values = ["all", "changed"],
+            description = "Which directories to gazelle. 'all' (default) lets gazelle walk every directory (subject to --dirs scoping when supplied). 'changed' restricts the walk to directories that contain files changed against the base branch — the GitHub PR Files API in PR contexts, otherwise `git diff <merge-base>`. Useful for fast incremental gazelle in CI on large monorepos.",
```
jbedard: FWIW gazelle will most likely still walk everything, it just only updates the directories passed here, maybe if the languages enabled support "lazy indexing" then it will only lazy-walk and be more efficient.

This is most likely too much info and not worth describing such (incorrect) details.

crates/aspect-cli/src/builtins/aspect/gazelle.axl
```diff
@@ -628,7 +818,8 @@ gazelle = task(
         ),
         "dirs": args.positional(
             minimum = 0,
-            description = "Directories to pass to gazelle as positional arguments. Gazelle limits its walk to these subtrees. When empty (default), gazelle walks the entire repo from the workspace root. Useful for limiting scope on large repos or when only certain packages need updating.",
+            maximum = 4096,
+            description = "Directories to limit gazelle's walk to (subtrees). When empty (default), gazelle walks the entire repo from the workspace root. Combine with --scope=changed to take the intersection with directories that contain changed files. Useful for scoping to a workspace within a monorepo and for limiting work on large repos.",
```
jbedard: Same here, by default gazelle walks everything but only updates the `dirs`. This is too much (incorrect) details about gazelle...

crates/aspect-cli/src/builtins/aspect/gazelle.axl
```diff
@@ -221,6 +239,128 @@ def _body_text(status, data):
...
+        overlapping subtrees."""
+    user_dirs = []
+    for d in (ctx.args.dirs or []):
+        if ctx.std.fs.exists(d):
+            user_dirs.append(d)
+        else:
+            warn(ctx.std, "Skipping non-existent directory: %s" % d)
+
+    if ctx.args.scope == "all":
+        return dedupe_subtrees(user_dirs), False
```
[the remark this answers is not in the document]
gregmagolan: Resolved

crates/aspect-cli/src/builtins/aspect/lib/gazelle_results.axl
```diff
@@ -53,6 +60,64 @@ def init_data():
...
+def dir_at_or_below(path, scopes):
+    """True if `path` equals any entry in `scopes` or lives under one
+    of them. Used by `--scope=changed × --dirs` intersection: a changed
+    subtree qualifies if it's at-or-below the user's interest scope."""
+    for s in scopes:
+        s_norm = s.rstrip("/") if s else s
+        if path == s_norm:
+            return True
+        if s_norm and path.startswith(s_norm + "/"):
+            return True
```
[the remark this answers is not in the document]
gregmagolan: Resolved

crates/aspect-cli/src/builtins/aspect/gazelle.axl
gregmagolan: Resolved

crates/aspect-cli/src/builtins/aspect/gazelle.axl
gregmagolan: Resolved

crates/aspect-cli/src/builtins/aspect/lib/gazelle_results.axl
```diff
@@ -191,7 +325,11 @@ def _build_details_data(data, status):
                      not data["gazelle"].get("gazelle_error") and
                      not data["gazelle"].get("apply_error"),
         "has_overflow": overflow > 0,
-        "check_mode": bool(data["gazelle"].get("check_mode")),
+        # In the template, `check_mode` means "did we run in CI-style
+        # gate mode (skipped the apply step)?" — keyed off the resolved
+        # `enforce_check` bool, not the raw arg string (which is always
+        # truthy under Jinja2's empty-string-only-falsy rule).
```
jbedard: WTF is Jinja2 and how is it relevant to this `check_mode`?

Is this entire comment block just AI junk that can be deleted?

gregmagolan: This file (gazelle_results.bzl) is specific to the status screen that are driven by Jinja templates so this block is specific to the display layer (GitHub Status Checks & Buildkite Annotations) as an overlay on top of the configuration and results that come out of the test. Tho the Jinja template reference specifically is unnecessary, the context that check_mode display state is driven off of enforce_check is valid since chech_mode may be "auto" but we should display what auto resolved to in the UI.

gregmagolan: I'll drop the Jinja reference tho

gregmagolan: Tho Jinja references are generally valid in this file since there is a Jinja template that all of the values data values are fed into:
```
_SUMMARY_TEMPLATE = """{% if detail_rows %}{% for row in detail_rows %}
{% for item in row %}{{ item.icon }} {{ item.value }}{% if not loop.last %} · {% endif %}{% endfor %}{% if not loop.last %}  {% endif %}
{%- endfor %}{% endif %}
{% if config_items %}
:gear: {% for item in config_items %}**{{ item.key }}:** `{{ item.value }}`{% if not loop.last %} · {% endif %}{% endfor %}{% if last_update %}  {% endif %}{% endif %}{% if last_update %}
:speech_balloon: **Last update:** `{{ last_update }}`{% endif %}"""

_DETAILS_TEMPLATE = """{% if gazelle_error %}
> ❌ Gazelle exited with code {{ gazelle_error_code }} without producing a diff. This usually indicates a configuration error — check the task log.

{% elif apply_error %}
> ❌ `git apply -p0` failed after generating the diff. This usually means the working tree changed during the run.
{% if apply_error_msg %}
> Error: {{ apply_error_msg }}
{% endif %}
...
```

PR #1084, opened 2026-05-15
feat(gazelle): default-on .gitignore-skipping plus opt-in untracked-file skipping

gregmagolan:
## Summary

Two new boolean flags on `aspect gazelle` that close gaps in how gazelle interacts with git state, plus a couple of adjacent fixes to the status surface.

### `--ignore-gitignored-files` (default **true**)

Gazelle's native ignore mechanisms (`.bazelignore` literal paths, `REPO.bazel ignore_directories([...])` dir globs, `# gazelle:exclude` per-package directives) don't honor `.gitignore`. Without this flag every gitignored dir users want gazelle to skip — `bazel-out/`, `node_modules/`, `.venv/`, `__pycache__/`, build artifacts — has to be mirrored elsewhere. Defaults **on** to close that gap automatically: queries `git ls-files --others --ignored --exclude-standard --directory -z`, dedups results against `.bazelignore` + `REPO.bazel` + directory symlinks gazelle doesn't follow anyway, and passes each surviving entry as `--gazelle-flag=-exclude=<path>`. Pass `--ignore-gitignored-files=false` to opt out (rare — e.g., a gitignored file that gazelle's indexing needs to see).

### `--ignore-untracked-files` (default **false**)

Symmetric mechanism for untracked files. When set, queries `git ls-files --others --exclude-standard --directory -z` and passes each result as `--gazelle-flag=-exclude=<path>` (same dedup pipeline). Off by default because untracked files are typically work-in-progress the user wants gazelle to process; turn it on in pre-commit / pre-push hooks where you want only committed-state rules.

### Adjacent fixes / cleanups

- **Status surface no longer renders "❌ 0 BUILD files out of date" with an empty file list during live runs** — gazelle's details template was missing the `is_running` short-circuit that format already has. Now shows `🔄 Gazelle task in progress...` during the bazel build / diff phase.
- **Configuration row keys** are sentence case across both `gazelle_results` and `format_results`: `Soft fail`, `Scope all on change`, `Changed file pattern` (previously kebab-case).
- **Each `Ignore-*` row surfaces only when its value departs from default** so a stock invocation has a clean Config row.
- trace.log the exact subprocess invocation (gazelle binary + every resolved flag / exclude / dir; same for format).

---

#### Suggested release notes

- **New `aspect gazelle --ignore-gitignored-files` flag, defaulting on.** Gazelle doesn't honor `.gitignore` natively. With this default-on, the task queries `git ls-files --others --ignored --exclude-standard --directory` on each invocation and passes each result as `--gazelle-flag=-exclude=<path>`, so `bazel-out/`, `node_modules/`, `.venv/`, `__pycache__/`, build artifacts no longer get walked. Pass `--ignore-gitignored-files=false` to opt out. No-op on clean CI checkouts; deduped against `.bazelignore` + `REPO.bazel ignore_directories()` so we don't restate what gazelle already skips.
- New `aspect gazelle --ignore-untracked-files` flag, defaulting off. Symmetric mechanism for untracked files. Useful in pre-commit / pre-push hooks where you don't want BUILD rules generated for files you haven't checked in yet.
- Both flags fall back to a no-op with a `WARNING` if `git ls-files` fails.
- Gazelle's status-surface details body no longer reads "❌ 0 BUILD files out of date" with an empty file list during the bazel build / diff phase — live renders show a `🔄 Gazelle task in progress...` placeholder.
- Status-check Configuration row keys are sentence case (`Soft fail`, `Scope all on change`, `Changed file pattern`, …) — previously kebab-case.
- With `ASPECT_DEBUG=1`, `aspect gazelle` and `aspect format` log a `trace` line with the exact subprocess command they're about to spawn (resolved flags / excludes / dirs / file list).

### Test plan

- Existing AXL cases pass: 779 total (763 before this PR + 16 new for `parse_bazelignore` and `parse_repo_bazel_ignore_directories`).
- Snapshot suites green: gazelle-template (17 scenarios, including the new `_make_ignore_untracked` and `_make_ignore_gitignored_disabled` Configuration-row coverage), PR-comment (17), BK annotation (21), plus the standard template / format / lint / delivery suites.
- Manual: in a git working tree with at least one untracked file (and a `.gitignore`-matched file on disk), run `aspect gazelle` and confirm the rendered Configuration row matches expectations and that the affected dirs don't get BUILD rules updated.

### DIFF
--- a/.aspect/axl.axl
+++ b/.aspect/axl.axl
@@ -20,6 +20,8 @@ load(
     "dedupe_subtrees",
     "dir_at_or_below",
     "extract_changed_dirs",
+    "parse_bazelignore",
+    "parse_repo_bazel_ignore_directories",
 )
 load("@aspect//lib/github.axl", "detect_build_url", "detect_commit_sha", "github")
 load("@aspect//lib/path_glob.axl", "match_any", "match_path")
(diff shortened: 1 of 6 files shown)


jbedard (changes requested):
This conflicts with the aspect-gazelle gitignore extension (`# gazelle:gitignore enabled`) and the PR summary is quite inaccurate AI slop.

Please do not merge such a thing into the CLI or AXL.

PR #1110, opened 2026-05-21
fix: correctly determine deliverable targets

thesayyn:

### Test plan


crates/aspect-cli/src/builtins/aspect/delivery.axl
```diff
@@ -422,27 +440,129 @@ def _get_output_shas(ctx, bazel_trait, lifecycle, build_events, targets, bazel_f
...
+    _unresolved = len(targets) - len(output_shas)
+    if _unresolved == 0:
+        print("  Resolved digests for all {} in {}.".format(targets_label, fmt_elapsed(time.monotonic() - _t_phase2)))
+    else:
+        # Always point at the phase-2 log when anything went unresolved —
+        # the gRPC log may contain zero useful entries (auth/network/config
+        # error before any GetActionResult fired), in which case the
+        # SUSPECT branch below never executes and the log file is the only
+        # diagnostic surface left.
+        print("  Resolved {} of {} delivery digests in {}; {} could not be hashed. Phase-2 raw log: {}".format(
```
gregmagolan, line 424: 👍

crates/aspect-cli/src/builtins/aspect/delivery.axl
```diff
@@ -422,27 +440,129 @@ def _get_output_shas(ctx, bazel_trait, lifecycle, build_events, targets, bazel_f
...
+            fmt_elapsed(time.monotonic() - _t_phase2),
+            _unresolved,
+            phase2_log_path,
+        ))
+
+    # Surface non-DeliveryHash cache misses. Lazy-parse the phase-1
+    # execution log (captured to disk above but skipped on the clean path)
+    # to build a per-`(target_label, mnemonic)` map of `remote_cacheable`,
+    # then split:
+    #   * INTENTIONAL — phase 1 recorded the action with
```
gregmagolan: If I'm reading this correctly, if a rule opt-out of remote cache, it won't work with delivery?

crates/aspect-cli/src/builtins/aspect/delivery.axl
```diff
@@ -422,27 +440,129 @@ def _get_output_shas(ctx, bazel_trait, lifecycle, build_events, targets, bazel_f
...
+            print("    tags/execution_requirements. Either transient (not yet uploaded; will")
+            print("    resolve on a repeat delivery) or non-hermetic (action key changes every")
+            print("    build; persistent until the rule is fixed). The affected delivery")
+            print("    targets may show 'no digest' in their per-target row.")
+            for producer, mnemonics in suspect.items():
+                print("      " + producer + ": " + ", ".join(mnemonics))
+            print("    Phase-2 raw log: " + phase2_log_path)
+        if intentional:
+            print("")
+            print("  Note: actions opted out of the remote cache (expected; not a delivery problem):")
```
gregmagolan: Can we expand on "not a delivery problem" to be more helpful for users? Adding more context that delivery actions must be remote cacheable to work with selective-delivery would be helpful if true.

crates/aspect-cli/src/builtins/aspect/delivery.axl
```diff
@@ -1037,17 +1168,24 @@ def _delivery_impl(ctx):
             # Phase 3 ran (any mode/flag combination except the dry-run +
             # --track-state=false preview) — run_tracker has entrypoints.
             entrypoint = run_tracker.determine_entrypoint(label)
-            if not entrypoint:
+            if entrypoint == None:
+                # Bazel never reported an output for this target in the build
+                # event stream — usually an alias, a transition redirecting
+                # the build, or a target filtered out by configuration.
+                warn_msg = "no build output reported for this target (likely an alias, filtered, or transitioned away)"
```
gregmagolan: "likely" => "possibly"

gregmagolan: we don't know what we don't know

gregmagolan (approved):

crates/aspect-cli/src/builtins/aspect/delivery.axl
thesayyn: ah this is a left over, will remove. we don't care if the target is in the remote cache, as long as its in the action cache phase 2 won't fail.

PR #1131, opened 2026-05-25
fix(bazelrc): make undefined-config error self-explanatory

gregmagolan:
### Before

```
→ ⚙ Setup · Authenticating and creating CI status checks
INFO: GitHub API throttle — Aspect Workflows GitHub App: factor=13× (quota=9% used, elapsed=0s)
error: Traceback (most recent call last):
  File <builtin>, in <module>
  * /home/runner/.cache/aspect/axl/deps/.../aspect/test.axl:13, in _impl
      return run_bazel_task(ctx, command = "test")
  * /home/runner/.cache/aspect/axl/deps/.../aspect/lib/bazel_runner.axl:262, in run_bazel_task
      rc_startup_flags, flags = rc.expand_all(command = command)
error: undefined config 'ci' for command 'test'
   --> /home/runner/.cache/aspect/axl/deps/.../aspect/lib/bazel_runner.axl:262:31
    |
262 |     rc_startup_flags, flags = rc.expand_all(command = command)
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### After

(rendered with ANSI colors on a TTY/CI; bold sections + red `bazelrc:` not shown in the markdown below)

```
→ 🎬 Running `test` task
error: bazelrc: --config=ci is not defined for command 'test'

  Aspect CLI parses .bazelrc files and resolves --config= flags like Bazel.

  Bazel workspace root:
    /Users/greg/aspect/aspect-cli

  Loaded .bazelrc files:
    /Users/greg/aspect/aspect-cli/.bazelrc
    /Users/greg/aspect/aspect-cli/bazel/defaults.bazelrc
    /Users/greg/.bazelrc

  Applicable rc state for 'test' (same as --announce-rc):
    ./.bazelrc
      common  --repo_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1 ...

    ./bazel/defaults.bazelrc
      common    --nobuild_runfile_links --enable_platform_specific_config ...
      common:ci --announce_rc --color=yes --curses=yes ...

    /Users/greg/.bazelrc
      common:ext-cache  --remote_accept_cached --remote_upload_local_results
      ...

    client  --isatty=1 --config=ci --build_metadata=ASPECT_TASK_NAME=Test ...

  Try one of:
    - Add `common:ci ...` (or `build:ci`, `test:ci`) to a loaded .bazelrc
    - Don't set --config=ci
```

Bazel itself has hard-errored on undefined `--config=` since 0.16.0; [bazel#7116](https://github.com/bazelbuild/bazel/issues/7116) (request for a tolerant mode) is closed won't-fix.

---

The `undefined config 'ci' for command 'test'` error rendered with no further context, fired late in the task lifecycle (after `task_started` had already created CI status checks and authenticated with GitHub), and the wrapping AXL traceback made it read like an internal Aspect crash rather than a bazelrc resolution issue users could fix themselves.

Five coordinated fixes:

**1. Validate bazelrc before lifecycle hooks.** Flag aggregation, `parse_rc`, and `expand_all` now run before `setup_phase` / `task_started` in [bazel_runner.axl](crates/aspect-cli/src/builtins/aspect/lib/bazel_runner.axl). An rc-resolution error fails with no CI side effects — no status checks created, no API calls burned.

**2. Enrich the error message.** The block leads with a `bazelrc:` prefix, surfaces the Bazel workspace root (so sub-workspace anchor mismatches are visible — parent `config.axl` sets `--config=ci` while the inner `.bazelrc` doesn't define `:ci`), lists the loaded rc files, and inlines the `--announce-rc` view for the failing command. For rc-file references it also names the file that owns the unresolved `--config=X` line.

**3. ANSI colors + bold headers.** Matching the SGR conventions used in `axl-runtime/src/eval/multi_phase.rs`: bold red for the `bazelrc:` prefix, bold for section headers, dim for the subsystem-intro line. Color detection lives in the new [`axl-runtime::term`](crates/axl-runtime/src/term.rs) module — single source of truth for the AXL runtime, the task lifecycle headers, and the bazelrc error block. Honors the [NO_COLOR](https://no-color.org/) convention (disable iff present and non-empty); AXL-side `lib/environment.axl::color_enabled` does the same.

**4. Strip the AXL traceback.** The error is propagated through anyhow without losing its `BazelRcError` type; `EvalError`'s Display renders pre-shaped user errors via `starlark::Error::without_diagnostic()` and `main.rs` switches from `{err:?}` to `{err}` when `EvalError::is_pre_shaped_user_error()` returns true. Non-bazelrc errors keep the full Starlark diagnostic so AXL/Aspect authors can debug.

**5. Workspace-relative paths in `announce()`.** In-tree rc files render as `./bazel/defaults.bazelrc`; external files (e.g. `~/.bazelrc`) keep their absolute path.

---

- Searched for relevant documentation and updated as needed: no — error text only
- Breaking change: no

#### Suggested release notes

- The `undefined config '<name>' for command '<cmd>'` error from `aspect build` / `aspect test` now fails before any lifecycle hooks fire (no stuck CI status check, no wasted API calls) and expands into a Bazel-style colored explanation: the Bazel workspace root, the loaded `.bazelrc` files with their workspace-relative or absolute paths, an inline `--announce-rc` view, and suggested fixes. The AXL traceback is stripped for these errors. ANSI colors honor the [NO_COLOR](https://no-color.org/) convention. Sub-workspace anchor mismatches (parent `config.axl` setting `--config=ci` while the inner workspace's `.bazelrc` doesn't define `:ci`) are now diagnosable at a glance.

### Test plan

- Eight unit tests in [crates/bazelrc/src/lib.rs](crates/bazelrc/src/lib.rs) cover the rendered output: a golden-text assertion of the full CLI-source block (including the inline `--announce-rc` dump and `./` workspace-relative paths), rc-file source attribution via the `Unresolved --config=X reference in:` block, transitive expansion source attribution, the `(none)` rendering under `--ignore_all_rc_files`, a regression guard against internal jargon (`injection`, `BazelTrait`, `extra_flags`), ANSI escape embedding under `with_ansi_errors(true)`, `loaded_rc_files()` filtering out the synthetic CLI source, and `announce()` rendering in-tree vs out-of-tree paths.
- Four unit tests in [crates/axl-runtime/src/term.rs](crates/axl-runtime/src/term.rs) cover the color-detection helper: `NO_COLOR` non-empty disables even on CI; `NO_COLOR=""` (empty) is treated as unset per the spec; no-CI-and-no-TTY means no color; each recognized CI env var triggers detection. Uses an RAII `EnvGuard` to snapshot/clear/restore process env around each case, serialized via a Mutex against cargo's parallel runner.
- Manual: reproduced in a tempdir with a parent `.aspect/config.axl` setting `--config=ci` and a sub-workspace with `MODULE.bazel` but no `.bazelrc`. Confirmed only `→ 🎬 Running test task` prints before the error, the bazelrc block renders in color on a TTY (and plain under `NO_COLOR=1`), and no AXL traceback wraps the message.

### Follow-ups (separate PRs)

- Auto-derive `skip_config_if_missing` from `BazelTrait.extra_flags` so framework-set configs become "soft" by design — the right primary fix for the sub-workspace case. This PR improves the error for cases where strict behavior remains correct.

### DIFF
--- a/crates/axl-runtime/src/eval/multi_phase.rs
+++ b/crates/axl-runtime/src/eval/multi_phase.rs
@@ -1,4 +1,3 @@
-use std::io::IsTerminal;
 use std::path::{Path, PathBuf};
 
 use anyhow::anyhow;
@@ -45,25 +44,13 @@ const SGR_RESET: &str = "\x1b[0m";
 const DEFAULT_HIGHLIGHT_COLOR: &str = "1;36";
 const HIGHLIGHT_COLOR_ENV: &str = "ASPECT_CLI_HIGHLIGHT_COLOR";
 
-/// True on any recognized CI host (Buildkite / GitHub Actions / CircleCI /
-/// GitLab CI / generic `CI=...`). Used to gate task-key suffix on the
-/// running-task header and to force-enable color when stderr is piped to
-/// a CI log viewer that renders ANSI.
-fn on_recognized_ci() -> bool {
-    std::env::var_os("BUILDKITE").is_some()
-        || std::env::var_os("CI").is_some()
-        || std::env::var_os("GITHUB_ACTIONS").is_some()
-        || std::env::var_os("CIRCLECI").is_some()
-        || std::env::var_os("GITLAB_CI").is_some()
-}
-
 /// Return `(verb_color_prefix, reset)` for the "Running" verb on the
 /// task-starting line. `verb_color_prefix` is empty when color is
 /// disabled or `ASPECT_CLI_HIGHLIGHT_COLOR=""`; `reset` is empty when
-/// color is disabled. Color is enabled when stderr is a TTY or we're on
-/// a recognized CI host.
+/// color is disabled. The color decision is delegated to
+/// [`crate::term::color_enabled`].
 fn running_verb_color() -> (String, &'static str) {
-    let color = std::io::stderr().is_terminal() || on_recognized_ci();
+    let color = crate::term::color_enabled();
     let highlight = std::env::var(HIGHLIGHT_COLOR_ENV)
         .ok()
         .unwrap_or_else(|| DEFAULT_HIGHLIGHT_COLOR.to_string());
@@ -374,7 +361,7 @@ impl<'v, 'l> MultiPhaseEval<'v, 'l> {
             EvalError::UnknownError(anyhow!("task index {} out of range", task_index))
         })?;
         let (verb_seq, reset) = running_verb_color();
-        let key_suffix = if on_recognized_ci() {
+        let key_suffix = if crate::term::on_recognized_ci() {
             format!(" [{}]", task_key)
         } else {
             String::new()
@@ -501,7 +488,7 @@ impl<'v, 'l> MultiPhaseEval<'v, 'l> {
         // the task body are still announced from AXL via
         // `lib/lifecycle.axl::announce`. Color setup mirrors the helper
         // so the verdict glyph and labels match.
-        let color = std::io::stderr().is_terminal() || on_recognized_ci();
+        let color = crate::term::color_enabled();
         let sgr = |params: &str| {
             if color {
                 format!("\x1b[{}m", params)
--- a/.aspect/config.axl
+++ b/.aspect/config.axl
@@ -48,6 +48,13 @@ def config(ctx: ConfigContext):
     # assert we can call a lambda with a global bind from another module in a config script
     lambda_with_global_bind()("assert")
 
+    is_ci = bool(ctx.std.env.var("CI"))
+
+    if is_ci:
+        ctx.traits[BazelTrait].extra_flags.extend([
+            "--config=ci",
+        ])
+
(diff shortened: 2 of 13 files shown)


crates/axl-runtime/src/eval/multi_phase.rs
```diff
@@ -501,7 +488,7 @@ impl<'v, 'l> MultiPhaseEval<'v, 'l> {
         // the task body are still announced from AXL via
         // `lib/lifecycle.axl::announce`. Color setup mirrors the helper
         // so the verdict glyph and labels match.
-        let color = std::io::stderr().is_terminal() || on_recognized_ci();
+        let color = crate::term::color_enabled();
```
thesayyn, line 491: its sort of sad that CI specific code is creeping into the rust code. ideally,  this should be knob on the parse_bazelrc call.

PR #1147, opened 2026-05-27
Add root BUILD aliases for buildifier, format, and gazelle tasks

gregmagolan:
Add convenience `//:buildifier`, `//:format`, and `//:gazelle` aliases in the root BUILD.bazel so developers can run `bazel run //:buildifier`, `bazel run //:format`, and `bazel run //:gazelle` directly.

---

### Test plan

- Covered by existing CI

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -26,3 +26,18 @@ exports_files(
     ],
     visibility = ["//visibility:public"],
 )
+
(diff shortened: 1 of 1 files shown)


gregmagolan:
On 2nd thought, lets not use this pattern in this repo

PR #1156, opened 2026-05-28
fix(delivery): run targets without a runfiles manifest

JesseTatasciore:
Custom rules that emit a `.sh` script directly (without `sh_binary`) produce no `.runfiles_manifest` in BES output — they don't use Bazel's runfiles protocol and don't need one. The guard added in #1142 treated a missing manifest as proof that a target is not runnable and blocked delivery with a warning, silently skipping all such targets.

`_spawn` already handles `ep.runfiles_dir = None` correctly: it skips the runfiles env vars and falls back to `current_dir()` as cwd. The guard was both incorrect (a missing manifest is valid for custom rules) and redundant. Removing it restores the previous behaviour.

Adds a `gen_sh_binary` custom rule to `examples/deliverable` to cover this shape in the delivery CI task and prevent future regressions.

---

**Suggested release notes:**
- `aspect delivery` now correctly runs targets from custom rules that emit a `.sh` script without a runfiles tree. Previously these targets were silently skipped with a "not runnable" warning.

### Test plan

- `gen_sh_binary` target added to `examples/deliverable` — exercised by the existing delivery CI task on every run

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/delivery.axl
+++ b/crates/aspect-cli/src/builtins/aspect/delivery.axl
@@ -1009,11 +1009,6 @@ def _delivery_impl(ctx):
                 pending_count += 1
                 data.append((label, "WARN", "warn", warn_msg, output_sha or "-", "-"))
                 delivery_add_result(dr, label, "warn", message = warn_msg, output_sha = output_sha or "", is_forced = is_forced)
-            elif not ep.runfiles_dir:
-                warn_msg = "tagged deliverable but not runnable (no .runfiles_manifest in BES output for {})".format(ep.path)
-                pending_count += 1
-                data.append((label, "WARN", "warn", warn_msg, output_sha or "-", "-"))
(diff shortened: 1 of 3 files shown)


crates/aspect-cli/src/builtins/aspect/delivery.axl
```diff
@@ -1009,11 +1009,6 @@ def _delivery_impl(ctx):
                 pending_count += 1
                 data.append((label, "WARN", "warn", warn_msg, output_sha or "-", "-"))
                 delivery_add_result(dr, label, "warn", message = warn_msg, output_sha = output_sha or "", is_forced = is_forced)
-            elif not ep.runfiles_dir:
```
thesayyn, line 1012: This is required to actually make sure that the targets passed to delivery is actually deliverable (runnable).

gregmagolan:
Closing in favor of https://github.com/aspect-build/aspect-cli/pull/1157

PR #1180, opened 2026-06-02
Error on undeclared task/feature arg in config.axl

gregmagolan:
Assigning an undeclared arg on a task's or feature's `.args` override store in `config.axl` silently succeeded, writing a junk key the runtime never reads. A typo like `ctx.tasks["delivery"].args.warm_not_runnable = True` (for `warn_not_runnable`) produced no error and no effect — the misspelled override was simply dropped.

`Arguments::set_attr` blindly inserted any attribute name into its map, with no validation against the task's/feature's declared args. (For contrast, reading an unknown arg already errors via `get_attr` returning `None`.)

This gives `Arguments` an optional `valid_keys` schema. The config-time override stores (`ctx.tasks[k].args`, `ctx.features[X].args`) are now constructed with `Arguments::with_schema(...)` from the declared arg names, so `set_attr` rejects an unknown name with `no such arg \`<name>\`; valid args are: ...`. The runtime-args store (`ctx.args`, built by merging CLI + config + defaults) stays permissive — its keys are already known-good and it inserts resolved values directly.

Wired into all five override-store creation sites: `Task::from_frozen`, the `task.alias(...)` builder, the `task(...)` global, `Feature::from_frozen`, and the `feature(...)` global.

---

### Suggested release notes

- Assigning an undeclared argument on `ctx.tasks[...].args` / `ctx.features[...].args` in `config.axl` now raises a script error naming the valid args, instead of silently doing nothing. This surfaces typos (e.g. `warm_not_runnable` for `warn_not_runnable`) that previously went unnoticed. A config that set a bogus arg name will now fail until the name is corrected or removed.

### Test plan

- New test cases added
- Unit tests in `arguments.rs` cover: permissive store accepts any name; schema store accepts a declared arg; schema store rejects a typo, names the bad key and lists valid args, and does not write the junk key.

### DIFF
--- a/crates/axl-runtime/src/engine/arguments.rs
+++ b/crates/axl-runtime/src/engine/arguments.rs
@@ -37,6 +37,17 @@ use starlark::values::starlark_value;
 /// alias-overridden defaults and config.axl overrides). Empty for the
 /// config-time override store role — nothing is "explicit on the CLI"
 /// there.
+///
+/// `valid_keys` constrains which attribute names `set_attr` will accept:
+///
+/// - `Some(set)` — config-time override store. Assigning `.args.<name> = ...`
+///   in config.axl is rejected unless `<name>` is one of the task's / feature's
+///   declared args. This turns a typo (`warm_not_runnable` for
+///   `warn_not_runnable`) into a script error instead of a silent no-op that
+///   writes a junk key the runtime never reads.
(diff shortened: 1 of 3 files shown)


gregmagolan:
Closing for https://github.com/aspect-build/aspect-cli/pull/1181

PR #1186, opened 2026-06-02
feat(gitlab): GitlabCommitStatuses + GitlabStatusComments features (ENG-1651)

jeffpignataro:
Two AXL features for GitLab MR code-review signal, mirroring the GitHub path.

## Summary

- **`GitlabCommitStatuses`** — per-task commit status on the MR SHA via the GitLab Commit Status API. State transitions `pending → running → success/failed/skipped/canceled` mirror the per-task lifecycle that `GithubStatusChecks` drives on GitHub.
- **`GitlabStatusComments`** — single rolling MR note aggregating every task's status. Marker-discovered (`<!-- aspect-ci-status:mr-N -->`), create-or-update via note id. Cross-task coordination uses the note's own state block (base64-JSON in footer) as the shared source of truth — no GitLab CI artifact API dependency needed.

Both features gate on `detect_gitlab_mr()` returning an MR context; skip silently outside MR pipelines. Authentication goes through the Aspect GitLab App via `gitlab.authenticate(...)`.

## Why commit_status (not External Status Checks)?

The ticket originally specced ESC. On implementation review, `commit_status` is the closer semantic parallel to GitHub's `check_runs` / `statuses` API and the better real-world fit:

- Free tier (works on every GL plan) vs Premium/Ultimate-only ESC
- Free-form per-task `name` at POST time vs admin-pre-registered ESC names
- 6 states matching task lifecycle (pending/running/success/failed/canceled/skipped) vs ESC's 3 (passed/failed/pending)
- Informational signal vs ESC's merge-blocking approval gate

ESC plumbing (`gitlab.external_status_checks.respond/list` in `lib/gitlab.axl`, the `external_status_checks.write` FE role + permission) is already in place and works unchanged for a future feature that wants merge-blocking semantics. Updated Linear ticket reflects this pivot.

## Token cache

GitLab OAuth refresh-grant rotation revokes the prior access token on each new mint, so two features in the same aspect-cli process calling `gitlab.authenticate(...)` independently end up with mutually revoking tokens — the second mint kills the first feature's token mid-build. `lib/gitlab_token_cache.axl` mints at most once per `(project_path, roles)` tuple per process and caches under tmpdir (same pattern as `lib/checkrun.axl`).

## Rendering

`feature/gitlab_status_comments.axl` reuses the VCS-agnostic helpers (`render_body`, `parse_state`, `merge_state`, `prepare_for_render`, `bucket_entries`, `results_subset`, `lease_held`, `max_recent_upsert_epoch`) directly from `feature/github_status_comments.axl` via the public re-exports the GH feature already declared for its own tests. Markup is identical to GH for equivalent inputs.

`feature/gitlab_commit_statuses.axl` delegates to the same per-kind renderer table (`check_dispatch.RENDERERS`) the GH check feature uses; only the rendered `title` lands on the commit status (in `description`, truncated to GL's 255-char cap) — the full body renders into the rolling MR note via `GitlabStatusComments`.

## Tests

`lib/gitlab_features_test.axl` covers the GL-specific bits:

- `state_for(status, final)` — Aspect task status → GL commit_status state mapping (warning→skipped, aborted→canceled, etc).
- `truncate_description` — handles GL's 255-char cap, newline collapse, ellipsis.
- `MARKER_FMT` — MR-keyed marker (`mr-N`, not `pr-N`) so projects mirrored to both GH and GL don't have rolling notes/comments collide.

Run: `aspect dev test-gitlab-features`.

The render path is reused verbatim from the GH feature; coverage already lives in `feature/github_status_comments_test.axl`.

## End-to-end verified

Tested against MR #4 of `aspect-build/aw-rut-gl-test-dev-gcp` in dev:

- Commit statuses landed on the SHA with proper state transitions (pending → running → success)
- Rolling MR note created and updated across multiple aspect-cli invocations
- Multiple features sharing the cached token without mutual revocation

## Follow-ups (filed in ENG-1652)

- Server-side `roles.ts` `mr_notes.write` declares `gitlabScope: null` which resolves to `read_api` only when narrowed — but the notes API requires `api` scope. We work around by requesting `roles=["api"]` instead of fine-grained. Fix is to declare `gitlabScope: 'api'` for `mr_notes.write` in `marvin-gitlab-app/roles.ts`.
- `lib/gitlab.axl` defaults `auth_kind="private_token"` on every API call. Marvin-minted tokens are always OAuth; defaults are inverted for the canonical use case. Flip in a separate cleanup PR (PAT users add `auth_kind="private_token"` to opt back).
- ESC feature for merge-blocking semantics if a customer asks.

Linear: ENG-1651


jeffpignataro:
Tested against an MR on `aw-rut-gl-test-dev-gcp`. Running well with build, test, and lint.

image

image

PR #1190, opened 2026-06-02
feat(cli): add tools/bazel reference wrapper

gregmagolan:
Customers who want Aspect CLI features but don't want their developers to learn `aspect` as a new command can drop this `tools/bazel` shell wrapper in their workspace. [Bazelisk](https://github.com/bazelbuild/bazelisk) execs `tools/bazel` when present (the wrapper hook is a Bazelisk feature — the real `bazel` binary doesn't look for it), so it routes each command to the right tool.

| You type | What runs |
| --- | --- |
| `bazel build //... --keep_going --config=ci` | `aspect build //... --bazel-flag=--keep_going --bazel-flag=--config=ci` |
| `bazel build -c opt //...` | `aspect build --bazel-flag=-c=opt //...` |
| `bazel test //... --test_output errors` | `aspect test //... --bazel-flag=--test_output=errors` |
| `bazel lint --config=ci //src/...` | `aspect lint --bazel-flag=--config=ci //src/...` |
| `bazel format --keep_going` | `aspect format --bazel-flag=--keep_going` |
| `bazel query 'deps(//foo)'` / `bazel info` | vanilla bazel, unchanged |
| `bazel my-custom-task //...` | `aspect my-custom-task //...` (custom `.axl` task) |

## Model

Two lists at the top of the script drive routing (edit them in your repo copy):

- **`ASPECT_VERBS_WITH_BAZEL_FLAGS`** → route to `aspect` with bazel-flag rewriting.
- **`BAZEL_VERBS`** → the closed set of Bazel commands.

A verb in `BAZEL_VERBS` but not the first list (`query`, `info`, `clean`, `mod`, …) goes to vanilla bazel untouched. A verb in neither is treated as a custom aspect task and forwarded verbatim.

Flag classification treats **bazel as the closed set**: a flag whose name is in the embedded `BAZEL_*_FLAGS` lists (value, boolean incl. `--no` negations, and the short forms `-c`/`-j`/`-k`/`-s`/`-t`) is wrapped as `--bazel-flag=` (post-verb) or `--bazel-startup-flag=` (pre-verb); space-separated value flags consume the next token and glue with `=`. Every other flag — aspect's own flags and anything unrecognized — passes through to aspect unchanged.

#### `aspect` not installed

If a command would route to aspect but `aspect` isn't on `PATH`, the wrapper prints install instructions (`curl -fsSL https://install.aspect.build | bash`). For a verb that's also a real bazel command (`build`/`test`/`run`/`coverage`) it then falls back to vanilla bazel so the command still runs; for aspect-only verbs (`lint`/`format`/custom tasks) it exits. Plain bazel verbs never need aspect.

#### `ASPECT_WRAPPER_SKIP`

A total escape hatch: forwards **everything** straight to vanilla bazel — no verb parsing, no list checks, no flag rewriting.

#### Anti-inception

When the wrapper routes a verb through aspect, aspect spawns its own child `bazel`. If `tools/bazel` is on PATH that child re-enters the wrapper, looping forever. aspect now sets `ASPECT_CLI_RUNNING=1` on every child bazel it spawns (via a new `bazel_command()` helper at every call site); the wrapper detects it on entry and forwards straight to the real bazel. Same pattern bazelisk uses for `BAZELISK_SKIP_WRAPPER`.

#### Notes

- Wrapper is bash 3.2 compatible (stock macOS `/bin/bash`); the test suite can be run under it with `WRAPPER_BASH=/bin/bash`.
- When `BAZEL_REAL` is unset, the real bazel is found on PATH via `type -aP bazel` (skipping the wrapper itself).
- `ASPECT_WRAPPER_TRACE` / `ASPECT_WRAPPER_QUIET` control the grey trace line.

#### Docs

[`tools/bazel.md`](tools/bazel.md) covers installation (copy both `tools/bazel` and `tools/bazel.md`), the routing model, the aspect-not-installed and skip behaviors, customization, anti-inception, and how to regen the embedded flag lists.

---

- Searched for relevant documentation and updated as needed: yes (`tools/bazel.md`)

#### Suggested release notes

- New `tools/bazel` reference wrapper lets customers expose Aspect CLI features through the familiar `bazel <verb>` interface, forwarding bazel-native flags (long, `=value`, space-separated, and short `-c`/`-j`) through to Bazel. See [`tools/bazel.md`](tools/bazel.md).
- Aspect now sets `ASPECT_CLI_RUNNING=1` on child `bazel` processes so wrapper scripts can detect they're a child of aspect and skip routing. Customers wrapping `aspect-cli` in their own scripts should propagate this variable through.

### Test plan

- Covered by existing `aspect-cli` unit tests.
- New `tools/bazel-test.sh` (100 checks, run under both bash 5 and macOS bash 3.2) covering verb routing, flag wrapping/passthrough, aspect-verb forwarding, pre-verb startup-flag classification, `ASPECT_WRAPPER_SKIP` total bypass, `ASPECT_CLI_RUNNING` bypass, `BAZEL_REAL` PATH fallback, the aspect-not-installed install hint + fallback, trace output, and edge cases.

### DIFF
--- a/tools/bazel
+++ b/tools/bazel
@@ -1,0 +1,869 @@
+#!/usr/bin/env bash
+# aspect-build/tools-bazel-wrapper
+#
+# tools/bazel — Bazel wrapper that transparently routes Aspect-specific
+# functionality through the `aspect` CLI without forcing developers to learn
+# a new command name.
+#
+# The marker comment above (`# aspect-build/tools-bazel-wrapper`) and the
+# ASPECT_BAZEL_WRAPPER_VERSION variable below identify this file as the
+# canonical Aspect-shipped wrapper. Aspect detects them to know that in
+# this workspace `bazel <verb>` is equivalent to `aspect <verb>` for
+# reproduce/fix commands and tooling that suggests "vanilla bazel"
+# alternatives. Do NOT remove these markers if you fork the script.
+#
+# Behavior (verb routing):
+#   - verb in ASPECT_VERBS_WITH_BAZEL_FLAGS  → exec aspect <verb> [rewritten]
+#       (the verbs aspect wraps plus the aspect verbs that drive bazel —
+#        bazel-native flags become --bazel-flag=<flag>; everything else
+#        passes through to aspect.)
+#   - verb in BAZEL_VERBS (not above)        → exec $BAZEL_REAL <args...>
+#       (query, info, clean, mod, version, … — vanilla bazel, untouched.)
+#   - any other verb (custom .axl task)      → exec aspect <verb> [args...]
+#       (forwarded verbatim — we don't assume an unknown verb is bazel's.)
+#   - no verb / ASPECT_CLI_RUNNING / SKIP    → exec $BAZEL_REAL <args...>
+#
+# When `aspect` is NOT on PATH but a command would route to it, the wrapper
+# prints install instructions. If the verb is also a real bazel command
+# (build/test/run/coverage) it then falls back to vanilla bazel so the command
+# still runs; otherwise (lint/format/custom .axl tasks, which bazel can't run)
+# it exits. Plain bazel verbs never need aspect and run regardless.
+#
+# Flag handling (for verbs routed to aspect with rewriting):
+#   Bazel is the closed set. A flag whose name is in one of the embedded
+#   BAZEL_*_FLAGS lists is wrapped as `--bazel-flag=<flag>` (post-verb) or
+#   `--bazel-startup-flag=<flag>` (pre-verb). Value-taking flags written
+#   space-separated (`--config ci`, `-c opt`) consume the next token and glue
+#   with `=` (`--bazel-flag=--config=ci`, `--bazel-flag=-c=opt`). Every other
+#   flag — aspect's own flags, and anything we don't recognize — passes
+#   through to aspect unchanged.
+#
+# Maintaining the lists (edit in your own repo copy):
+#   ASPECT_VERBS_WITH_BAZEL_FLAGS and the BAZEL_*_FLAGS lists at the top of
+#   this file are the source of truth. If Bazel adds a flag we don't know, or
+#   you add an aspect command that accepts bazel flags (e.g. a custom task
+#   that shells out to bazel), add it to the relevant list. An unknown bazel
+#   flag simply passes through to aspect until you add it.
+#
+# Drop this script at `tools/bazel` in your workspace and make it executable
+# (`chmod +x tools/bazel`). The `tools/bazel` hook is a Bazelisk feature (the
+# real bazel binary does not look for it; see
+# https://github.com/bazelbuild/bazelisk): Bazelisk execs this script and
+# passes the resolved real-bazel path in BAZEL_REAL. When BAZEL_REAL is unset
+# — the wrapper was invoked directly, or by an aspect child — we fall back to
+# finding the next `bazel` on PATH that isn't this wrapper.
+#
+# Environment variables:
+#   ASPECT_WRAPPER_SKIP=1    Forward EVERYTHING straight to bazel, untouched —
+#                            no verb parsing, no list checks, no flag rewriting.
+#                            A total escape hatch for a 1:1 vanilla-bazel
+#                            experience; reach for `aspect <verb>` directly when
+#                            you want the wrapped behavior.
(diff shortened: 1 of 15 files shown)


tools/bazel
```diff
@@ -0,0 +1,567 @@
...
+        t="${tokens[$i]}"
+        if [[ $passthrough -eq 1 || "$t" != -* ]]; then
+            eval "$out_name+=(\"\$t\")"; i=$((i + 1)); continue
+        fi
+        if [[ "$t" == "--" ]]; then
+            passthrough=1
+            eval "$out_name+=(\"\$t\")"; i=$((i + 1)); continue
+        fi
+        body="${t#--}"
+        name="${body%%=*}"
```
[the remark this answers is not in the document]
gregmagolan: Resolved

tools/bazel
```diff
@@ -0,0 +1,567 @@
...
+    fi
+    local self
+    self="$(cd "$(dirname "$0")" && pwd)/$(basename "$0")"
+    local candidate
+    while IFS= read -r candidate; do
+        if [[ "$(cd "$(dirname "$candidate")" && pwd)/$(basename "$candidate")" != "$self" ]]; then
+            echo "$candidate"
+            return
+        fi
+    done < <(command -v -a bazel 2>/dev/null || true)
```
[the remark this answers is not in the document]
gregmagolan: Resolved

tools/bazel.md
```diff
@@ -0,0 +1,124 @@
+# `tools/bazel` wrapper — transparent Aspect routing
+
+A drop-in `tools/bazel` shell script that lets developers keep typing `bazel` while still reaching Aspect-specific functionality (`lint`, `format`, `delivery`, …) and Aspect's value-added wrappers around `build` / `test` / `run`.
+
+This is the **source of truth** for customers who want to adopt Aspect CLI features without forcing their teams to learn a new command name. Copy `tools/bazel` from this repo into your own workspace, adjust the verb lists at the top, commit it, and you're done.
+
+## What it does
+
+When Bazel (or Bazelisk) sees a file at `tools/bazel` in your workspace, it execs that script instead of the real bazel binary, passing the resolved bazel path as `$BAZEL_REAL`. This wrapper uses that hook to dispatch each command to the right tool:
```
jbedard: Isn't it only Bazelisk?

tools/bazel.md
```diff
@@ -0,0 +1,124 @@
...
+export ASPECT_WRAPPER_SKIP=1
+```
+
+…and the wrapper forwards `build`/`test`/`run` straight to `$BAZEL_REAL` with flags untouched. Aspect-only verbs (`lint`, `format`, `delivery`, anything in `ASPECT_VERBS`) still route through aspect — those have no bazel equivalent, and aspect tasks can be defined with arbitrary names in `.axl`, so the wrapper can't safely assume an unknown verb is something bazel would understand.
+
+This is per-shell, per-developer. To make the choice repo-wide, edit `COMMON_VERBS` in the script (set to empty `()` to make every common verb route straight to bazel).
+
+## How it avoids infinite recursion
+
+When `tools/bazel` routes `build`/`test`/`run` through `aspect`, aspect then needs to spawn its own child `bazel` to do the actual work. If your `PATH` still has `tools/bazel` first (the normal case), that child `bazel` invocation re-enters the wrapper — which routes it back to `aspect` — which spawns `bazel` again — and so on forever.
```
jbedard: > If your `PATH` still has `tools/bazel` first (the normal case)

I don't think `PATH` ever has `tools/bazel`, the `PATH` only has `bazelisk` which inspects `tools/bazel` relative to the workspace you're in?

tools/bazel.md
gregmagolan: No. bazel itself respects tools/bazel

tools/bazel.md
gregmagolan: Again, bazel itself respects tools/bazel so we need to avoid recursion since aspect cli calls bazel directly

gregmagolan: Oh you know what, I was mistaken. tools/bazel _is_ a bazelisk only feature. I'll update the docs accordingly.

tools/bazel.md
gregmagolan: Oh you know what, I was mistaken. tools/bazel _is_ a bazelisk only feature. I'll update the docs accordingly.

Issue #1060, opened 2026-05-04
[Bug]: build.build_events() appearing to hang even though bazel build is complete


tgrue-openai:
### What happened?

I apologize for the poor bug report, please feel free to close if not helpful as is.

I'm experiencing an issue where when I, in python, call `for event in build.build_events()` from a call to :

```
    build = ctx.bazel.build(
        build_events = True,
        flags = [
            "--keep_going",
            "--build_runfile_links",
            "--remote_download_outputs=toplevel",
        ],
        current_dir = repo_root,
        *targets,
    )
```

Occasionally (some builds, once and a while) I'm getting hung waiting on the iterator.

My AI debugger believes this:

```
for event in build.build_events() blocks on the iterator’s receiver until Aspect’s BES broadcaster closes.

Build.wait() also joins the same BES stream thread. So if the stream thread fails to conclude, switching from “blocked in iterator” to “blocked in wait” does not really solve the class of hang.

build.build_events() subscribes to a non-replay broadcaster. Late subscribers only receive future events, and events sent before subscription are dropped. That is awkward for target-completion gates because cached/no-op builds can emit target events quickly.

BuildEventIterator.done() is not a safe polling primitive as implemented: it calls try_recv() internally, which can consume and discard an event when one is present. try_pop() is safer, but using it well requires a polling loop with build.try_wait(), and AXL does not appear to have a clean sleep/timer primitive.

Aspect’s file sink path still goes through Aspect’s BES pipe/stream machinery, so it does not avoid the stream lifecycle issue.
```

Doing what it suggest fixes my issue, which occurs intermittently and rarely, but it's a little hard to tell if this an accurate analysis.

Opening in case awareness of this bug and this analysis is helpful of tracking down a potential issue.

### Version

Development (host) and target OS/architectures: macos arm (tahoe 26.4.1 )

Output of `bazel --version`: bazel 9.1.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
Aspect CLI: 2026.13.5
From MODULE.bazel:
- bazel_lib: 3.2.2, with local patches
- aspect_rules_py: 1.11.2, with local patches
- aspect_rules_js: 3.0.1, with local patch
- aspect_rules_esbuild: 0.25.1, with local patch
- aspect_rules_ts: 3.7.0

Language(s) and/or frameworks involved:

### How to reproduce

```shell

```

PR #1197, opened 2026-06-03
feat(bazel): capture Bazel stderr through a processing pipeline

gregmagolan:
When a task sets `BazelTrait.capture_output`, the Bazel child's stderr is captured by the runtime, run through a processing pipeline, and forwarded to the real stderr — instead of being inherited straight to the terminal, which left no chance to pre-process it.

**Off by default.** With `capture_output` unset, stderr stays `Stdio::inherit()` and behavior is unchanged.

### How it works

- **Capture mode** — when capture is on, `bazel_runner.axl` picks the fd to match the destination stderr and sets Bazel's flags accordingly:
  - stderr is an interactive TTY → a **PTY** (`--isatty=1`), so Bazel keeps its live curses progress UI; bytes are forwarded near-verbatim.
  - otherwise (CI, redirected) → a **plain pipe** (`--isatty=0 --curses=no`), so Bazel emits clean newline-terminated lines.
- **Reader/forwarder** (`stream/output.rs`) — a dedicated thread reads the captured fd, splits records on `\r` and `\n` (so the `\r`-driven progress UI stays live), runs each through an ordered `LineProcessor` chain, and writes to the real stderr with a per-read flush. Single producer→consumer with a blocking write, so a slow terminal back-pressures the kernel buffer and ultimately Bazel — bounding memory without an unbounded queue. A stage can transform a record, drop it (with its boundary), or hold state and flush it at end-of-stream via `finish()`.
- **Lifecycle** — the fds are `O_CLOEXEC` so a concurrent `fork`+`exec` can't leak the write end (which would hang the forwarder); the PTY slave is dropped right after spawn so the master sees EOF; the stream is joined in `Build::wait` after the child is reaped, so all stderr flushes before the task's terminal summary.

### Built-in processors (`stream/processors.rs`)

All patterns are regex, compiled at `processor(...)` construction (invalid patterns error there).

- **Noise reduction** — `collapse_repeats` folds runs of consecutive identical lines into the first occurrence plus a `(last line repeated N more times)` annotation.
- **Pattern observers** — `match_patterns = {id: regex}` records hits, readable from Starlark via `build.output_matches()` (during or after the build; capped at 1024 hits, lines truncated to 2048 bytes) — the input for hooks and features, e.g. a `bazel_attempt_end` handler.
- **Health/liveness signals** — `fatal_patterns` latch `build.output_fatal` / `build.output_fatal_line` (e.g. a dead-server message), and `build.output_idle_ms` reports captured-output silence for stall detection.
- **Interactive respond/replace** — `respond_patterns = {id: regex}`: a matching line is **held before display** and surfaced on `build.output_events()`; AXL answers each event with `keep()`, `replace(text)`, or `drop()`, deciding what the user sees. Output order is always preserved (everything behind the held line waits, back-pressured by the kernel buffer). Fail-open by design: an unanswered event forwards the original line after `respond_timeout_ms` (default 1000) — a slow or absent consumer can delay matched lines, never wedge the build.

```python
processor = bazel.output.processor(
    tty = ctx.std.io.stderr.is_tty,
    collapse_repeats = True,
    match_patterns = {"oom": r"java\.lang\.OutOfMemoryError"},
    fatal_patterns = [r"Server terminated abruptly"],
    respond_patterns = {"cred": r"password|token"},
)
build = ctx.bazel.build("//...", output = processor, flags = ["--isatty=0", "--curses=no"])

events = build.output_events()
for _tick in sleep_iter(50):
    ev = events.try_pop()
    if ev:
        ev.replace("(credential elided)")  # or ev.keep() / ev.drop()
    if events.done:
        break

status = build.wait()
if build.output_fatal:
    print("bazel server died: " + (build.output_fatal_line or ""))
for id, line in build.output_matches():
    print("matched " + id + ": " + line)
```

Chain order: observer matchers first (they see every line as Bazel wrote it, including repeats), then the responder (may rewrite), then the collapser (dedups the final stream).

### Alternatives considered

**"Bazel can already write output to a file — why build this?"** A file (or the BEP `progress` events, which also carry console text) gives a *copy* of the output for later; this feature is an *interceptor* for the output the user is watching:

- **Modifying what the user sees** (dedup, respond/replace) requires sitting between Bazel and the terminal. With a file, the raw output still reaches the user (inherit + tee), or nothing does and we'd have to re-render Bazel's curses UI from a log — a terminal emulator, far more machinery than a pass-through pipe/PTY.
- **The live progress UI** — pointing stderr at a file makes it non-TTY and Bazel drops its interactive UI. PTY interposition preserves it, and is the standard tool (`script(1)`, `ssh`, `tmux` do exactly this).
- **Real-time reaction and liveness** — a pipe gives blocking reads, EOF-on-death, and back-pressure for free; files need tail/poll loops, size tracking, and cleanup, and have no death semantics. BEP progress events are batched/async (`fully_async` upload mode) and can themselves stall when the server wedges (#1060); the raw stderr fd is the most primitive, most trustworthy "is it alive" channel.
- The mechanism itself is small and precedented: `Stdio::piped()` instead of `Stdio::inherit()` plus a forwarding thread — the same shape as the existing BES FIFO reader and `ctx.std.process` piped spawns.

---

### Changes are visible to end-users: no

- Searched for relevant documentation and updated as needed: yes (Starlark API docstrings on `ctx.bazel.build/.test`, `bazel.output.processor`, `output_events`/`output_matches`, and `BazelTrait.capture_output`)
- Breaking change: no

### Test plan

- New Rust tests: `OutputStream` record splitting / CR-LF / partial-line / carry-cap / drop-with-boundary / `finish()`-at-EOF; `processors.rs` unit tests (collapse runs + trailing-run flush, regex matcher hits, responder keep/replace/drop verdicts, fail-open on timeout / no consumer / event dropped unanswered, fatal latch, match cap, truncation); a pipeline integration test proving observer→responder→collapser ordering with a live replace; Starlark eval tests against fake-bazel — `processor(...)` args + `output_*` attrs, and a full respond/replace loop over `output_events()` (new `stderr_chatter` basil scenario); end-to-end `capture.rs` round-trips through a real subprocess for both pipe and PTY, plus CLOEXEC no-hang regression tests.
- Covered by existing test cases: full AXL suite passes; no-opt-in build verified byte-unchanged.
- Manual testing: `capture_output` build of a real target with stderr redirected (pipe path) and under a PTY (`script`) — output forwarded, exit 0, no hang; flags confirmed via `ASPECT_DEBUG=1`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -362,6 +362,7 @@ dependencies = [
  "hyperlocal",
  "include_dir",
  "keyring",
+ "libc",
  "liquid",
  "liquid-core",
  "md-5",
@@ -370,6 +371,7 @@ dependencies = [
  "prost",
  "psl",
  "rand 0.8.5",
+ "regex",
  "reqwest",
  "semver",
  "serde",
@@ -1193,7 +1195,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1312,7 +1314,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "778e2ac28f6c47af28e4907f13ffd1e1ddbd400980a9abd7c8df189bf578a5ad"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -3284,9 +3286,9 @@ dependencies = [
 
 [[package]]
 name = "psl"
-version = "2.1.216"
+version = "2.1.218"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9826c2fe4dade07e9da1af2e18427257877bc8bc27aa810f6fbb52d907a480cc"
+checksum = "f558d33d882cb296fb65dbe0246b98b90e761fbe00c552701d2030eb5bbce63a"
 dependencies = [
  "psl-types",
 ]
@@ -3693,7 +3695,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.9.4",
- "windows-sys 0.52.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -3897,7 +3899,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b55fb86dfd3a2f5f76ea78310a88f96c4ea21a3031f8d212443d56123fd0521"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -4517,7 +4519,7 @@ dependencies = [
  "getrandom 0.3.3",
  "once_cell",
  "rustix 1.0.8",
- "windows-sys 0.52.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -5481,7 +5483,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


thesayyn:
you can already get bazel to write to a file today. what's the reason for inventing something new here?

gregmagolan:
> you can already get bazel to write to a file today. what's the reason for inventing something new here?

the intention is to be able to process bazel's output and forward it to stderr after process (filters, replacements, ammendments, etc...). If we asks bazel to send stderr to a file we'd loose the curses

PR #1234, opened 2026-06-15
fix(runnable): replay the target's env attribute when spawning built binaries

gregmagolan:
`aspect run` / `format` / `gazelle` / `delivery` spawn the built binary directly from its runfiles tree rather than shelling out to `bazel run`. That path never applied the target's `env = {...}` attribute — which `bazel run` injects at run time from the `RunEnvironmentInfo` provider — so any var declared in `env` was simply absent. A `set -u` script that reads one (e.g. `JQ_BIN`) failed with `JQ_BIN: unbound variable`.

The resolved `env` lives in neither the on-disk wrapper nor BES, so it can only be recovered via the provider. A small `runenv` aspect reads `RunEnvironmentInfo.environment` / `.inherited_environment` and writes them to JSON in an `aspect_runenv` output group. The runnable tracker collects that file from BES and `spawn` replays the env — `environment` values verbatim (already rootpath-expanded relative to the runfiles workspace dir, which is the spawn cwd), `inherited_environment` names copied from the process env. Target env is applied before the runfiles/`BUILD_*`/`ASPECT_*` vars so the runfiles plumbing wins on any collision.

The aspect is a no-op for rules that don't populate `RunEnvironmentInfo` (e.g. legacy native `sh_binary`, [bazelbuild/bazel#23658](https://github.com/bazelbuild/bazel/issues/23658)), in which case the spawn behaves as before.

---

`aspect run`, `format`, `gazelle`, and `delivery` now apply a target's `env` attribute when running the built binary, matching `bazel run`. Fixes `unbound variable` errors for scripts that read env-injected vars under `set -u`.

### Test plan

- New test cases added (`lib/runnable_test.axl`: env collection, partial JSON, absent provider, ctx-less, and the aspect-flags helper)
- Covered by existing test cases (`aspect dev test-runnable`, `aspect tests axl`)

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/delivery.axl
+++ b/crates/aspect-cli/src/builtins/aspect/delivery.axl
@@ -144,7 +144,7 @@ load("@aspect//lib/github.axl", "detect_commit_sha")
 load("@aspect//lib/health_check.axl", "HealthCheckTrait")
 load("@aspect//lib/lifecycle.axl", "Phase", "ProgressStyles", "ReproFixCommand", "TaskLifecycleTrait", "TaskUpdate", "apply_repro_fix_hooks", "dispatch_task_update", "setup_phase", "task_update")
 load("@aspect//lib/repro_commands.axl", "build_aspect_command", "print_repro_and_fix", "repro_prefix", "task_cli_path")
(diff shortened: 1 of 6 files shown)


thesayyn:
I'll make another attempt at this.

PR #1235, opened 2026-06-15
feat(gitlab): support self-hosted GitLab via configurable API base

b3cramer:
## Problem

The GitLab features hard-coded `https://gitlab.com/api/v4` as the API base (`DEFAULT_GITLAB_API` in `lib/gitlab.axl`). Every `gitlab.*` helper already accepted an `api_base` parameter, but no caller ever passed one — so commit statuses, the rolling MR status comment, and inline lint discussions could only ever target `gitlab.com`. Self-hosted GitLab instances had no way in.

## Change

Add `gitlab.resolve_api_base(env)` and thread the resolved base through every `gitlab.*` call in the three GitLab features.

Resolution precedence:

1. **`ASPECT_GITLAB_API_URL`** — explicit override, matching the existing `ASPECT_GITLAB_*` escape-hatch family in `gitlab_detect.axl` (for Aspect-Workflows runners on a non-GitLab CI host posting back to a self-hosted GitLab MR). Accepts either a full API base (`https://self-hosted.gitlab.com/api/v4`) or a bare instance root (`https://self-hosted.gitlab.com`), to which `/api/v4` is appended. A trailing slash is trimmed.
2. **`CI_API_V4_URL`** — set natively by GitLab CI on every job and already points at the running instance's API, so **self-hosted GitLab CI needs zero extra config**.
3. **`https://gitlab.com/api/v4`** — unchanged default when neither is set.

Threaded through all three features:
- `feature/gitlab_commit_statuses.axl` — the commit-status POST
- `feature/gitlab_status_comments.axl` — all six notes list/create/update/delete calls
- `feature/gitlab_lint_comments.axl` — `merge_requests.list_changes` + all five discussions calls (carried on the `gl` context dict)

---

- Searched for relevant documentation and updated as needed: no — flagged for a follow-up docsite update
- Breaking change (forces users to change their own code or config): no — default behavior on `gitlab.com` is unchanged

- GitLab integrations (commit statuses, MR status comments, inline lint discussions) now work against self-hosted GitLab. Under GitLab CI this is automatic via `CI_API_V4_URL`; elsewhere set `ASPECT_GITLAB_API_URL` to your instance's API base.

### Test plan

- Added five `resolve_api_base` unit cases to `lib/gitlab_test.axl` (default, `CI_API_V4_URL`, override precedence, bare-root `/api/v4` append, trailing-slash trim) plus the facade shape check.
- `aspect dev test-gitlab-client` → `gitlab.axl smoke: OK`
- `aspect tests axl` → `790 tests passed` (confirms all three edited feature files still load)

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/feature/gitlab_commit_statuses.axl
+++ b/crates/aspect-cli/src/builtins/aspect/feature/gitlab_commit_statuses.axl
@@ -172,6 +172,7 @@ def _gitlab_commit_statuses(ctx: FeatureContext):
     project_path = mr["project_path"]
     sha = mr["sha"]
     mr_iid = mr["mr_iid"]
+    api_base = gitlab.resolve_api_base(ctx.std.env)
 
     _ci_links_base = _collect_ci_links(ctx, workflows_results_url(ctx.std))
 
@@ -233,6 +234,7 @@ def _gitlab_commit_statuses(ctx: FeatureContext):
             name = _state["_status_name"],
             target_url = target_url or None,
             description = description or None,
+            api_base = api_base,
             auth_kind = "bearer",
         )
         if not result["success"]:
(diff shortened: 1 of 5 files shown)


b3cramer:
Docs for this (and the auth follow-up #1237): aspect-build/site#1072

PR #1239, opened 2026-06-16
feat(wrapper): add tools/bazel wrapper install/uninstall subcommand

jeffpignataro:
## Summary

Adds an `aspect wrapper` task group that vendors the [`tools/bazel`](https://github.com/aspect-build/aspect-cli/blob/main/tools/bazel) Bazelisk wrapper into a Bazel workspace, so a team can keep typing `bazel` while transparently routing through the Aspect CLI. Today that setup is fully manual (download the script, drop it at `tools/bazel`, `chmod +x`, figure out the env vars); this automates it end-to-end. Implements [ENG-1669](https://linear.app/aspect-build/issue/ENG-1669).

## What changed

New `crates/aspect-cli/src/builtins/aspect/wrapper.axl` with two tasks (registered in the builtin `MODULE.aspect`):

- **`aspect wrapper install`**
  - Detects the workspace root by walking ancestors for a Bazel marker (`MODULE.bazel` / `MODULE.bazel.lock` / `REPO.bazel` / `WORKSPACE` / `WORKSPACE.bazel`); refuses to run outside a workspace.
  - Downloads the wrapper from a **pinned release tag** of `aspect-build/aspect-cli` (never HEAD) via `raw.githubusercontent.com`. `--version=<tag>` selects the tag; the default `latest` resolves the newest published release through the GitHub releases API.
  - Writes `tools/bazel` (`0o755`, via `ctx.http().download(mode=)` so the executable bit lands without shelling out to chmod) and its companion `tools/bazel.md` (`0o644`), creating `tools/` if missing.
  - Prints copy-pasteable shell snippets for bash / zsh / fish (detected from `$SHELL`, or `--all-shells`). **Shell rc files are never mutated** — instructions are printed only.
  - Idempotent: a re-run that byte-matches the target no-ops; a differing target is reported as an upgrade (interactive confirm unless `--yes`). `--check` is a CI gate that exits non-zero when `tools/bazel` is missing or stale and writes nothing; `--dry-run` previews without writing.
- **`aspect wrapper uninstall`** — removes `tools/bazel` and `tools/bazel.md`, refusing to delete a `tools/bazel` that lacks the `# aspect-build/tools-bazel-wrapper` marker (won't clobber a hand-rolled wrapper).

Pure logic (root detection, URL building, version parsing, install classification, shell detection + snippet rendering) lives in `lib/wrapper.axl` so the command file stays thin and the logic is unit-testable without IO. Records/enums and full type annotations per the repo AXL conventions.

## Testing

- 22 unit tests in `lib/wrapper_test.axl` — run with `aspect dev test-wrapper` (registered in `.aspect/config.axl`).
- Manual end-to-end against live GitHub in throwaway workspaces: fresh install, no-op re-run, upgrade (`--yes`), `--check` (verified real exit 1 on both stale and missing), `--dry-run` (no write), subdirectory/ancestor root detection, refuse-outside-workspace, `uninstall` + foreign-file marker safety, and the default `latest`-release path (resolved `v2026.25.2`).


jeffpignataro:
Thanks for the review — all three addressed in b4b5c363:

1. **Exec bit before no-op** — `classify_install` now takes the file's executable bit. Matching bytes without `chmod +x` classify as a repair, not a no-op: `install` re-applies `0o755` and `--check` exits non-zero (`"...right contents but is not executable"`). `download` only sets the mode on create, so the existing file is removed before re-downloading. Gated to non-Windows, where the bit is always `false`.

2. **Foreign overwrite in non-interactive mode** — an upgrade now refuses to overwrite a `tools/bazel` lacking the `# aspect-build/tools-bazel-wrapper` marker unless `--yes` is passed, in *any* mode (TTY or not). Symmetric with `uninstall`'s marker check, so CI/bootstrap runs can't silently clobber a hand-maintained wrapper. Our own (marked) wrapper still upgrades non-interactively.

3. **Moving refs in `--version`** — kept `--version` flexible (a commit SHA is an immutable, reproducible pin that a tag-only check would wrongly reject, and no extra API round-trip is needed), but removed the contradiction: the "never HEAD" guarantee now scopes to the default (latest release tag), and the help/docstring state that `--version` takes a tag or SHA for reproducibility, or a branch (not reproducible).

Verified: 24 unit tests (added exec-bit cases) + e2e for repair, foreign-overwrite refusal, and the `--yes` override.

crates/aspect-cli/src/builtins/aspect/lib/wrapper.axl
```diff
@@ -0,0 +1,171 @@
...
+    """Decide what an install would do, given the on-disk wrapper content
+    (`existing`, None when absent), the freshly-fetched `target`, and whether
+    the on-disk file currently has its executable bit set.
+
+    Matching bytes that lack the executable bit are classified `upgrade`, not
+    `noop` — Bazelisk execs `tools/bazel` directly, so a copy committed or
+    pulled without `chmod +x` is a broken hook that install must repair rather
+    than report as up to date. Callers on platforms where the bit is
+    meaningless (Windows) pass `existing_executable = True` so the content
+    comparison alone decides.
```
thesayyn: can we update these comments?  i recently updated our instructions to write shorter comments https://github.com/aspect-build/aspect-cli/blob/main/CLAUDE.md

jeffpignataro: Cleaned up the comment-spam quite a bit.

PR #1278, opened 2026-06-23
fix: inject args and env for runnable.axl

thesayyn:
Built a runinfo aspect that replays each target's executable, args attribute, and RunEnvironmentInfo env so aspect run/format/gazelle/delivery match bazel run byte-for-byte, exposed it through a single load-bearing runnable().flags attribute (with a --remote_download_outputs floor that upgrades minimal→toplevel), and cleaned up the call sites so no command hand-assembles spawn flags.

---

`aspect run` and `runnable.axl` now supports `env` and `args` of runnable targets. 

### Test plan

- New test cases added


thesayyn:
@codex

PR #1303, opened 2026-06-30
feat(ci): add `aspect ci runner-metadata` and `aspect ci runner-health-check`

gregmagolan:
Adds two standalone `aspect ci` commands that run, on demand, the two diagnostics that otherwise run only inside a Bazel task's Setup phase:

- **`aspect ci runner-metadata`** — print the Aspect Workflows runner metadata table (cloud/region/instance, cache and storage paths, warming status, bootstrap-logs command).
- **`aspect ci runner-health-check`** — wait for cache warming, print the last runner health check and warming status, probe the runner's Bazel server, and signal the runner unhealthy if it is wedged. Exits non-zero when the server is unhealthy.

Both are thin wrappers over the existing `print_environment_info` / `agent_health_check` library functions, in the same spirit as the standalone `aspect ci warming` and `aspect ci bazelrc` commands. Off a Workflows runner there is nothing to inspect, so each prints an informational line and exits 0.

---

- Add `aspect ci runner-metadata` to print the Aspect Workflows runner metadata table on demand.
- Add `aspect ci runner-health-check` to run the Aspect Workflows runner health check on demand.

### Test plan

  - Off a runner: `aspect ci runner-metadata` and `aspect ci runner-health-check` each print an info line and exit 0.
  - With `ASPECT_WORKFLOWS_RUNNER` (+ `ASPECT_WORKFLOWS_RUNNER_*`) set: `aspect ci runner-metadata` prints the metadata table.
  - On a Workflows runner: `aspect ci runner-health-check` runs the Bazel server probe and exits 0 (healthy) / non-zero (unhealthy).

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/MODULE.aspect
+++ b/crates/aspect-cli/src/builtins/aspect/MODULE.aspect
@@ -19,6 +19,8 @@ use_task("wrapper.axl", "uninstall")
 use_task("init.axl", "init")
 use_task("bazelrc.axl", "bazelrc")
 use_task("warming.axl", "warming")
+use_task("runner_metadata.axl", "runner_metadata")
+use_task("runner_health_check.axl", "runner_health_check")
 
 use_feature("feature/artifacts.axl", "ArtifactUpload")
 use_feature("feature/circleci_test_results.axl", "CircleCITestResults")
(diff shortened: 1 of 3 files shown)


thesayyn:
should the namespace be `aspect runner`? none of these actually about CI, its more about runners.

`aspect runner metadata`
`aspect runner health` 

reads better imo

PR #1334, opened 2026-07-21
feat(delivery): local-hash fallback for targets not in the remote cache

jeffpignataro:
Delivery's change-detection digest is read from the remote cache (`--experimental_remote_require_cached` denies any action not in the cache), so a deliverable whose actions never upload — e.g. large ML targets tagged `no-remote-exec` — can never resolve a digest and is dropped from delivery with a FAIL/WARN. `--warn-unresolved-digest` only downgrades the error; the target still doesn't deliver. [ENG-1799](https://linear.app/aspect-build/issue/ENG-1799/axl-delivery-manifest-support-targets-that-are-not-in-the-remote-cache)

This adds an opt-in `--local-hash-fallback` flag (default **off** — with the flag unset, behavior is byte-identical to today). When set, a new phase 2b runs after the remote checksum phase for deliverables that are still unresolved **and** deterministic: it rebuilds just those targets with a local-executing variant of the hashsum aspect under `--nouse_action_cache`, reads each `DeliveryHash` action digest from the execution log, and merges it into the digest set — so the targets deliver and are change-tracked like any other.

**Design notes**

- The local digest is the same REv2 action digest the remote path reads (`ExecLogEntry.Spawn.digest.hash`), and `no-local` is a strategy selector that doesn't enter the Action proto — verified empirically that the digest is identical across the aspect refactor and stable across fresh output bases. A target hashed locally in one run and read from the remote cache in another yields the same key: no spurious re-delivery.
- `--nouse_action_cache` defeats the local action-cache blind spot (an unchanged `DeliveryHash` spawn is otherwise a cache hit and absent from the execlog) without changing the digest; with a disk/remote cache present the upstream actions are cache hits, not rebuilds.
- Volatile targets (any `cacheable=false` action, e.g. a stamped `py_binary`'s `PyWriteBuildData`) are excluded from the fallback — their digest is unstable by construction — and keep today's FAIL/WARN + `--warn-unresolved-digest` behavior.
- Phase 2b is best-effort: a fallback build failure warns and falls through to the existing unresolved-digest path; it can't change exit codes on its own.
- The pure logic (eligibility gate, execlog digest extraction, merge) lives in `private/lib/delivery_digests.axl` with unit tests (`aspect dev test-delivery-digests`).

**Dogfood / e2e**

`examples/deliverable` gains `no_remote_exec_deliverable` — a deterministic target tagged `no-remote-exec` (the in-tree ENG-1799 repro, contrast to the volatile `py_deliverable` fixture). The CI `delivery-task` job now passes `--local-hash-fallback` and repeats the `delivery-gl` invocation: the first run must deliver the fixture via the local hash, the repeated run must SKIP it under the same state prefix — proving digest stability and change-detection continuity end-to-end.

---

- Searched for relevant documentation and updated as needed: no — customer docs for the delivery-manifest guide follow in a separate docs-site PR

- New `aspect delivery --local-hash-fallback` flag (default off): deliver deterministic targets that are not in the remote cache (e.g. large targets tagged `no-remote-exec`) by computing their change-detection digest with a local build. Volatile (stamped/non-cacheable) targets are unaffected and still require `--warn-unresolved-digest`.

### Test plan

- New unit tests: `aspect dev test-delivery-digests` (eligibility gate, execlog digest extraction, merge semantics)
- Existing suites: `aspect dev test-delivery-results` (11 tests) green; `aspect delivery --help` shows the new flag
- Live e2e: the `delivery-task` CI job on this PR runs delivery three times — verify `no_remote_exec_deliverable` shows OK (delivered via local hash) on the first `delivery-gl` invocation and SKIP on the repeated one, while `py_deliverable` stays WARN
- Verified in an isolated Bazel workspace that the `DeliveryHash` action digest is stable across fresh output bases, content-sensitive, and identical between the `hashsum` and `hashsum_local` aspect variants

### DIFF
--- a/.aspect/config.axl
+++ b/.aspect/config.axl
@@ -16,6 +16,7 @@ load("@aspect//private/lib/bazelrc_test.axl", "bazelrc_tests")
 load("@aspect//private/lib/bes_sinks_test.axl", "bes_sinks_tests")
 load("@aspect//private/lib/ci_test.axl", "ci_tests")
 load("@aspect//private/lib/circleci_test.axl", "circleci_tests")
+load("@aspect//private/lib/delivery_digests_test.axl", "delivery_digests_unit_tests")
 load("@aspect//private/lib/delivery_results_test.axl", "delivery_results_unit_tests", "delivery_template_snapshot_tests")
 load("@aspect//private/lib/deployment_flags_test.axl", "deployment_flags_tests")
 load("@aspect//private/lib/format_results_test.axl", "format_template_snapshot_tests")
@@ -388,6 +389,10 @@ def config(ctx: ConfigContext):
     # Delivery results unit: init_data/add_result/build_manifest schema shape.
     # Run with: aspect dev test-delivery-results
     ctx.tasks.add(delivery_results_unit_tests)
+
+    # Delivery digests unit: --local-hash-fallback eligibility + digest merge.
+    # Run with: aspect dev test-delivery-digests
+    ctx.tasks.add(delivery_digests_unit_tests)
 
     # format_results template snapshots.
     # Run with: aspect dev test-format-template-snapshots
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -128,9 +128,10 @@ delivery-task:
     ASPECT_WORKFLOWS_DELIVERY_FORCE_TARGETS: $FORCE_TARGETS
   script:
     - .aspect/bootstrap.sh
-    # --warn-unresolved-digest: the deliverable set includes py_deliverable, a
-    # stamped py_binary whose PyWriteBuildData action is uncacheable and never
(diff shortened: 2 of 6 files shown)


thesayyn:
Hmm, i don't think this is the correct fix here. We actually don't care how the action is produced as long as it can be presented in the merkle tree. Since https://github.com/aspect-build/aspect-cli/pull/1317 we stop trashing the skyframe state so i am surprised that you even hit the  `--experimental_remote_require_cached` gate for the non-RE target.   

Also delivery no longer uses: `--nouse_action_cache` so i am not sure why the robot thinks thats the case. 

There is one thing i am not clear on:  does `no-remote-exec` imply `no-remote-cache`, that might do it.


In any case this is not what we want. Local checksum is not the answer, we learned that the hard way.

PR #1326, opened 2026-07-15
refactor: cleanup bad patterns

thesayyn:
### Test plan

- Covered by existing test cases
- New test cases added

PR #1364, opened 2026-07-31
feat(runtime): send-side BES event filtering, iterator heartbeat

jbedard:
Allow BES event filtering within rust on the send-side. Extracted from #1326

### Test plan

- New test cases added


jbedard:
Moved to #1366

PR #1370, opened 2026-08-04
fix(cli): IDEs with plugins using BSP can now use aspect cli instead of vanilla bazel

mcook42:
The JetBrains Bazel plugin (`bazelbsp:3.2.0`) drives IDE sync by shelling out to `bazel` on `PATH`. In a repo with this project's `tools/bazel` hook, that hook routes `build`/`test` to `aspect` and rewrites every Bazel-native flag to `--bazel-flag=…`. Three independent defects in that path caused an IntelliJ sync to fail outright, so the only way to use the plugin was to set `ASPECT_WRAPPER_SKIP=1` and give up aspect's BES streaming, remote config, and task reporting entirely. See the [Build Server Protocol (BSP)](https://build-server-protocol.github.io/docs/specification) used by the plugin, if curious.

Each fix is a separate commit.

A caller-forwarded `--target_pattern_file` conflicted with the default target pattern. `run_bazel_task` only suppressed its own setting, but Bazel's `--bazel-flag=--target_pattern_file=<path>` was invisible to the runner. As a result, `ctx.args.targets` defaulted, causing Bazel to reject the invocation with an error. Both sync passes failed, with no last-wins rule, so the runner can only decide to invent a pattern, while precedence rules are Bazel's responsibility.

### 2. A caller's `--build_event_binary_file` was silently never written

`Build::spawn` appends the CLI's own `--build_event_binary_file` after every user flag. Bazel's option is single-valued, so last-wins meant that a caller who asked for a BEP file got a path that had been created but never written to. The plugin *parses sync results from that file*, so even with (1) fixed, it would have read an empty file.

Fixed by collecting a file sink for the caller's path in `collect_bes_sinks`, rather than stripping or reordering flags. The CLI's path still wins on the command line; the caller's file is recreated from Bazel's own byte stream, since file sinks share the BES reader's raw-bytes path.

### 3. `bazel config` was misrouted to `aspect`

```
$ bazel config --dump_all --output=json
error: unrecognized subcommand 'config'
```

`BAZEL_VERBS` in `tools/bazel` was generated from `bazel help`'s "Available commands" list, which *hides* `config`. A verb missing from that table is treated as a custom aspect task and routed to `aspect`. The plugin calls `bazel config --dump_all --output=json` during sync and parses stdout as JSON, so this failed the sync on its own.

The authoritative list is `bazel help completion`'s `BAZEL_COMMAND_LIST`, which includes hidden commands — verified against Bazel 9.0.1, `config` is the sole difference. The regeneration recipe and *why it must not be `bazel help`* are now recorded in both the script comment and `tools/bazel.md`:

```
bazel help completion | sed -n 's/^BAZEL_COMMAND_LIST="\(.*\)"$/\1/p'
```

The gap also silently degraded the pre-verb disambiguation walk, since `KNOWN_VERBS_STR` is built from the same table.

Also ignores `.idea/` and `.bazelbsp/`, both of which are generated by the plugin.

---

- Searched for relevant documentation and updated as needed: yes — `tools/bazel.md` documents `BAZEL_VERBS` and now carries the regeneration source and rationale
- Breaking change (forces users to change their own code or config): no — each change either honors a flag that was previously ignored or routes a verb that previously errored

The Aspect CLI now serves as the `bazel` binary for IDE and BSP tooling, such as the IntelliJ Bazel plugin. A caller-forwarded `--target_pattern_file` no longer collides with the CLI's default target pattern; a caller's `--build_event_binary_file` is now written instead of being silently clobbered by the CLI's own; and `bazel config` reaches Bazel through the `tools/bazel` wrapper rather than erroring as an unknown aspect subcommand. An IntelliJ sync no longer needs `ASPECT_WRAPPER_SKIP=1`, so IDE-driven builds keep BES streaming, remote config, and task reporting.

### Test plan

- New test cases added — `target_patterns` in `bazel_runner_test.axl` (default forwarded when nothing supplies patterns; suppressed by a forwarded flag; suppressed by aspect's own arg; explicit patterns always forwarded; the lookup is command-scoped). The `=`-form / two-token / last-wins matching is deliberately *not* re-asserted against a faked `rc` — that behavior is `flag_value_list`'s and is covered in `crates/bazelrc/src/lib.rs` (`flag_value_eq_form_last_wins`, `flag_value_two_token_form`). Plus `bazel_bep_file` resolution and the `collect_bes_sinks` file-sink append in `bazel/build_events_test.axl`
- Covered by existing test cases — `aspect tests axl` (910 cases) and the full `aspect test` suite (26/26) both pass
- New CI coverage — the existing `test-flags-task` step in both `.buildkite/pipeline.yaml` and `.github/workflows/ci-workflows.yaml` now runs a single `aspect build` in the plugin's exact shape and asserts a non-empty BEP at the caller's path. The unit tests cover the two resolvers in isolation; this asserts that the *composed* command line is accepted by Bazel and that the file lands on disk.

The failure mode was captured on `main` first, so each result below is a before/after rather than an assertion that nothing broke:

| Check | Before (on `main`) | After |
|---|---|---|
| Plugin-shaped invocation | `cannot both be specified`, exit 2 | exit 0, `Found 1 target` from the file |
| Caller's BEP file | **never created** | **non-empty** (~48 KB locally), contains the target label |
| `WARNING: BES was not properly closed` | present | gone |
| Two-token `--target_pattern_file <path>` | `cannot both be specified` | exit 0, BEP non-empty (~50 KB) |
| Pattern file from a `.bazelrc` line | default pattern collided | exit 0 — now suppresses the default too |
| aspect's `--target-pattern-file` **+** a forwarded one | base-flag position: forwarded won | aspect's arg wins (Bazel last-wins, verified by which target built) |
| `bazel config --dump_all --output=json` | `unrecognized subcommand 'config'`, exit 2 | clean JSON on stdout, exit 0 |
| `BAZEL_VERBS` vs `BAZEL_COMMAND_LIST` | differs by `config` | identical |
| Default target pattern (no flag forwarded) | `-- //... -//exclude/...` | unchanged |
| Explicit pattern **+** forwarded file | Bazel's collision error | unchanged — still Bazel's own error, exit 2 |

```bash
# Build and put the binary on a stable path (see the note below on why the copy matters).
ASPECT_WRAPPER_SKIP=1 bazel build //:cli
mkdir -p target/ci && cp -f "$(ASPECT_WRAPPER_SKIP=1 bazel info bazel-bin)/crates/aspect-cli/aspect-cli" target/ci/aspect-cli

# Unit tests (these are `group = ["dev"]` tasks, not bazel test targets).
aspect dev test-bazel-runner
aspect dev test-bes-sinks
aspect tests axl

# Replay the plugin's invocation shape.
PATTERNS=$(mktemp); echo "//examples/test_states:always_pass" > "$PATTERNS"
BEP=$(mktemp); rm -f "$BEP"
aspect build \
  --bazel-flag=--target_pattern_file="$PATTERNS" \
  --bazel-flag=--build_event_binary_file="$BEP" \
  --bazel-flag=--build_event_binary_file_upload_mode=wait_for_upload_complete \
  --bazel-flag=--tool_tag=bazelbsp:3.2.0
test -s "$BEP" && grep -qa always_pass "$BEP" && echo "caller's BEP written with real events"

# A pattern file from an rc file (or a --config expansion) suppresses the default too,
# which the earlier ctx.args.bazel_flags scan could not see.
RC=$(mktemp); echo "build --target_pattern_file=$PATTERNS" > "$RC"
aspect build --bazel-startup-flag=--bazelrc="$RC"

# Both spellings at once: aspect's own arg is appended last, so Bazel's last-wins picks it.
OTHER=$(mktemp); echo "//examples/test_states:always_fail" > "$OTHER"
aspect build --target-pattern-file="$PATTERNS" --bazel-flag=--target_pattern_file="$OTHER"  # builds always_pass

# Verb routing. Note 2>/dev/null, not 2>&1 — Bazel's "INFO: Invocation ID" goes to
# stderr and would corrupt the parse. The plugin reads stdout only.
comm -3 \
  <(ASPECT_WRAPPER_SKIP=1 bazel help completion | sed -n 's/^BAZEL_COMMAND_LIST="\(.*\)"$/\1/p' | tr ' ' '\n' | sed '/^$/d' | sort -u) \
  <(sed -n '/^BAZEL_VERBS=(/,/^)/p' tools/bazel | sed '1d;$d' | tr ' ' '\n' | sed '/^$/d' | sort -u)
bazel config --dump_all --output=json 2>/dev/null | python3 -c 'import json,sys; json.load(sys.stdin); print("valid JSON")'
```

#### End-to-end in IntelliJ IDEA 2026.2 with the Bazel plugin

Both the plugin's **build** and **Sync Project** actions complete against this branch. The build was confirmed from aspect's own BEP for the invocation (`/tmp/<uuid>.bep.binpb`), which carries the plugin's full fingerprint:

- tool tag `bazelbsp:3.2.0`
- `--target_pattern_file=…/IntelliJIdea2026.2/tmp/targets-8273310712140375819`
- `--build_event_binary_file=/tmp/bazel-bep-output9212484283891080652.tmp` alongside `--build_event_binary_file_upload_mode=wait_for_upload_complete`
- targets resolved *from that file* — `//examples/deliverable:py_deliverable`, `:py_deliverable2`
- outcome `SUCCESS`, with no `cannot both be specified` and no `unrecognized subcommand`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of caller-provided Bazel Build Event Protocol (BEP) output files.
  - Preserved BEP file output alongside other configured event sinks.
  - Improved forwarding and resolution of target patterns, including target-pattern files.
  - Corrected routing for the Bazel `config` command.

- **Documentation**
  - Expanded guidance for IDE/BSP build flag forwarding and BEP output.
  - Updated Bazel command-list customization instructions.

- **Chores**
  - Ignored IDE-generated project and Bazel plugin state files.


crates/aspect-cli/src/builtins/aspect/private/lib/bazel_flags.axl
```diff
@@ -151,6 +156,32 @@ def bazel_flag_args(build_phrase: str) -> dict:
...
+    Reads `ctx.args.bazel_flags` rather than the parsed rc, so it does NOT see a
+    flag from `.bazelrc` or a `--config` expansion. That is deliberate: its one
+    caller resolves target patterns before `parse_rc` runs (the resolved pattern
+    file has to be in the `base_flags` the rc is parsed from), and the flags it
+    asks about are per-invocation ones no rc file should carry. Anything that can
+    wait for the rc should use `rc.flag_value` instead.
+    """
+    eq_prefix = name + "="
+    value = ""
+    for i, flag in enumerate(bazel_flags):
```
thesayyn: this sounds really unprincipled/workaroubd. can we understand what's really wrong and fix it? 

I'd rather let bazel decide what's last to win rather than us deciding it.

jbedard:
This seems like 3 separate issues that can have independent fixes?

mcook42:
> This seems like 3 separate issues that can have independent fixes?

You're right that these are three separate issues, and I could split them into three separate PRs. However, until all three of these issues are fixed, the IntelliJ/JetBrains Bazel IDE plugin won't work. The three fixes in this PR need to land before the plugin works with aspect CLI, so in my mind it was best to land them all in a single review.

I'm happy to close this PR and open three independent PRs for smaller reviews if that's preferred.

crates/aspect-cli/src/builtins/aspect/private/lib/bazel_flags.axl
mcook42: It was a workaround, and I should have caught this before posting, so thanks for holding me accountable.

The runner already avoided sending a target pattern when Aspect's own --target-pattern-file flag was set. The bug is that it only checked that one spelling. When the pattern file arrives as a forwarded `--bazel-flag=--target_pattern_file=<path>`, as an IntelliJ sync sends, the runner never sees it and falls back to its default pattern of `...`. Bazel then gets both that pattern and the pattern file and refuses to run, since it has no last-wins rule for that pair.

The fix is to check the flags Bazel will actually run with, rather than a specific arg. We now ask the parsed command with rc.flag_value("--target_pattern_file", ...), which detects the flag regardless of how it arrived, and skips sending a pattern if it is set. If the user typed patterns, we always send them and let Bazel report the conflict, so our own error in that case is removed.

Getting to that point took a reordering. The parsed command did not yet exist when patterns were being decided, because we were adding Aspect's own pattern file flag to the same list of flags that gets parsed. Moving that flag out of that list let target resolution happen after the parse, which is what makes rc.flag_value available to ask.

jbedard:
I still keep getting confused when trying to review this. Would it be easy to split parts of this into separate PRs? Things like the sink fix, targets files etc.

If I see something easy to understand I'm significantly more likely understand+approve before getting pulled into something else and forgetting all the context here...

mcook42:
Superseded — split into three separate PRs per review feedback, one per fix:

1. #1388 — `--target_pattern_file` conflict
2. #1389 — `--build_event_binary_file` silently clobbered
3. #1390 — `bazel config` misrouted to `aspect`

Closing this in favor of those.

PR #1372, opened 2026-08-05
refactor(bazel): move the hook dispatchers next to the trait they fire

jbedard:
`dispatch_bazel_attempt_end` / `dispatch_bazel_build_end` walk `BazelTrait.bazel_attempt_end` / `.build_end`, so they belong with the trait definition rather than with the build/test task runner. Every task that drives its own spawn loop already loads `@aspect//bazel.axl`, so they now reach the dispatchers through that facade instead of importing the private `bazel_runner.axl`.

No behaviour change.

### Test plan

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/bazel/trait.axl
+++ b/crates/aspect-cli/src/builtins/aspect/bazel/trait.axl
@@ -1,4 +1,5 @@
-"""The `BazelTrait` interface, Bazel exit codes, and the retry budget constants.
+"""The `BazelTrait` interface, Bazel exit codes, the retry budget constants, and
+the `dispatch_*` helpers that fire the trait's lifecycle hooks.
 
 Exit codes correspond to Bazel's ExitCode.java:
 https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/util/ExitCode.java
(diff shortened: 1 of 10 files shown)


crates/aspect-cli/src/builtins/aspect/bazel/trait.axl
```diff
@@ -85,3 +86,29 @@ BazelTrait = trait(
     bazel_attempt_end = attr(list[typing.Callable[[TaskContext, int], None]], default = [], description = "Hooks called once after EACH Bazel invocation attempt (including retries), with that attempt's exit code. Runs after the wait() returns and before the retry decision, so a hook can take corrective action (e.g. clean stranded sandbox state) that might let the next retry succeed. Distinct from `build_end`, which fires once after the final attempt."),
     build_end = attr(list[typing.Callable[[TaskContext, int], None]], default = [], description = "Hooks called once after the Bazel invocation completes (after all retries), with the final exit code"),
 )
+
+def dispatch_bazel_attempt_end(ctx, bazel_trait, exit_code):
```
thesayyn: So much comment for a simple function :(

crates/aspect-cli/src/builtins/aspect/bazel/trait.axl
```diff
@@ -85,3 +86,29 @@ BazelTrait = trait(
...
+    errors). Inside a retry loop, fires BEFORE the retry decision so a hook
+    can repair state that would otherwise make the next attempt fail the same
+    way (e.g. `bazel.recover_poisoned_sandbox` for bazelbuild/bazel#23880).
+    Outside a retry loop (single-attempt tasks), fires exactly once.
+    """
+    for handler in bazel_trait.bazel_attempt_end:
+        handler(ctx, exit_code)
+
+def dispatch_bazel_build_end(ctx, bazel_trait, exit_code):
+    """Fire every `bazel_trait.build_end` hook with the FINAL `exit_code`.
```
thesayyn: Same here.

thesayyn: these probably don't deserve to be a utility function. what they do could easily be inline at callsite.

jbedard: They are invoked enough that I would rather not inline them, but yes this comment is way too large and just spam and I'll reduce that at least.

thesayyn (approved):

PR #1374, opened 2026-08-05
feat(bazel): featurize deployment behind a generic BazelTrait seam

jbedard:
Tasks no longer know about deployments. `BazelTrait` grows the seams a feature needs to wire a build:

  base_flags          default flags prepended BEFORE the user's --bazel-flag,
                      so a user override still wins (extra_flags appends after
                      and therefore beats them)
  bes_backends        extra CLI-streamed BES backend URIs
  bes_results_sources (backend, viewer) pairs the build-result link is drawn
                      from, matched against what the build actually reports to
  rc_flags            hooks called with (ctx, resolved rc, command) after the
                      .bazelrc is parsed, for flags that depend on the FINAL rc

A new `Deployment` feature resolves `--remote` / `--deployment` and fills those in, including the `--remote_header` / `--bes_header` auth hook that `bazel/flags.axl` used to own. A bazel-driving task now applies one generic call, `bazel.rc_flags(ctx, rc, command)`, and stays ignorant of what a feature injected.

`advertised_results_url` / `bes_results_url_flag` move to `bazel/build_events.axl` and take the registered `(backend, viewer)` pairs rather than a `DeploymentFlags` record, so the BES layer no longer depends on the deployment module. Their tests move with them.

The unused `BazelTrait.flags` / `.startup_flags` transforms are dropped; `extra_flags` / `base_flags` cover both directions.

`--remote` and `--deployment` keep their names via an explicit `long =` (a feature arg is otherwise exposed as `--<feature>:<arg>`), and now reach every bazel-driving task rather than just build/test.

### Test plan


jbedard:
@codex

thesayyn (approved):

PR #1377, opened 2026-08-07
feat(cli): make the CLI surface discoverable and pipe-safe for AI agents

jeffpignataro:
AI coding agents struggle to drive this CLI, and the reflexive explanation — "AWS CLI and gcloud are in the training data and we aren't" — is only half true. Model knowledge of those tools' flags is substantially stale; what actually makes them work in an agent loop is the recovery loop: cheap structured help, machine-readable output on read commands, and errors that name the next command. More decisively, **Aspect's surface is per-repo** — `aspect <task>` means whatever this repo's `.aspect/*.axl` defines, and customer tasks have by definition never been in any training set. So "get into the weights" is both unreachable and not the bottleneck. This makes the surface answerable at runtime instead.

**Launcher progress.** The download progress renderer chose its line format from the `CI` env var. Anything non-TTY and non-CI — a pipe, a log file, an agent capturing stderr — took the interactive `\r` path, and since a pipe does not process `\r`, every chunk accumulated rather than redrawing in place. That branch was also unthrottled; the 30s interval only applied when `is_ci` was set. A cold `aspect --help` with stderr piped produced 57,178 bytes across 2 lines; it now produces 107. Interactive and CI behavior are unchanged by construction: `redraw_in_place` is `is_terminal() && !is_ci`, so a TTY takes the branch `!is_ci` used to take and CI takes the branch `is_ci` used to take. Only the non-TTY non-CI population moves, onto the throttled path.

**Task summaries.** `build`, `format`, `gazelle`, `lint` and `test` had no `summary`, so `aspect --help` rendered the `"<name> task defined in @aspect//<file>.axl"` fallback for five of nine tasks. Their per-flag help is already detailed, which left the tree descriptive one level down and empty at the top — the level a reader uses to choose a task at all.

**`aspect describe`.** The resolved surface as JSON: every reachable task with a copy-pasteable `command` string, group path, summary, defining module, and every flag with type, default, allowed values and description, plus the feature flags accepted everywhere. Flag strings are produced by `long_flag`, the same helper that builds the real Clap arg, so what is described and what parses cannot drift. `config.axl` overrides are reported as the effective default — re-typed rather than stringified, so an int override reads `2` and not `["2"]` — flagged with `default_from_config`, so a reader sees what will actually happen in this repo rather than the schema value it can never observe. It defaults to JSON because requiring a flag to be known up front defeats the purpose.

**`aspect auth status --output=json`.** The other half: why a thing is broken. `{account, deployments, default_deployment}`, each entry carrying `logged_in`, `status`, `identity`, its endpoints and, most usefully, the `login_command` that re-authenticates it — so a caller that finds an expired token gets the exact remedy without parsing `●`/`○` glyphs out of prose. stdout is clean JSON; the task header stays on stderr. Text output is unchanged.

`--output` is the convention for both. `aspect cache diff` already shipped `--format`, so it gains `--output` as the documented spelling and keeps `--format` working as a deprecated alias that warns.

**Color.** Clap strips the escapes baked into our help templates when rendering off a terminal, but hand-rendered output has no such protection, so `aspect feature` emitted raw SGR sequences into a pipe. Adds `color::stdout_supports_color()` (TTY, and `NO_COLOR` honored) following the pure-core pattern `banner.rs` already uses, plus `banner::line_colored`.

Tracked in ENG-1975, which also carries the follow-ups deliberately left out here: distinct exit codes (breaking — needs a decision on additive-only codes first), generated `AGENTS.md` and an `aspect mcp` adapter over the describe data, task-group summaries (blocked on a new AXL declaration surface — groups are synthesized in Rust from namespace structure with no metadata slot), and an eval harness to keep this measurable.

---

```
- New `aspect describe` prints the resolved task, flag and feature surface as JSON — every task with a runnable command string, its flags, types, defaults and allowed values — so scripts and AI agents can discover a repo's Aspect surface in one call.
- New `aspect auth status --output=json` reports login state, endpoints and the exact `login_command` that re-authenticates each entry.
- `aspect cache diff` gains `--output` as the documented spelling of `--format`. `--format` still works and now warns.
- `aspect build`, `format`, `gazelle`, `lint` and `test` now describe themselves in `aspect --help` instead of showing a placeholder.
- The launcher no longer floods non-terminal consumers with download progress: a piped cold download dropped from ~57KB of output to ~107 bytes.
- `aspect feature` no longer emits ANSI escape codes when its output is piped, and honors `NO_COLOR`.
```

### Test plan

- Covered by existing test cases (910 AXL cases via `aspect tests axl`; launcher suite)
- New test cases added:
  - `describe_emits_runnable_commands_and_real_flag_strings` (`crates/aspect-cli/src/cmd.rs`) — command string joins the group path, `_`→`-` flag conversion matches what Clap parses, positionals carry no flag.
  - `describe_retypes_config_overrides_rather_than_stringifying_them` — int/bool overrides come back as JSON numbers/bools; genuine list args keep their array shape.
  - `line_colored_drops_the_grey_wrapper_when_uncolored` (`crates/axl-runtime/src/banner.rs`).
  - `color_only_on_a_tty_without_no_color` (`crates/axl-runtime/src/color.rs`).
- Manual testing:
  1. Cold download with stderr piped, before vs after — `ASPECT_LAUNCHER_CACHE=$(mktemp -d) aspect --help 2>err >/dev/null; wc -c err` → 57,178 bytes on the released launcher, 107 on this branch.
  2. `aspect describe | python3 -c 'import json,sys; json.load(sys.stdin)'` parses; reports 74 tasks and 12 features in this repo. Verified `--run_count` (a task that overrides its flag name) is reported exactly as it parses.
  3. `aspect auth status --output=json 2>/dev/null` is valid JSON on stdout with the task header on stderr; `aspect auth status` text output is unchanged.
  4. `aspect feature | cat -v` shows no escape sequences; on a TTY the listing is still bold/colored.

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
+++ b/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
@@ -396,13 +396,12 @@ def _impl(ctx: TaskContext) -> int:
         return 1
 
     # `--format` is the pre-`--output` spelling, kept working so existing
-    # scripts don't break. It loses to an explicit `--output`; since an
-    # unset `--output` is indistinguishable from one passed as its default,
-    # "still at the default" is what yields to it.
+    # scripts don't break. An explicitly typed `--output` always wins, even
+    # when its value happens to equal the default.
     fmt = ctx.args.output
     if ctx.args.format:
         warn(ctx.std, "`--format` is deprecated; use `--output` instead.")
-        if ctx.args.output == _OUTPUT_DEFAULT:
+        if not ctx.args.is_explicit("output"):
             fmt = ctx.args.format
     using_exec = bool(ctx.args.exec)
 
--- a/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
+++ b/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
@@ -13,7 +13,7 @@ Output contract (built for pipes):
   - stdout — the affected test labels, one per line, nothing else. Pipe it:
         aspect cache diff | xargs bazel test
   - stderr — progress, per-target reasons, the summary, and errors.
-  - --format=json — one document on stdout instead of bare lines.
+  - --output=json — one document on stdout instead of bare lines.
   - --exec="<cmd>" — run `<cmd> <labels...>` instead of writing stdout.
 
 Modes (`--mode`):
@@ -73,10 +73,11 @@ load(
     "resolve_flags",
     "resolve_startup_flags",
 )
-load("./private/lib/environment.axl", "color_enabled")
+load("./private/lib/environment.axl", "color_enabled", "warn")
 load("./private/lib/remote_executor.axl", "start_dummy_executor", "stop_dummy_executor")
 load("./private/lib/runnable.axl", "apparent_label")
 
+_OUTPUT_DEFAULT = "lines"
 _GRPC_GET_ACTION_RESULT = "build.bazel.remote.execution.v2.ActionCache/GetActionResult"
 _TEST_MNEMONIC = "TestRunner"
 
@@ -394,7 +395,15 @@ def _impl(ctx: TaskContext) -> int:
     if entries == None:
         return 1
 
-    fmt = ctx.args.format
+    # `--format` is the pre-`--output` spelling, kept working so existing
+    # scripts don't break. It loses to an explicit `--output`; since an
+    # unset `--output` is indistinguishable from one passed as its default,
+    # "still at the default" is what yields to it.
+    fmt = ctx.args.output
+    if ctx.args.format:
+        warn(ctx.std, "`--format` is deprecated; use `--output` instead.")
+        if ctx.args.output == _OUTPUT_DEFAULT:
+            fmt = ctx.args.format
     using_exec = bool(ctx.args.exec)
 
     # stdout carries the result: stream bare labels only for `lines` (and not
@@ -484,8 +493,8 @@ diff = task(
                 "genuine input change)."
             ),
         ),
-        "format": args.string(
-            default = "lines",
+        "output": args.string(
+            default = _OUTPUT_DEFAULT,
             values = ["lines", "json"],
             description = (
                 "stdout format. `lines`: affected labels, one per line (pipeable; " +
@@ -494,12 +503,20 @@ diff = task(
                 "[{target, mnemonic}]}]}` for scripting (parse with jq)."
             ),
         ),
+        "format": args.string(
+            default = "",
+            values = ["", "lines", "json"],
+            description = (
+                "Deprecated alias for --output; still accepted, and warns when " +
+                "set. Use --output instead."
+            ),
+        ),
         "exec": args.string(
             default = "",
             description = (
                 "Run this command with the affected labels appended as arguments " +
                 "(e.g. --exec=\"bazel test\") instead of writing stdout. No-op when " +
-                "nothing is affected. Overrides --format. For anything fancier, omit " +
+                "nothing is affected. Overrides --output. For anything fancier, omit " +
                 "it and pipe stdout."
             ),
         ),
--- a/crates/aspect-cli/src/builtins/aspect/test.axl
+++ b/crates/aspect-cli/src/builtins/aspect/test.axl
@@ -101,6 +101,7 @@ def _impl(ctx: TaskContext) -> int | TaskConclusion:
     return conclusion
 
 test = task(
+    summary = "Run Bazel tests. Wraps `bazel test` with retries, remote cache/BES wiring, and CI status reporting.",
     args = {
         # TODO: Support - (list from stdin)
         "targets": args.positional(
--- a/crates/aspect-cli/src/cmd.rs
+++ b/crates/aspect-cli/src/cmd.rs
@@ -19,6 +19,7 @@ use std::process::ExitCode;
 
 use aspect_telemetry::cargo_pkg_display_version;
 use axl_runtime::banner;
+use axl_runtime::color;
 use axl_runtime::diag;
 use axl_runtime::engine::arg::Arg;
 use axl_runtime::engine::arguments::Arguments;
@@ -262,7 +263,10 @@ impl<'a, 'v> Cmd<'a, 'v> {
 
         match name {
             None => {
-                print!("{}", render_feature_list(version, &blocks));
+                print!(
+                    "{}",
+                    render_feature_list(version, &blocks, color::stdout_supports_color())
+                );
                 ExitCode::SUCCESS
             }
             Some(name) => match blocks.iter().find(|(f, _)| f.name() == name) {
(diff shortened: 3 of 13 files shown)


crates/aspect-cli/src/cmd.rs
```diff
@@ -752,22 +804,247 @@ fn feature_block(feat: &dyn FeatureLike<'_>) -> Option<FeatureBlock> {
...
+
+    let described_features: Vec<serde_json::Value> = features
+        .iter()
+        .filter_map(|feat| {
+            let block = feature_block(*feat)?;
+            let flags: Vec<serde_json::Value> = block
+                .args
+                .iter()
+                .map(|(arg_name, arg)| describe_arg(Scope::Feature(&block.prefix), arg_name, arg))
+                .collect();
```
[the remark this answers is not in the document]
jeffpignataro: Confirmed and fixed in ba8002f2. Verified against this repo: `expand_phases` was reported as `[]` while `.aspect/config.axl:222` sets `["test"]`. Overrides are now re-applied for feature args exactly as task args already did, with the `default_from_config` marker.

Worth noting the root cause is broader than describe: `feature_block`'s override baking has a `_ => {}` arm that skips all list-valued types, so the gap exists for any consumer of that schema. Re-applying in the describe path (rather than changing `feature_block`) keeps `--help` output untouched in this PR; the `feature_block` gap is worth a follow-up.

crates/aspect-cli/src/cmd.rs
```diff
@@ -752,22 +804,247 @@ fn feature_block(feat: &dyn FeatureLike<'_>) -> Option<FeatureBlock> {
...
+        ),
+        // Filtered out by `is_cli_exposed` before reaching here.
+        Arg::Custom { description, .. } => ("custom", json!(null), json!(null), description, &None),
+    };
+    json!({
+        "name": name,
+        "type": kind,
+        "flag": flag,
+        "short": short.as_ref().map(|s| format!("-{s}")),
+        "required": arg.is_required(),
```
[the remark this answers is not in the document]
jeffpignataro: Confirmed and fixed in ba8002f2 — this was the most valuable of the four, since it directly undermines the property this command is supposed to guarantee. Requiredness is now derived from the cardinality and `minimum`/`maximum` are reported; `aspect run` describes as `required: true, minimum: 1, maximum: 1`. Covered by `describe_reports_positional_cardinality_and_derives_requiredness`.

One correction to the diagnosis: Clap does *not* reject `aspect run` with no target. It parses fine and the task implementation then fails with a Starlark `Index `0` is out of bound` traceback. So the describe output was wrong, and separately the empty-target path has no clean error — tracking that as its own follow-up.

crates/aspect-cli/src/cmd.rs
```diff
@@ -752,22 +804,247 @@ fn feature_block(feat: &dyn FeatureLike<'_>) -> Option<FeatureBlock> {
...
+/// benefit. Handing that back verbatim would describe an int default as
+/// `["2"]`, so scalars collapse to a single value parsed back to the arg's own
+/// type and only genuine list args stay arrays.
+fn override_as_default(arg: &Arg, over: &[String]) -> serde_json::Value {
+    use serde_json::json;
+
+    match arg {
+        Arg::StringList { .. } => json!(over),
+        Arg::BooleanList { .. } => json!(over.iter().map(|v| v == "true").collect::<Vec<_>>()),
+        Arg::IntList { .. } | Arg::UIntList { .. } | Arg::Positional { .. } => json!(over),
```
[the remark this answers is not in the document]
jeffpignataro: Confirmed and fixed in ba8002f2. `int_list`/`uint_list` overrides are now re-typed elementwise to match the scalar arms. Positionals stay strings — target patterns genuinely are — so they moved to the `StringList` arm rather than staying grouped with the numeric ones. Covered by an added case in `describe_retypes_config_overrides_rather_than_stringifying_them`.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
```diff
@@ -394,7 +395,15 @@ def _impl(ctx: TaskContext) -> int:
...
-    fmt = ctx.args.format
+    # `--format` is the pre-`--output` spelling, kept working so existing
+    # scripts don't break. It loses to an explicit `--output`; since an
+    # unset `--output` is indistinguishable from one passed as its default,
+    # "still at the default" is what yields to it.
+    fmt = ctx.args.output
+    if ctx.args.format:
+        warn(ctx.std, "`--format` is deprecated; use `--output` instead.")
+        if ctx.args.output == _OUTPUT_DEFAULT:
+            fmt = ctx.args.format
```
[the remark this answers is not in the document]
jeffpignataro: Confirmed and fixed in ba8002f2, using exactly the suggested API. I had knowingly approximated "not passed" with "value equals the default" and left a comment admitting it — but `ctx.args.is_explicit` exists and is already used in `bazel_runner.axl:291`, so there was no reason to approximate. `--output=lines --format=json` now correctly yields lines.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
```diff
@@ -13,7 +13,7 @@ Output contract (built for pipes):
   - stdout — the affected test labels, one per line, nothing else. Pipe it:
         aspect cache diff | xargs bazel test
   - stderr — progress, per-target reasons, the summary, and errors.
-  - --format=json — one document on stdout instead of bare lines.
+  - --output=json — one document on stdout instead of bare lines.
```
jbedard, line 16: This seems like a breaking change, are we ok with that?

crates/aspect-cli/src/builtins/aspect/test.axl
```diff
@@ -101,6 +101,7 @@ def _impl(ctx: TaskContext) -> int | TaskConclusion:
     return conclusion
 
 test = task(
+    summary = "Run Bazel tests. Wraps `bazel test` with retries, remote cache/BES wiring, and CI status reporting.",
```
jbedard, line 104: Can we throw an error if `task()` isn't passed a summary? Or make it a required param or whatever...?

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
jeffpignataro: It's not a breaking change. There's an alias for `--format` to maintain backwards compatibility

jbedard: Actually I see you kept support for the old one so maybe we're good and can delete `--format` later?

crates/aspect-cli/src/builtins/aspect/test.axl
jeffpignataro: Oh I like that. Good idea.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
jbedard: Can we think of a better name then `--output` or is this a standard you're adopting from somewhere?

"output" is very generic, even just `--output-type` or `--output-format`?

jeffpignataro: Yeah `--format` is deprecated and we can plan a removal in the future, as we see fit. Personally I prefer the more generic verbiage. It matches other common tools https://docs.aws.amazon.com/cli/v1/userguide/cli-usage-output-format.html and we can more flexibility if we want to use at as a format or a type (e.g. `--output=csv` or `--output=pdf` versus json, yaml, etc).

crates/aspect-cli/src/builtins/aspect/test.axl
jeffpignataro: Since an error could create a breaking change, I'm going to do a `warn` currently and we can plan an `error` in a future major release with proper visibility.

thesayyn (approved):

jbedard:
This PR seems to be doing many unrelated things? Can we get things like the formatting, new commands, command renaming, attr renaming etc at least in isolated commits? Why put them all in one PR?

jeffpignataro:
I wasn't ready to merge that. I'll create another PR for the `task()` summary warn/error.

jeffpignataro:
> This PR seems to be doing many unrelated things? Can we get things like the formatting, new commands, command renaming, attr renaming etc at least in isolated commits? Why put them all in one PR?

It's all generally under the concept of making the CLI easier to use in `non-terminal` + `non-CI` environments (e.g. LLM tool calls). That said it's too late now...

PR #1378, opened 2026-08-07
feat(cli): warn when a task declares no summary, and document every built-in

jeffpignataro:
Follow-up to #1377. A task declared without a `summary` renders as `"<kind> task defined in <file>"` in `aspect --help` — the level at which a reader chooses a task at all — so an undocumented task is effectively invisible. #1377 fixed the five built-ins that showed that placeholder; this makes the rule stick.

`summary` is slated to become a required argument to `task()`. Making it required today would be a hard break: any customer task lacking one would fail to *load*, so the CLI would refuse to start in that repo, with no migration window. This is the deprecation warning ahead of that, plus the cleanup that makes the warning credible.

**Warning.** `Cmd::build` now warns per task lacking a summary, naming the kind and the defining file:

```
WARNING: task "deploy" in .aspect/deploy.axl has no summary, so it shows as
"deploy task defined in .aspect/deploy.axl" in `aspect --help`. Add
summary = "…" to task(...) — this becomes an error in a future release.
```

It is scoped to tasks defined in the repo being run. A module's tasks come back from `defined_in_label` as `@mod//…`, and warning about a third-party module's task would be noise the reader cannot act on. That scoping is also why the check lives in `Cmd::build` rather than in the `task()` builtin — `task()` knows the defining path but not the module list, so it cannot tell a first-party task from a dependency's.

**Built-ins.** Filling every gap first is what keeps the warning credible; one that fires 43 times per invocation just teaches people to ignore warnings. Before this change 51 of 74 tasks had no summary. `aspect github token` gained a real one, the 42 internal `aspect dev test-*` tasks gained generated ones derived from their kind ("Run the tips AXL unit tests."), and this repo's own four root tasks are documented, so a local `aspect` run is now silent.

**Guard.** Because the runtime warning deliberately stays quiet for `@aspect//` tasks, nothing at runtime holds our own built-ins to the rule. `every_builtin_task_declares_a_summary` does, scanning the built-in source tree so a regression fails CI and names the `file:line` a contributor just wrote rather than surfacing in a customer's terminal.

Tracked in ENG-1975.

---

```
- `aspect` now warns when a task in your repo is declared without a `summary`, naming the task and the file. A task without one shows as "<kind> task defined in <file>" in `aspect --help`. `summary` will become required in a future release — adding it now is forward-compatible.
- Every built-in task now has a summary, including `aspect github token` and the internal `aspect dev test-*` tasks.
```

### Test plan

- Covered by existing test cases (910 AXL cases via `aspect tests axl`; the `aspect-cli` unit suite)
- New test cases added:
  - `every_builtin_task_declares_a_summary` — scans the built-in source tree for a `task(` call with no `summary`, failing with the offending `file:line`.
- Manual testing:
  1. `aspect --help` in a repo with an undocumented task emits one warning per task, on stderr, naming the kind and file. Verified against this repo's four root tasks before documenting them.
  2. Tasks from `@dev//` and `@demo//` modules do **not** warn, confirming the first-party scoping.
  3. After documenting the four root tasks, `aspect --help 2>&1 >/dev/null | grep -c WARNING` returns 0.
  4. The guard was verified to fail, not just pass: removing a single `summary` makes it report `private/lib/tips_test.axl:738`. It caught a real regression during this change, when a `git checkout` reverted one of the insertions.

### DIFF
--- a/crates/aspect-cli/src/cmd.rs
+++ b/crates/aspect-cli/src/cmd.rs
@@ -1881,32 +1881,26 @@ mod tests {
     }
 
     /// The runtime warning skips `@aspect//` tasks, so enforce the rule here.
+    ///
+    /// Reads the `include_dir!` embed, not the filesystem, so it works under
+    /// Bazel's sandbox where the cargo layout isn't on disk.
     #[test]
     fn every_builtin_task_declares_a_summary() {
-        let root = Path::new(env!("CARGO_MANIFEST_DIR")).join("src/builtins/aspect");
-        let mut offenders = Vec::new();
+        use include_dir::{Dir, include_dir};
+        static ASPECT_DIR: Dir<'_> = include_dir!("$CARGO_MANIFEST_DIR/src/builtins/aspect");
 
-        let mut stack = vec![root.clone()];
-        while let Some(dir) = stack.pop() {
-            for entry in std::fs::read_dir(&dir).expect("readable builtin dir") {
-                let path = entry.expect("readable entry").path();
-                if path.is_dir() {
-                    stack.push(path);
-                    continue;
-                }
-                if path.extension().is_none_or(|e| e != "axl") {
-                    continue;
-                }
-                let text = std::fs::read_to_string(&path).expect("readable .axl");
-                // `<var> = task(` at column 0, keywords at one indent.
-                for (idx, _) in text.match_indices(" = task(\n") {
-                    let body = &text[idx..];
-                    let end = body.find("\n)").unwrap_or(body.len());
-                    if !body[..end].contains("\n    summary = ") {
-                        let line = text[..idx].lines().count() + 1;
-                        let rel = path.strip_prefix(&root).unwrap_or(&path);
-                        offenders.push(format!("{}:{}", rel.display(), line));
-                    }
+        let mut offenders = Vec::new();
+        for f in ASPECT_DIR.find("**/*.axl").unwrap() {
+            let Some(text) = f.as_file().and_then(|file| file.contents_utf8()) else {
+                continue;
+            };
+            // `<var> = task(` at column 0, keywords at one indent.
+            for (idx, _) in text.match_indices(" = task(\n") {
+                let body = &text[idx..];
+                let end = body.find("\n)").unwrap_or(body.len());
+                if !body[..end].contains("\n    summary = ") {
+                    let line = text[..idx].lines().count() + 1;
+                    offenders.push(format!("{}:{}", f.path().display(), line));
                 }
             }
         }
--- a/crates/aspect-cli/src/cmd.rs
+++ b/crates/aspect-cli/src/cmd.rs
@@ -1153,8 +1153,6 @@ fn task_command(
         kind, label,
     );
     let about = if task.summary().is_empty() {
-        // Warn beside the fallback rather than at the call site, so the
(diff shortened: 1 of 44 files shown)


crates/aspect-cli/src/cmd.rs
```diff
@@ -104,6 +104,7 @@ impl<'a, 'v> Cmd<'a, 'v> {
         let mut tree = Tree::default();
         for (idx, task) in self.tasks.iter().enumerate() {
             let label = defined_in_label(task.path(), self.aspect_root, self.modules);
+            warn_missing_summary(*task, &label);
```
jbedard: Shouldn't this go into `task_command()` or are we distinguishing tasks created via this path from others? (I don't know this code so take this with a grain of salt 🤷)

crates/aspect-cli/src/cmd.rs
```diff
@@ -1356,6 +1357,31 @@ fn group_section(group_names: &[String]) -> String {
...
 // ── Misc helpers ───────────────────────────────────────────────────────────
 
+/// Warn once for a task declared without a `summary`.
+///
+/// A task with no summary renders as `"<kind> task defined in <file>"` in
+/// `aspect --help` — the level at which a reader chooses a task at all, so an
+/// undocumented task is effectively invisible. This is a deprecation warning:
+/// `summary` is slated to become required.
+///
+/// Scoped to tasks defined in this repo, which [`defined_in_label`] reports as
```
jbedard: Docs for a method referencing arbitrary other methods like `defined_in_label` and referencing the test methods (`every_builtin_task_declares_a_summary`) seems like a lot of AI-slop spam. Claude needs to stfu in this repo.

The docs should say what it does, not tell the life story and history of everything related to it except in exceptional circumstances (which should not exist for a basic warning method 🤦).

crates/aspect-cli/src/cmd.rs
jeffpignataro: Good catch — moved in 46fa764a.

To answer the question directly: no, nothing is being distinguished. `task_command()` has exactly one call site (this loop), so there was no second path and the placement was not load-bearing.

Your instinct is right for a reason I had missed. `task_command` is where the fallback is actually *applied*:

```rust
let about = if task.summary().is_empty() {
    context_line.clone()      // "<kind> task defined in <label>"
} else {
    task.summary().clone()
};
```

The warning quotes that exact string back to the user. With the two a thousand lines apart, changing the fallback text would silently falsify the warning. They now sit together, and since `task_command` already receives `label` the move is behavior-preserving.

Checking this also turned up something you did not ask about but which your comment led me to: every `Cmd::build` unit test uses `stub_task`, whose empty summary and non-module path trip the warning — so the suite was emitting **19 spurious WARNING lines**. That is exactly the "warnings people learn to ignore" failure this PR argues against, showing up in our own test output. `StubTask` now carries a summary, and the rule those tests were incidentally exercising is covered on purpose instead: `needs_summary_warning` is split out as a pure predicate (the pattern `banner.rs` already uses) and tested over all four cases.

So: thanks, the comment was worth more than the line it was on.

crates/aspect-cli/src/cmd.rs
jeffpignataro: Claude agreed and tidied things up.

jbedard (approved):

PR #1385, opened 2026-08-11
feat(cli): make task groups and describe navigable

jeffpignataro:
Three fixes to how the CLI surface is discovered, measured with a clean-room agent asked to find the affected-tests command with no prior knowledge of the tool. Before: 7 commands to the answer, one of them brute-forcing `--help` across all 10 task groups. After: 3 commands, 66% fewer bytes to get there.

**Task groups list their members.** `cache task group` said nothing about whether the command you want is inside, so the agent iterated every group to find out. Now:

```
Task Groups:
  auth     configure, login, logout, remove, status, use
  cache    diff
  wrapper  install, uninstall
```

Derived from the command tree rather than a new AXL field, so every repo's groups get it with no opt-in.

**`aspect describe` is tiered.** It was 185KB — too expensive to be the first call a reader makes, which defeats its purpose. It now prints a ~10KB index of commands and summaries; `aspect describe '<command>'` gives one task's full flags (~6KB). Discover-then-drill is ~16KB instead of 185KB.

**`aspect test --help` points at `aspect cache diff`**, which is what the re-run agent named as answering its question. The feature footer now says what features are (CI reporting and artifact integrations) instead of only that flags are hidden — the old wording sent the first agent hunting for test selection among them.

ENG-1975.

---

```
- `aspect --help` now lists what each task group contains instead of repeating the group's name.
- `aspect describe` now prints a compact index of commands and summaries; use `aspect describe '<command>'` for one task's full flags. The previous full-surface dump was ~185KB.
- `aspect test --help` cross-references `aspect cache diff` for running only affected tests.
```

### Test plan

- Covered by existing test cases (68 `aspect-cli` unit tests; 910 AXL cases via `aspect tests axl`)
- Manual testing:
  1. `aspect --help` shows member lists per group; long groups elide as `… (+41 more)`.
  2. `aspect describe` = 10,206 bytes; `aspect describe 'cache diff'` = 6,311 bytes; both valid JSON.
  3. `aspect describe 'nope'` errors and exits non-zero; unquoted `aspect describe cache diff` errors on the extra argument.
  4. Clean-room agent re-run reached the answer at command 3 vs command 7 previously.

### DIFF
--- a/crates/aspect-cli/src/cmd.rs
+++ b/crates/aspect-cli/src/cmd.rs
@@ -984,10 +984,6 @@ fn override_as_default(arg: &Arg, over: &[String]) -> serde_json::Value {
 }
 
 /// Every runnable command with its summary, and nothing else.
-///
-/// This is the default because the full surface is ~185KB — too expensive to
-/// be the first thing a reader fetches. The index is ~9KB and answers "what
-/// can I run here"; `aspect describe <task>` answers "how do I run it".
 fn describe_index(
     version: &str,
     tasks: &[&dyn TaskLike<'_>],
--- a/crates/aspect-cli/src/cmd.rs
+++ b/crates/aspect-cli/src/cmd.rs
@@ -108,7 +108,7 @@ impl<'a, 'v> Cmd<'a, 'v> {
             tree.insert(*task, &label, task_cmd)?;
         }
 
-        let group_section = group_section(&tree.group_names());
+        let group_section = group_section(&tree.group_rows());
         let root = Command::new("aspect")
             .bin_name("aspect")
             .about("Aspect's programmable task runner built on top of Bazel\n{ Correct, Fast, Usable } -- Choose three")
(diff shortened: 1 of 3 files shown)


crates/aspect-cli/src/cmd.rs
```diff
@@ -963,6 +983,52 @@ fn override_as_default(arg: &Arg, over: &[String]) -> serde_json::Value {
     }
 }
 
+/// Every runnable command with its summary, and nothing else.
+///
+/// This is the default because the full surface is ~185KB — too expensive to
```
jbedard: Outputting things like ~185k vs ~9kb sounds like slop that will be out of date within a week. I think half this second sentence can be dropped?

jbedard (approved):

crates/aspect-cli/src/cmd.rs
jeffpignataro: Yeah, definitely fair. Point-in-time comments seem dumb.

PR #1388, opened 2026-08-13
fix(cli): honor a caller-forwarded --target_pattern_file

mcook42:
Split out of #1370 (fix 1 of 3). See that PR's description for full context on all three defects found in the same investigation.

A caller-forwarded `--target_pattern_file` conflicted with the default target pattern. `run_bazel_task` only suppressed its own `--target-pattern-file` setting, but Bazel's `--bazel-flag=--target_pattern_file=<path>` — what `tools/bazel` produces for every Bazel-native flag, and so what an IntelliJ BSP sync sends — was invisible to the runner. `ctx.args.targets` then fell back to its declared default (`["..."]`), which reaches Bazel as residue alongside the forwarded flag:

```
ERROR: Command-line target pattern and --target_pattern_file cannot both be specified
```

That pair has no last-wins rule in Bazel. It's a hard conflict, so the only decision `run_bazel_task` owns is whether to invent a pattern the user never typed. Every precedence question belongs to Bazel's own option parser. `target_patterns()` now asks the effective `RunCommand` (`rc.flag_value("--target_pattern_file", ...)`) whether a pattern file reached Bazel by any spelling, using the same last-wins / `=`-form / two-token matching as `crates/bazelrc`.

### Test plan

- New test cases — `target_patterns` in `bazel_runner_test.axl`: default forwarded when nothing supplies patterns; suppressed by a forwarded `--bazel-flag`; suppressed by aspect's own `--target-pattern-file`; explicit patterns always forwarded; the lookup is command-scoped. Verified locally: `bazel run //crates/aspect-cli:aspect-cli -- dev test-bazel-runner` → `bazel_runner_test.axl: OK (10 sections)`.
- New CI coverage — extends `test-flags-task` in both `.buildkite/pipeline.yaml` and `.github/workflows/ci-workflows.yaml` with one `aspect build` in the plugin's exact shape (`--bazel-flag=--target_pattern_file=...`), asserting the composed command line is accepted by Bazel.
  - This is the `--target_pattern_file` half of #1370's combined IDE/BSP smoke-test commit — the `--build_event_binary_file` half went to that fix's own PR instead, since a single combined step would make each split PR's CI depend on the other's fix also being merged.

### Before / after (captured on `main`)

| Check | Before | After |
|---|---|---|
| Plugin-shaped invocation | `cannot both be specified`, exit 2 | exit 0 |
| Two-token `--target_pattern_file <path>` | `cannot both be specified` | exit 0 |
| Pattern file from a `.bazelrc` line | default pattern collided | exit 0 — now suppresses the default too |

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/private/lib/bazel_runner.axl
+++ b/crates/aspect-cli/src/builtins/aspect/private/lib/bazel_runner.axl
@@ -218,6 +218,37 @@ def _emit_terminal(ctx, lifecycle, command, data, exit_code, targets = []):
         flagged = final_status == "warning",
     )
 
+def target_patterns(explicit: bool, cli_targets: list[str], own_pattern_file: str, rc, command: str) -> list[str]:
+    """The target patterns to place on Bazel's command line.
+
+    Bazel rejects an invocation carrying both command-line patterns and
+    `--target_pattern_file` — that pair has no last-wins rule — so the single
+    decision here is whether to materialize the `targets` arg's declared
+    default (`["..."]`), a pattern the user never typed. Explicit patterns
+    always pass through: when they conflict with a pattern file, Bazel reports
+    it, and this runner does not grow a second spelling of that error.
+
+    A pattern file counts no matter which spelling delivered it: aspect's own
+    `--target-pattern-file` arrives as `own_pattern_file`, while Bazel's
+    `--target_pattern_file` reaches `rc` — from `--bazel-flag`, `.bazelrc`, or
+    a `--config` expansion alike, resolved with Bazel's own last-wins parsing.
+
+    Args:
+        explicit:         whether the user actually passed target patterns
+                          (`ctx.args.is_explicit("targets")`) rather than
+                          inheriting the arg's default.
+        cli_targets:      `ctx.args.targets` — never empty (the arg declares
+                          `default = ["..."]`, `minimum = 1`).
+        own_pattern_file: value of aspect's `--target-pattern-file`, or `""`.
+        rc:               the active `RunCommand`; only `flag_value` is used.
+        command:          Bazel subcommand whose options `rc` resolves.
+    """
+    if explicit:
+        return cli_targets
+    if own_pattern_file or rc.flag_value("--target_pattern_file", command = command):
+        return []
+    return cli_targets
+
 def run_bazel_task(ctx: TaskContext, command: str, targets = None) -> TaskConclusion:
     """Shared impl for build/test tasks.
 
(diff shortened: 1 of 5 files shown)


crates/aspect-cli/src/builtins/aspect/private/lib/bazel_runner.axl
```diff
@@ -218,6 +218,37 @@ def _emit_terminal(ctx, lifecycle, command, data, exit_code, targets = []):
         flagged = final_status == "warning",
     )
 
+def target_patterns(explicit: bool, cli_targets: list[str], own_pattern_file: str, rc, command: str) -> list[str]:
```
jbedard, line 221: Can `rc` have a type? I have no idea what it is...

jbedard (approved):

crates/aspect-cli/src/builtins/aspect/private/lib/bazel_runner.axl
mcook42: I traced it back to `lifecycle.axl.setup_phase` where it's defined as https://github.com/aspect-build/aspect-cli/blob/031564d5d9172e46a0eccb97984f806006cc0253/crates/aspect-cli/src/builtins/aspect/private/lib/lifecycle.axl#L875-L877

There are a few other places where the `rc` value is passed around, so to type this for real I'd probably need to follow up with another PR to correctly type the `rc` value throughout the call chain

jbedard: Maybe you can do that in a followup if the context is in your (or claudes) brain atm?

crates/aspect-cli/src/builtins/aspect/private/lib/bazel_runner.axl
```diff
@@ -299,6 +326,26 @@ def run_bazel_task(ctx: TaskContext, command: str, targets = None) -> TaskConclu
...
+    if targets == None:
+        # aspect's own `--target-pattern-file` (declared by build/test only) is a
+        # per-invocation flag, not rc material — and keeping it out of the parsed
+        # rc is what lets `target_patterns` ask `rc` about the forwarded
+        # spelling. Forwarded rather than expanded into argv, since the flag
+        # exists to bypass OS command-line length limits.
+        own_pattern_file = getattr(ctx.args, "target_pattern_file", "") if hasattr(ctx.args, "target_pattern_file") else ""
+        if own_pattern_file:
+            if not ctx.std.fs.exists(own_pattern_file):
+                fail("--target-pattern-file: file not found: " + own_pattern_file)
```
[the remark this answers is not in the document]
jbedard: @mcook42 please read over this and debate if it's worth thinking about

PR #1396, opened 2026-08-22
feat(cli): type Bazel flags directly on aspect commands

gregmagolan:
`aspect build --remote --deployment=aspect-cloud-prod --remote_download_all //...` failed with `error: unexpected argument '--remote_download_all' found`. Every plain Bazel flag had to be re-spelled as `--bazel-flag=--remote_download_all`.

Bazel flags can now be typed directly. A task opts in by declaring `args.passthrough(position = ...)`; before argv reaches clap, the tokens clap would reject are rewritten onto that bucket, so the parse succeeds and the task reads them off `ctx.args`. Which bucket a flag lands in follows Bazel's own split — before the task name it becomes a startup option, after it a command option — so each set is forwarded to the slot it was typed for.

```
aspect build --remote_download_all //...     →  bazel build --remote_download_all -- //...
aspect --output_base=/tmp/o build //...      →  bazel --output_base=/tmp/o build -- //...
aspect build -c opt --config ci //...        →  bazel build -c opt <ci expansion> -- //...
aspect --output_base /tmp/o build //...      →  bazel --output_base /tmp/o build -- //...
aspect init --remote_download_all            →  error: unexpected argument (declares no bucket)
```

Three properties hold it together:

- **Forwarding is a claim the runtime enforces, before the work starts.** Reading a bucket means claiming it (`ctx.args.claim(name)`). Every `ctx.bazel` call that reaches a server — build, test, query, info, version, shutdown, health_check, recover_poisoned_sandbox, cancel_invocation — refuses to run while a bucket holds unclaimed flags — naming them and the line about to spawn — so a dropped flag costs milliseconds instead of a build; a check after the task returns is the backstop for a task that spawns nothing. A plain attribute read deliberately does not claim, so a task can inspect before deciding.
- **"Recognized" means what clap accepts, in that slot.** The router keeps no list of flag names — it reads the built clap surface. Since clap propagates only globals into a subcommand, a root-only flag like `--version` is recognized before the task name and forwarded after it. A short token counts as recognized only when every character of the cluster does, so `-vv` forwards while a declared `-bn` does not.
- **Arity mirrors Bazel, never exceeds it.** A collected flag claims the following token only for flags Bazel itself takes a separate value for, from lists generated out of `bazel help` — one for command flags, one for startup options, each the union across the latest patch of Bazel 6, 7, 8 and 9 plus the newest 10 pre-release, so a repo pinned to an older Bazel gets the spellings its Bazel accepts and a repo on a newer one is covered before the release lands. A flag whose arity *disagrees* across those versions is dropped rather than guessed at. Label-shaped flags (`--//pkg:flag`) are excluded because Bazel requires `=` for them — accepting a space there would make the `bazel …` repro commands the CLI prints unrunnable.

`tools/bazel` gets to stop caring about flags: it embedded four generated Bazel flag lists (~350 lines) and a classifier purely to turn `--keep_going` into `--bazel-flag=--keep_going` before exec'ing aspect. Every argument is now forwarded verbatim — **972 lines to 470** — and its 87-case test suite asserts pass-through instead of rewriting. One 16-entry list stays, for argv tokenization rather than flags: verb detection has to know that `bazel --bazelrc build query …` means the *query* command, not `build`. The wrapper's schema version goes to 2, since it needs a CLI with passthrough; version 1 remains the answer for an older CLI, and `--bazel-flag=` keeps working either way.

Includes the `--config` fix this depends on: the rc expander matched only `--config=NAME`, so the space form reached Bazel unexpanded and — since a `RunCommand` adds `--ignore_all_rc_files` — came back as `Config value 'ci' is not defined in any .rc file` while sitting in the user's `.bazelrc`. That is broken on main today for `--bazel-flag=--config --bazel-flag=ci` and for an rc line spelled `build --config ci`.

## Before

```mermaid
flowchart TD
    A["argv"] --> B["launcher: consumes -v/--version,<br/>forwards the rest verbatim"]
    B --> C{"argv[1] == 'get'?"}
    C -- yes --> D["credential helper, exits"]
    C -- no --> E["Cmd::build → clap Command tree<br/>(root globals + task subcommands + hidden feature args)"]
    E --> F["clap parses argv"]
    F --> G{"every token recognized?"}
    G -- no --> H["error: unexpected argument '--remote_download_all'<br/>tip: a similar argument exists: '--remote'"]
    G -- yes --> I["Dispatch → ctx.args"]
    I --> J["task impl → bzl.flags.resolve*<br/>(--bazel-flag / --bazel-startup-flag only)"]
    J --> K["bazel"]
    style H fill:#a33,color:#fff
```

Every Bazel flag had one way in: the `--bazel-flag=` / `--bazel-startup-flag=` wrappers at step J.

## After

```mermaid
flowchart TD
    A["argv"] --> B["launcher: unchanged"]
    B --> C{"argv[1] == 'get'?"}
    C -- yes --> D["credential helper, exits"]
    C -- no --> E["Cmd::build → clap Command tree"]
    E --> E2["root_cmd.build()<br/>finalize: clap injects --help/-h"]
    E2 --> R["Cmd::route_unrecognized_flags"]

    subgraph ROUTE["routing pre-pass — argv in, argv out"]
      R --> R1["select_command: walk subcommand names to the leaf task,<br/>skipping each declared flag's value; collect the flags<br/>clap accepts at the leaf, and those declared anywhere"]
      R1 --> R2{"leaf task found?"}
      R2 -- no --> R9["argv unchanged"]
      R2 -- yes --> R3{"task declares<br/>a bucket for<br/>this slot?"}
      R3 -- no --> R9
      R3 -- yes --> R4["Selection::routable: scan the slot, stopping at '--'"]
      R4 --> R5{"clap accepts it<br/>in this slot?"}
      R5 -- yes --> R6["leave it; skip its value token"]
      R5 -- no --> R7{"listed in<br/>bazel_value_flags?"}
      R7 -- yes --> R7a["collect the flag<br/>+ its value token"]
      R7 -- no --> R7b["collect the flag alone"]
      R7a --> R8{"typed before<br/>the task name?"}
      R7b --> R8
      R8 -- yes --> R8a["pre_command bucket<br/>(moved to after the task name)"]
      R8 -- no --> R8b["post_command bucket<br/>(rewritten in place)"]
    end

    R9 --> F
    R6 --> F
    R8a --> F
    R8b --> F["clap parses the rewritten argv"]
    F --> G{"every token recognized?"}
    G -- no --> H["error: unexpected argument<br/>(a task with no bucket, e.g. init)"]
    G -- yes --> I["Dispatch → ctx.args"]
    I --> J["task impl runs. bzl.flags.resolve* folds each bucket in<br/>and CLAIMS it: pre → startup flags, post → command flags.<br/>--config=NAME and --config NAME both expand via the rc.<br/>A path that spawns no bazel calls disclaim_passthrough."]
    J --> P{"ctx.bazel invocation:<br/>any bucket unclaimed?"}
    P -- yes --> Q["ERROR before the spawn,<br/>naming the flags and the line"]
    P -- no --> K["bazel"]
    K --> Z{"on return, bucket<br/>still unclaimed?"}
    Z -- no --> N["task's own exit code"]
    Z -- yes --> Y{"task already<br/>failed?"}
    Y -- yes --> W["WARN, keep the task's exit code"]
    Y -- no --> M["ERROR: the flags had no effect<br/>exit 2"]
    style H fill:#a33,color:#fff
    style M fill:#a33,color:#fff
    style Q fill:#a33,color:#fff
    style W fill:#a83,color:#fff
```

A task that hard-errors (`fail()`) never reaches the claim check, so a real failure is never masked by it.

## User-facing AXL API

Nothing to write for the built-in Bazel tasks — `build`, `test`, `run`, `lint`, `format`, `gazelle`, `cache diff`, `ci warming`, `ci runner-health-check` and any `task.alias` of them get this from the shared `bzl.flags.args()` bundle, so those task files did not change. `delivery` opts in explicitly via `bzl.flags.passthrough_args()`, keeping its own `--bazel-flag` wording; a forwarded flag reaches every delivery phase, so its help points non-deterministic flags at `--release-bazel-flag`. What a **repo** may want, in `.aspect/config.axl`:

```python
load("@aspect//bazel.axl", bzl = "bazel")

def config(ctx):
    # Teach the parser that a wrapper's own flag takes a separate value, so
    # `aspect build --my_wrapper_flag value //...` forwards both tokens.
    ctx.tasks["build"].args.bazel_value_flags = bzl.flags.BAZEL_VALUE_FLAGS + ["--my_wrapper_flag"]
```

What a **task author** writes, wrapping any tool:

```python
def _impl(ctx: TaskContext) -> int:
    # claim() is the act of forwarding: it returns the collected flags and tells
    # the runtime this task is responsible for them. A plain `ctx.args.tool_flags`
    # read inspects without claiming, and a run that never claims fails.
    tool_flags = ctx.args.claim("tool_flags")
    return ctx.std.process.command("some-tool").args(tool_flags).spawn().wait().code

my_task = task(
    implementation = _impl,
    args = {
        "targets": args.positional(minimum = 1, maximum = 512),
        # Flags this task does not declare, collected instead of rejected.
        "tool_flags": args.passthrough(
            position = "post_command",        # or "pre_command", one bucket each
            value_flags_from = "tool_value_flags",
            description = "Shown as the unrecognized-flag footer in --help.",
        ),
        # Which of those take a separate value. config_only: settable from
        # config.axl, never a flag — routing reads it before clap parses.
        "tool_value_flags": args.string_list(default = ["-c", "-j"], config_only = True),
    },
)
```

## API changes

Everything new is additive; nothing was removed or renamed.

**New arg type — `args.passthrough(...)`**

| Parameter | Required | Meaning |
|---|---|---|
| `position` | yes | `"pre_command"` (typed before the task name) or `"post_command"` (after it). |
| `value_flags_from` | no | Name of a sibling `args.string_list` listing the flags that take a separate value, so `-c opt` is collected as a pair. Omitted means no token is ever claimed as a value. |
| `description` | no | Rendered as the unrecognized-flag footer on the task's `--help`, since a bucket has no flag to list. |

The value arrives on `ctx.args` as a `list[str]` in command-line order. "Before the task name" spans the whole command path, so for a grouped task `aspect --a cache --b diff --c` puts `--a` and `--b` in the pre-command bucket and `--c` in the post-command one; a group or task name is never routed, nor taken as a flag's value. Rejected at evaluation time: two buckets at the same position on one task, a bucket declared on a `feature()` (its args are injected into every task), and overriding one via `task.alias(defaults = ...)` (the schema carries no default to replace).

**New kwarg — `config_only`, on every flag-shaped kind**

`True` keeps the flag off the command line, settable only from `config.axl`. Available on `args.string`, `boolean`, `int`, `uint` and the four `_list` variants, and rejected alongside `required`, `short` or `long`, which name or demand a command-line spelling that no longer exists. The argv-shaped kinds (`positional`, `trailing_var_args`, `passthrough`) take no such parameter — their content comes from the command line — and neither does `args.custom`, which has always been config-only by construction.

`args.custom(type, default = ...)` stays the right choice for values the CLI cannot express (lambdas, complex dicts). What forced a second spelling is that it cannot carry a **list** default: a live list cannot be frozen at declaration time, so `list[str]`, `list[int]` and `list[bool]` defaults all resolve silently to `None`. For scalars the win is narrower — the declared kind stays visible (`args.int(default = 3, config_only = True)`) — but the knob is on all eight so the surface is consistent.

Worth knowing, and pre-existing: a `config.axl` override is not value-checked, so `values = [...]` has no effect on an arg only config can set. `describe` omits `config_only` args, as it always has for `args.custom`.

**New method — `ctx.args.claim(name)`**

Returns the arg's value and records that this task is responsible for it. `[]` for an arg the task does not declare, so it stays safe to call from shared helpers. A plain `ctx.args.<name>` read does not claim.

Enforced at two moments by `engine::passthrough`: every `ctx.bazel` invocation errors while a bucket holds unclaimed flags, so the failure lands before the build rather than after it; and an otherwise-successful run that returns with flags still unclaimed exits **2** with the same diagnostic. A run that already failed on its own keeps its exit code and gets the text as a warning, and a task that hard-errors never reaches either check.

**New `bzl.flags` surface** (`@aspect//bazel.axl`)

| Name | Kind | Purpose |
|---|---|---|
| `bzl.flags.BAZEL_VALUE_FLAGS` | `list[str]` | The generated command-flag arity table, re-exported so `config.axl` can extend rather than restate it. |
| `bzl.flags.BAZEL_STARTUP_VALUE_FLAGS` | `list[str]` | The same for startup options, read by the pre-command bucket. |
| `bzl.flags.disclaim_passthrough(ctx)` | function | Account for both buckets on a task path that spawns no Bazel, so a legitimate no-op is not reported as dropping flags. |
| `bzl.flags.args(phrase)` | function | Now returns five keys instead of two, adding `bazel_passthrough_flags`, `bazel_passthrough_startup_flags` and `bazel_value_flags`. Additive for anything splatting it. |
| `bzl.flags.passthrough_args(post_note = "")` | function | Just those three, for a task that declares its own `bazel_flags` / `bazel_startup_flags` with wording of its own — how `delivery` opts in. `post_note` appends a caveat to the bucket's `--help` text. |

**New loadable module** — `load("@aspect//bazel/value_flags.axl", "BAZEL_VALUE_FLAGS")`, for a `config.axl` that would rather not go through the `bzl.flags` facade.

**New config.axl-settable args** — `ctx.tasks["<task>"].args.bazel_value_flags = [...]` and `.bazel_startup_value_flags = [...]` on every Bazel-spawning task. Replacement only, not `.append()`: reading an un-overridden arg in `config.axl` is an error today, so `bzl.flags.BAZEL_VALUE_FLAGS + [...]` is the spelling.

**`aspect describe` output** — every arg object gains a `"position"` field, `null` except on a passthrough, where it is `"pre_command"`/`"post_command"`. A passthrough reports `"type": "passthrough"` and `"flag": null`, since no caller types it.

**This repo's own usage** — the CI configs (GitHub, Buildkite, GitLab, CircleCI), the builtins README and the task help now use the direct spelling: `--bazel-flag=--config=pure_go` became `--config=pure_go`. `--bazel-flag` / `--bazel-startup-flag` stay supported permanently and are documented as the way out when a Bazel flag collides with one of a task's own; one GHA smoke deliberately keeps the wrapper spelling so CI still covers it, and a new one exercises Bazel's own spellings (`-c opt`, `--jobs 8`, a space-form `--output_base`) in both slots.

**Behavior changes**

- Unrecognized flags on a task that declares a bucket are forwarded instead of failing the parse. Tasks with no bucket are unchanged (`auth *`, `axl add`, `ci bazelrc`, `ci runner-metadata`, `github token`, `init`, `wrapper *`).
- A flag only the *root* declares (`--version`, `-v`) typed after a task name is now forwarded rather than reported as `unexpected argument`, matching what clap accepts in that position. Before the task name it is still clap's.
- `--config NAME` now expands wherever `--config=NAME` did — from a `.bazelrc` line, a `--bazel-flag` pair, or a forwarded flag.

**Rust surface** (for reviewers): the new `engine::passthrough` module (the contract, both enforcement points, one diagnostic, one exit code), `Cmd::route_unrecognized_flags`, `Arg::Passthrough` + `PassthroughPosition`, `Arg::{config_only, value_flags_from, string_list_default, description, passthrough_position}`, `Arguments::{claim, is_claimed_key}`, and `EvalBuilder::with_string_list_args` for seeding `ctx.args` in runtime tests.

## The AXL behind it

`bazel/value_flags.axl` — generated source of truth, 409 command flags and 16 startup options:

```python
"""The Bazel flags that take a separate value (`-c opt`, `--jobs 8`).

GENERATED by tools/gen_bazel_value_flags.sh from `bazel help` — do not edit by
hand. Regenerate after a Bazel release.

The union across every Bazel a workspace might be pinned to, so a repo on an
older — or newer — Bazel gets the same spellings its Bazel accepts: 6.6.0, 7.7.1,
8.7.0, 9.2.0, 10.0.0-pre.20260811.3.

A flag that takes a value in one version and is boolean in another is left out —
listing it would swallow the following target pattern for anyone on the boolean
version, so those keep their `--flag=value` spelling only. Left out on those
grounds: --disk_cache, --execution_log_binary_file, --execution_log_compact_file,
--execution_log_json_file.

A few entries come from a pull request expected to land rather than from any
Bazel above; see PENDING_VALUE_FLAGS in the generator.
...
Label-shaped flags (`--//pkg:flag`) are absent on purpose: Bazel requires `=`
for those, so accepting a space-separated value would diverge from Bazel and
break the `bazel …` repro commands the CLI prints.
"""

BAZEL_VALUE_FLAGS = [
    "--action_env",
    ...
    "--compilation_mode",
    "--config",
    "--jobs",
    ...
    "-c",
    "-j",
]
```

The generator reads arity off Bazel's own rendering with awk, so the table mirrors Bazel rather than approximating it: `--jobs [-j] (an integer…)` takes a value, `--[no]announce_rc (a boolean…)` does not, and an expansion flag like `--remote_download_all` renders bare and is skipped. bazelisk resolves each spec, so no version list is hardcoded — `<major>.x` is that major's latest patch and `rolling` the newest pre-release, which is how Bazel 10 contributes 6 flags a release-only table would miss. `bazel help` runs from an empty directory: outside a workspace Bazel answers in batch mode without a server, and an older Bazel never has to read this repo's `MODULE.bazel`.

`PENDING_VALUE_FLAGS` covers the gap the other way — a flag from a PR expected to land, which no Bazel above has yet. It carries `--sandbox_backend` and `--sandbox_backend_opt` from [bazelbuild/bazel#29886](https://github.com/bazelbuild/bazel/pull/29886), each named with its PR so it can be dropped once a listed version ships it. Listing one early costs nothing: a Bazel without the flag rejects it on its own.

```bash
BAZEL_VERSIONS=(6.x 7.x 8.x 9.x rolling)

value_flags() {
    awk '
        /^  --[a-z_]+ \(/ { print $1; next }
        /^  --[a-z_]+ \[-[a-zA-Z]\] \(/ {
            print $1; short = $2; gsub(/[][]/, "", short); print short
        }
    '
}

# Value-taking in every version that has the flag at all, plus the pending ones.
printf '%s\n' "${PENDING_VALUE_FLAGS[@]}" >>"$values"
mapfile -t flags < <(sort -u "$values" | comm -23 - <(sort -u "$booleans"))
```

Two guards keep a parse slip from shipping quietly: each version must yield at least 100 value-taking flags, and every emitted entry must look like `--flag` or `-x`. The first is how the initial run caught itself invoking Bazel 6 from inside this workspace and parsing nothing.

The subtraction on that last line is no longer hypothetical, so the generator now reports what it removed rather than dropping it silently:

```
  10.0.0-pre.20260811.3: 342 value-taking command flags
  excluded (value-taking in one version, boolean in another): --disk_cache
  --execution_log_binary_file --execution_log_compact_file --execution_log_json_file
```

Bazel 10 gave those four an optional value — `--[no]disk_cache (a path, or a boolean to use the default disk cache location)` — where Bazel 6-9 required one. On Bazel 10, `aspect build --disk_cache //...` means "cache to the default location, build `//...`"; had the table listed the flag, the CLI would have paired it with `//...` and built nothing. Excluding them costs the space-separated spelling on Bazel 6-9 and keeps `--disk_cache=/tmp/dc` working everywhere, which is the cheaper side of that trade. The generated docstring names the four, so the answer to "why isn't `--disk_cache` in here?" lives in the file, and an AXL test pins two of them against a future regeneration off an older Bazel alone.

`bazel/flags.axl` — the one place tasks pick all of this up, with the arg names as constants so a rename cannot leave a reader behind:

```python
_PASSTHROUGH_FLAGS_ARG = "bazel_passthrough_flags"
_PASSTHROUGH_STARTUP_FLAGS_ARG = "bazel_passthrough_startup_flags"
_PASSTHROUGH_ARGS = [_PASSTHROUGH_FLAGS_ARG, _PASSTHROUGH_STARTUP_FLAGS_ARG]
_VALUE_FLAGS_ARG = "bazel_value_flags"

        _VALUE_FLAGS_ARG: args.string_list(
            default = BAZEL_VALUE_FLAGS,
            config_only = True,
            description = "Forwarded Bazel flags that take a separate value (`-c opt`) … set from `.aspect/config.axl` to extend it for a wrapper's flags.",
        ),
        _PASSTHROUGH_FLAGS_ARG: args.passthrough(
            position = "post_command",
            value_flags_from = _VALUE_FLAGS_ARG,
            description = "A flag this task does not recognize is forwarded to Bazel rather than rejected. …",
        ),
```

```python
def _passthrough(ctx, name: str) -> list:
    """Claim the flags the CLI parser could not attribute to a declared arg …"""
    return list(ctx.args.claim(name))

def _disclaim_passthrough(ctx) -> None:
    """Account for both passthrough buckets on a path that spawns no Bazel at
    all, so the runtime does not report the user's flags as silently dropped."""
    for name in _PASSTHROUGH_ARGS:
        _passthrough(ctx, name)

def _resolve_flags(ctx) -> list:
    flags = list(ctx.args.bazel_flags)
    flags.extend(_passthrough(ctx, _PASSTHROUGH_FLAGS_ARG))   # claims as it folds in
    ...
```

## Deliberate limits

- Outside the value-flag list, a collected flag never claims the next token. `--unknown_flag 8 //...` forwards the flag and leaves `8` as a target pattern for Bazel to reject; on a task with no positionals, clap reports `unexpected argument '8'`.
- `--` still ends routing, so `aspect build -- //... -//experimental/...` and `aspect run //x -- --binary-arg` are untouched.
- A flag the task declares but typed on the wrong side of its name (`aspect --deployment=prod build`) is left for clap, so its suggestion still appears.
- Features may not declare a bucket — their args are injected into every task, so it would claim flags meant for any of them.
- On the tasks that opt in, a mistyped *aspect* flag now reaches Bazel and returns `Unrecognized option` instead of clap's "did you mean". That is the cost of `--config=ci` working everywhere; the lever to narrow it is a parameter on `bzl.flags.args()`.

---

- Searched for relevant documentation and updated as needed: yes (`docs/design.md`; docsite follow-up)

Bazel flags can now be typed directly on `aspect` commands instead of being wrapped in `--bazel-flag=`. `aspect build --remote_download_all //...` forwards the flag as a Bazel command option, `aspect --output_base=/tmp/o build //...` forwards it as a startup option, and flags that take a separate value work in Bazel's own spellings (`-c opt`, `--jobs 8`, `--jobs=8`). Tasks that don't forward flags to Bazel (e.g. `delivery`) still reject unrecognized flags. `--config ci` now expands from the space form as well as `--config=ci`. Task authors get `args.passthrough(position = "pre_command" | "post_command")`, read a bucket with `ctx.args.claim(name)`, and can keep any flag off the CLI with `config_only = True` — replaceable per repo from `config.axl`.

### Test plan

- New test cases added:
  - `cmd.rs` — 26 routing tests: in-place vs moved routing, flags interleaved with a group path, the `--` boundary, misplaced task flags, a task with no bucket (still a parse error), routed values reaching `ctx.args` split by position, value-flag pairs in both spellings, nothing-to-take cases (`--flag=value`, `--flag=`, `-c=opt`, hyphen-led next token, end of argv), scan resumption after a taken value, a pre-command flag and its value moving together, a command name never taken as a value, a stray pre-command token ending the walk, a non-UTF8 argv token, short clusters classified by all of their characters, a root-only flag forwarded after the task name, a config override replacing the arity list, `config_only` unparseable yet delivered for all eight flag kinds, and `no_flag_is_routed_where_clap_accepts_it` over the whole built clap surface.
  - `bazelrc` — 4 expander tests: space-form expansion, consuming only the value, a valueless `--config` passing through, and cycle/skip handling through either spelling.
  - `passthrough.rs` / `multi_phase.rs` — 8 claim-contract tests: the shared `unclaimed` core (claimed, empty and non-list buckets), the diagnostic's wording, the pre-spawn refusal across every server-touching `ctx.bazel` entry point, that claiming first lets the same task through: claimed passes, read-without-claim fails with exit 2, empty bucket is not a failure, a failing task keeps its own exit code, and the diagnostic's wording.
  - `arg.rs` / `task.rs` / `names.rs` — arg-type validation: unknown position, duplicate position, both positions allowed, alias-default rejection, features rejected, and `config_only` accepted on all eight flag kinds, rejected with `required`/`short`/`long` on each, and absent from the four argv-shaped kinds.
  - `axl-smoke` — a flag Bazel rejects must come back as Bazel's error rather than hang, asserted for both spellings: `--bazel-flag=--definitely_not_a_flag` (the escape hatch) and the bare `--definitely_not_a_flag` this PR routes on its own. Both produce byte-identical output and exit 2. Sits beside the `--task:id` / `--task:name` checks, which assert the *CLI's* own validation message so a misspelled flag there can't pass for the wrong reason after a full Bazel invocation.
  - `tools/wrapper-test.sh` — 87 cases, green under both bash 5 and the bash 3.2 macOS ships; the rewriting assertions are replaced with verbatim-forwarding ones for every shape (space values, short flags, `--`, mixed aspect/bazel, pre-verb startup options).
  - `bazel_flags_test.axl` — 9 tests: passthrough ordering, claim-on-resolve, disclaim, bucket optionality, the arg-shape contract, the generated list's shape (value-taking flags present, booleans and the arity-disagreeing four absent, no label-shaped entries), space-form `--config` name extraction, the `passthrough_args` shape that `delivery` opts in with, and the startup-option arity list.
- Manual testing — in a scratch workspace with a genrule at `//pkg:g`:
  ```
  aspect --output_base=/tmp/ob build --announce-bazel-command=true -c opt --config ci --remote_download_all //pkg:g
  # bazel --output_base=/tmp/ob --ignore_all_rc_files build --isatty=0 -c opt --announce_rc --remote_download_all … -- //pkg:g
  # → startup slot, `-c opt` as a pair, ci config expanded, flag forwarded. ✅ Passed

  aspect <task> --definitely_not_a_bazel_flag …
  # ERROR: --definitely_not_a_bazel_flag :: Unrecognized option   (bazel's error, not clap's)
  # confirmed on build, test, run, lint, format, gazelle, ci warming

  aspect build -vv //pkg:g                    # ERROR: -vv :: Invalid options syntax: -vv
  aspect init --remote_download_all           # error: unexpected argument  (no bucket)
  aspect spawns --remote_download_all         # scratch task that spawns bazel without claiming
  # error: the flag --remote_download_all would have no effect … before it was about to run bazel
  #        (pointing at the ctx.bazel.build line, before bazel starts)
  aspect looks --remote_download_all          # scratch task that reads without claiming
  # ERROR: the flag --remote_download_all had no effect: looks collected it … → exit 2
  ```


crates/aspect-cli/src/cmd.rs
```diff
@@ -633,6 +650,261 @@ fn with_override(mut clap_arg: ClapArg, arg: &Arg, override_strings: &[String])
...
+
+/// Split a flag token into the name Clap would look up, or `None` when the
+/// token is not a flag at all (`-` and `--` are values and separators).
+fn parse_flag(token: &str) -> Option<Flag<'_>> {
+    if let Some(long) = token.strip_prefix("--") {
+        let name = long.split('=').next().unwrap_or(long);
+        return (!name.is_empty()).then_some(Flag::Long(name));
+    }
+    let short = token.strip_prefix('-')?;
+    short.chars().next().map(Flag::Short)
```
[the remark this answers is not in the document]
gregmagolan: Good catch, and it was worse than reported — fixed in e7606e2c.

The cluster half is now what you described: a short token is recognized only when *every* character of the run resolves to a declared short, with the first value-taking one ending the walk (the rest of the token is its value, per clap's `-j8`). One undeclared character makes the whole token unknown, so `-bx` routes while a genuine `-bn` cluster stays clap's.

Chasing your `-vv` example turned up a second bug behind it, though: fixing the cluster alone would not have forwarded it. Clap propagates only *globals* into a subcommand, and `--version`/`-v` are declared at the root only — so `aspect build --version` is `error: unexpected argument '--version' found`, and `-vv` was reported as `unexpected argument '-v'`. The router was accumulating every root arg as "declared" at every level, so it declined to route a token clap was going to reject either way.

Recognition is now per slot, matching what clap accepts in each position: after the task name a flag must be one the *task* accepts (its own args, feature args, generated `--help`, and inherited globals), while before the task name anything declared anywhere stays put so a misplaced task flag (`aspect --deployment=prod build`) still gets clap's suggestion instead of being handed to Bazel.

Verified end to end — `aspect build -vv //pkg:g` now reaches Bazel, which answers about its own flag namespace:

```
ERROR: -vv :: Invalid options syntax: -vv
```

Tests added: `classifies_a_short_cluster_by_all_of_its_characters` and `routes_a_root_only_flag_typed_after_the_task_name`. The existing invariant test became `no_flag_is_routed_where_clap_accepts_it` and now checks the two slots against their own sets, still driven off the built clap surface rather than a list of names.

crates/aspect-cli/src/builtins/aspect/bazel/value_flags.axl
```diff
@@ -0,0 +1,378 @@
+"""The Bazel flags that take a separate value (`-c opt`, `--jobs 8`).
+
+GENERATED by tools/gen_bazel_value_flags.sh from `bazel help` for Bazel
+9.0.1 — do not edit by hand. Regenerate after a Bazel upgrade.
```
jbedard: Many flags have been removed before 9.0.1 or added after, we can't assume the flags based on a single version like this.

gregmagolan: Updated to take the unions of flags with values for last 4 major versions: 6, 7, 8, 9

crates/axl-runtime/src/engine/bazel/mod.rs
```diff
@@ -296,6 +330,18 @@ fn read_startup_flags<'v>(this: values::Value<'v>) -> anyhow::Result<Vec<String>
...
 
 /// The active `RunCommand` set via `use_rc`, if any.
+/// Gate every invocation on the passthrough claim contract: a flag the task
+/// collected and never claimed would be silently absent from the command line
+/// we are about to run, so refuse here rather than after the build.
+///
+/// A frozen `Bazel` (config-time) carries no task and so nothing to check.
+fn require_claimed_flags<'v>(this: values::Value<'v>) -> anyhow::Result<()> {
+    match this.downcast_ref::<Bazel>() {
+        Some(bazel) => bazel.claims.require_claimed(),
```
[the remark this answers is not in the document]
gregmagolan: Right, and it was wider than the two you named — fixed in the latest commit.

`shutdown`, `health_check`, `recover_poisoned_sandbox` and `cancel_invocation` all read the active rc's startup flags and then talk to a server, so all four now run the same gate as build/test/query/info. Your reasoning is what makes these worse than the build case rather than equivalent: an unclaimed startup option means the active rc names a different server than the user asked for, so stopping or probing the default one is worse than doing nothing at all. The helper's doc now says that, so the next method added here has the rule in front of it.

The test walks every entry point instead of asserting the one that happened to be wired first:

```rust
for call in [
    r#"ctx.bazel.build("//...", stdout = None, stderr = None).wait()"#,
    r#"ctx.bazel.query("//...")"#,
    r#"ctx.bazel.info()"#,
    "ctx.bazel.shutdown()",
    "ctx.bazel.health_check()",
    "ctx.bazel.recover_poisoned_sandbox()",
] { … expect_err … }
```

`version` and `aspect` are untouched: neither reads startup flags nor reaches a server.

crates/aspect-cli/src/builtins/aspect/bazel/value_flags.axl
jbedard: That means you can't use the CLI with bazel@HEAD? Requiring these to be hardcoded still seems wrong to me.

Why can't we just passthru what the CLI does not handle?

jbedard:
Your "After" flowchart in the PR description is broken

crates/aspect-cli/src/builtins/aspect/bazel/value_flags.axl
gregmagolan: Discussed out-of-band on Slack

gregmagolan:
> Your "After" flowchart in the PR description is broken

Works on my chrome 🤷 

Screenshot 2026-08-23 at 1 52 06 PM

crates/axl-runtime/src/engine/bazel/mod.rs
```diff
@@ -296,6 +330,25 @@ fn read_startup_flags<'v>(this: values::Value<'v>) -> anyhow::Result<Vec<String>
...
+/// An unclaimed startup option means the active rc names a different server than
+/// the user asked for, so stopping or probing the default one is worse than
+/// doing nothing.
+///
+/// A frozen `Bazel` (config-time) carries no task and so nothing to check.
+fn require_claimed_flags<'v>(this: values::Value<'v>) -> anyhow::Result<()> {
+    match this.downcast_ref::<Bazel>() {
+        Some(bazel) => bazel.claims.require_claimed(),
+        None => Ok(()),
+    }
```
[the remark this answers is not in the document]
gregmagolan: Confirmed and fixed. `version()` does reach a server, and worse than the general case: `info::release_version()` → `server_info()` → `server_info_with_startup_flags(&[])`, which spawns `bazel info server_pid release` with **empty** startup flags — so it probes the *default* server, which is precisely what an unclaimed startup option means we shouldn't do. Memoized in a `OnceLock`, so the one probe that escapes sets the value for the whole process.

My own earlier pass missed it: I grepped each method body for `read_startup_flags|spawn` and `version` has neither — the spawn is one call deeper.

Now gated, and `ctx.bazel.version()` is in the `reaching_bazel_without_claiming_fails_before_the_spawn` loop alongside build/test/query/info/shutdown/health_check/recover_poisoned_sandbox. The `#[allow(unused)] this` came off, since the receiver is used now.

tools/bazel
```diff
@@ -718,188 +341,64 @@ EOF
...
+        ;;
+    esac
+    if [[ -z "$first_bare" ]]; then
+        first_bare="$arg"
     fi
-    pre_verb+=("$arg")
-    prev_was_value_taking=0
-    i=$((i + 1))
 done
+[[ -z "$verb" ]] && verb="$first_bare"
```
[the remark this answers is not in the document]
gregmagolan: Valid — fixed, and thanks for the concrete argv.

`bazel --bazelrc build query 'deps(//x)'` was selecting `build` and routing the query through aspect, which then read `query` and `deps(//x)` as target patterns for a build. So the wrapper keeps one list after all: `BAZEL_STARTUP_VALUE_FLAGS`, the 16 startup options that take a separate value.

Worth being precise about why that isn't a step back from removing the flag lists: this isn't flag *rewriting*, it's argv tokenization. The wrapper has to find the verb to decide which binary to exec, and finding it means knowing which bare tokens are values. Only *startup* options matter, since scanning stops once the verb is found — which is what keeps it at 16 entries rather than 400.

Boolean startup options are deliberately excluded. A list-free "skip the token after any `=`-less flag" rule would have fixed your case but broken `bazel --batch query 'deps(//x)'` — `--batch` consumes nothing, so skipping `query` would lose the verb and route a plain-bazel query to aspect. That trade is worse: `--batch`, `--noblock_for_lock` and friends are common in scripts.

Three regression tests added:

```
run --bazelrc build query 'deps(//foo)'    → INVOKED:bazel   (value skipped)
run --batch query 'deps(//foo)'            → INVOKED:bazel   (boolean consumes nothing)
run --bazelrc=build query 'deps(//foo)'    → INVOKED:bazel   (=value consumes nothing)
```

87 cases green under both bash 5 and the bash 3.2 macOS ships.

tools/gen_bazel_value_flags.sh
```diff
@@ -0,0 +1,158 @@
...
+    sed -nE \
+        -e 's/^  (--[a-z_]+) \(.*/\1/p' \
+        -e 's/^  (--[a-z_]+) \[(-[a-zA-Z])\] \(.*/\1\n\2/p'
+}
+
+boolean_flags() {
+    sed -nE \
+        -e 's/^  --\[no\]([a-z_]+) \(.*/--\1/p' \
+        -e 's/^  --\[no\]([a-z_]+) \[(-[a-zA-Z])\] \(.*/--\1\n\2/p'
+}
```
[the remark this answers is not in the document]
gregmagolan: Fixed, though with one correction to the premise: this machine's `/usr/bin/sed` (macOS, BSD) *does* expand `\n` in a replacement — I checked before changing anything, since the table it generated was correct:

```
$ printf '  --jobs [-j] (an integer)\n' | sed -nE -e 's/^  (--[a-z_]+) \[(-[a-zA-Z])\] \(.*/\1\n\2/p' | od -c
0000000    -   -   j   o   b   s  \n   -   j  \n
```

So the generated list was never mangled. The portability concern stands anyway, and is worth acting on: `\n` in a replacement is a GNU extension, a sed that takes it literally would emit `--jobsn-j`, and the failure mode is silent — the flag *count* stays plausible, so the existing per-major floor check wouldn't catch it.

Your first suggestion, separate single-capture rules, turned out not to work here, which is worth recording: chained `s///p` expressions each rewrite the pattern space the next one would match, so after rule 2 substituted `--jobs`, rule 3 saw `--jobs` rather than the original line and `-c`/`-j` silently disappeared (399 entries instead of 401 — caught by diffing against the previous table).

So parsing moved to awk, your second suggestion — one pass, both captures, no escape-sequence dependency:

```awk
/^  --[a-z_]+ \(/ { print $1; next }
/^  --[a-z_]+ \[-[a-zA-Z]\] \(/ { print $1; short = $2; gsub(/[][]/, "", short); print short }
```

Regenerated output is byte-identical to the sed version. I also added the guard this class of bug actually needs: every emitted entry must match `--flag` or `-x`, so a future parse slip fails loudly instead of shipping a plausible-looking count of mangled names.

tools/gen_bazel_value_flags.sh
```diff
@@ -0,0 +1,200 @@
...
+# Deliberately omitted: label-shaped flags from rule sets (`--//pkg:flag`) and
+# their `--flag_alias` spellings. Bazel *requires* `=` for those, so accepting a
+# space-separated value would make `aspect` take command lines that plain `bazel`
+# rejects, and the repro commands we print would not run when pasted. A repo that
+# needs more can extend the list from `config.axl`; see `bazel/flags.axl`.
+set -o errexit -o nounset -o pipefail
+
+SCRIPTPATH="$(cd -- "$(dirname "$0")" >/dev/null 2>&1 && pwd -P)"
+OUT="${SCRIPTPATH}/../crates/aspect-cli/src/builtins/aspect/bazel/value_flags.axl"
+MAJORS=(6 7 8 9)
```
jbedard: Should we support the next RC so we don't have to publish a new aspect-cli version when that RC gets released?

gregmagolan: Yes. I'll pull in Bazel 10 RC flags too.

jbedard: and drop bazel6?

gregmagolan: no reason to drop bazel 6... if on a newer bazel version where a flag is removed and you use it, bazel will error either way whether we forward it with a value or not

PR #1420, opened 2026-09-01
fix(bes): retry Cancelled and Unknown so a GOAWAY does not lose build events

cristifalcas:
## Problem

The BuildBarn frontend retires gRPC connections after `maxConnectionAge: 600s`, with a `maxConnectionAgeGrace: 3600s`. Its `:8980` listener also carries BES traffic, so long-running invocations can encounter GOAWAY while publishing build events.

We observed two failure modes that `is_retryable` treated as terminal:

### `Cancelled` during post-stream lifecycle publishing

```text
invocation_finished: status: 'The operation was cancelled',
self: "operation was canceled"
```

`invocation_finished` is issued after the event stream completes. When its existing channel has been retired, hyper can cancel the queued request and tonic surfaces `Code::Cancelled`. `retry_lifecycle` could recover by retrying on a usable connection, but the classifier previously rejected the error.

### `Unknown` during event streaming

```text
status: 'Unknown error',
self: "h2 protocol error: error reading a body from connection"
```

When a frontend pod terminates mid-stream, tonic can map the h2 failure to `Code::Unknown`. `drive_stream` previously treated this as fatal instead of reconnecting and replaying unacknowledged events.

## Evidence

Measured on `awd-silo-prod` against the `silo-gcp` Buildkite pipeline:

| Failure mode | Before frontend change | After frontend change |
|---|---:|---:|
| `Cancelled` on `invocation_finished` | 0 / 127 jobs | 6 / 91 jobs |
| `Unknown` mid-stream | 3 / 127 jobs | 7 / 91 jobs |

The `Cancelled` failure affected 6 of 18 invocations longer than 600 seconds and none of the 72 shorter invocations. Fisher's exact test gives `p = 3.0e-5`.

## Change

`is_retryable` now classifies `Code::Cancelled` and `Code::Unknown` as transient.

- Lifecycle calls are retried through the existing bounded `retry_lifecycle` path.
- Broken event streams reconnect and replay buffered events under their original sequence numbers.

## Safety

- BES lifecycle calls are idempotent.
- Stream replay preserves `OrderedBuildEvent.sequence_number`, allowing the server to deduplicate replayed events.
- Local shutdown closes the upstream event channel and produces `DriveOutcome::UpstreamClosed`; it does not enter this status classifier.
- Lifecycle retries remain bounded by `max_retries`, per-attempt deadlines, backoff, and the post-stream `bes_timeout`.
- Stream failures without acknowledgement progress exhaust the existing retry budget.

## Tests

- `retry::tests::classifier_status_codes` covers the complete retryable set, including `Cancelled` and `Unknown`, while keeping authentication, permission, argument, and endpoint errors terminal.
- `grpc::tests::unknown_mid_stream_reconnects_and_replays` simulates tonic mapping an h2 failure without a gRPC status to `Unknown`; the sink reconnects, replays, and delivers all ten events.
- `bazel test //crates/axl-runtime:test` passes.

### DIFF
--- a/crates/axl-runtime/src/engine/bazel/sink/grpc.rs
+++ b/crates/axl-runtime/src/engine/bazel/sink/grpc.rs
@@ -1044,10 +1044,8 @@ mod tests {
         /// long-lived streams expecting the client to reconnect and
         /// resume from the last acked event.
         AckThenShed(u32),
-        /// Ack the first N requests of each connection, then end the
-        /// response stream with UNKNOWN — how tonic surfaces an h2
-        /// protocol error (no gRPC status on the wire) when the backend
-        /// pod is terminated mid-stream.
+        /// Ack the first N requests on each connection, then simulate tonic
+        /// mapping an h2 failure without a gRPC status to `Unknown`.
         AckThenUnknown(u32),
     }
 
--- a/crates/axl-runtime/src/engine/bazel/sink/grpc.rs
+++ b/crates/axl-runtime/src/engine/bazel/sink/grpc.rs
@@ -1015,8 +1015,6 @@ mod tests {
     use super::*;
 
     use std::pin::Pin;
-    use std::sync::Arc;
-    use std::sync::atomic::{AtomicU32, Ordering};
     use std::time::Duration;
 
     use axl_proto::build_event_stream::{Progress, build_event::Payload};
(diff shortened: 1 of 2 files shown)


gregmagolan (approved):
🌮

PR #1421, opened 2026-09-01
feat(launcher): ASPECT_CLI_DEBUG_BUILD to select the debug build per invocation

gregmagolan:
Today the `-debug-` CLI variant can only be requested by committing `version("…", debug = True)` to a repository's `.aspect/version.axl`. That is the wrong granularity for the case it exists for — reproducing a crash on one CI job. The pin lands in the repo, applies to every task and every developer, and has to be reverted afterwards, so in practice people either skip the debug build or carry a revert commit.

`ASPECT_CLI_DEBUG_BUILD` (any non-empty value) forces the debug variant for a single invocation, so a job opts in through its own environment with no repository change:

```yaml
environment:
  ASPECT_CLI_DEBUG_BUILD: "1"
```

It only turns the variant *on* — a `version.axl` that already asks for it is unaffected — and the version pin is still honored. It's applied in `resolve_config`, so it covers all three config sources: the repo's `version.axl`, the user-level one, and the built-in default where there is no file to carry a `debug = True` at all. Like `debug = True`, it selects the asset name the launcher derives, so a source naming an explicit `artifact` is unaffected, and a release with no `-debug-` asset still falls back to the primary binary.

`ASPECT_DEBUG` is unrelated and unchanged: it makes the launcher log verbosely and does not change which binary is fetched. That distinction has already cost a user a round-trip while chasing a crash, so the new variable is named to not collide with it.

---

- Searched for relevant documentation and updated as needed: yes (documented on the constant alongside the `debug = True` config it overrides)

- feat: `ASPECT_CLI_DEBUG_BUILD=1` makes the launcher fetch the `-debug-` CLI build for that invocation only, so a single CI job can run the diagnostic build without committing `debug = True` to `.aspect/version.axl`.

### Test plan

- New test cases added: `debug_build_env_forces_the_debug_variant` covers no config file, a `version.axl` with no `debug` key (pin still honored), a config that already sets `debug = True`, and empty-means-unset. Verified it fails with the override removed.
- Manual testing: against the real `v2026.36.1` release, with a `version.axl` pinning `2026.36.1` and no `debug` key — `ASPECT_CLI_DEBUG_BUILD=1` downloads `aspect-cli-debug-<triple>` and `aspect version` prints `2026.36.1 (debug build)`; unset or empty downloads the primary and prints `2026.36.1`.

### DIFF
--- a/crates/aspect-launcher/src/config.rs
+++ b/crates/aspect-launcher/src/config.rs
@@ -58,8 +58,32 @@ pub struct AspectCliConfig {
     /// by querying the releases API for the latest available release.
     version: Option<String>,
     /// Whether to fetch the `-debug-` build variant instead of the primary
-    /// binary.
+    /// binary. Set by `version("…", debug = True)` or forced per-invocation by
+    /// [`DEBUG_BUILD_ENV`].
     debug: bool,
+}
+
+/// Environment variable that forces the `-debug-` build variant for a single
+/// invocation. Any non-empty value enables it, matching `ASPECT_DEBUG` (which
+/// is unrelated: it makes *this launcher* log verbosely, and does not change
+/// which binary is fetched).
+///
+/// Reproducing a crash usually means running the debug build for one CI job,
+/// which `version.axl` serves badly: that pin is committed, applies to every
+/// task and every developer on the repository, and has to be reverted
+/// afterwards. An environment variable scopes the switch to the job that
+/// needs it.
+///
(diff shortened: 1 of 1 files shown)


gregmagolan:
Closing — this duplicates `ASPECT_DEBUG_CLI`, which already does exactly this (added in #1359, `debug_cli_mode` in the launcher's main.rs). I missed it because my search pattern looked for `env::var` while the call site uses the bare imported `var(...)`. No change needed.

PR #1427, opened 2026-09-07
feat(cli): apply performance defaults on Workflows runners, starting with the server heap

cristifalcas:
## Why

Bazel adds no heap flag to the server JVM ([`blaze.cc`](https://github.com/bazelbuild/bazel/blob/master/src/main/cpp/blaze.cc) sets none), so the server runs on HotSpot's default of **25% of RAM**. On a Workflows runner that is a quarter of an instance the job has to itself, and nobody writes `startup --host_jvm_args` into a CI bazelrc.

Evidence from silo-gcp Test jobs on the 4-vCPU, 16 GB runners (build 61152 on the default; 61160 after aspect-build/silo#12234 capped it at 40%):

| | 61152, default 25% | 61160, 40% cap |
|---|---|---|
| Heap ceiling | ~3.7 GB | ~5.9 GB |
| Heap in use, max | 3,833 MB | 5,255 MB |
| Time at or above 3.5 GB | 25% of the build | 38% |
| Minor GCs | 282, 37.5 s of 596 s | 124, 14.7 s of 381 s |
| Whole machine, peak | 8.5 GB of 14.8 | 9.2 GB of 14.8 |

The default was binding: heap pinned within 100 MB of the ceiling for a quarter of the build, 6% of wall in GC. The builds are on different branches, so the comparison is directional.

## What

A `performance_defaults` boolean, **default on**:

- on the `Workflows` feature (`ctx.features[Workflows].args.performance_defaults = False` / `--workflows:performance-defaults=false`), injected into the runner startup and build flags;
- on `aspect ci bazelrc` (`--performance-defaults`), written into the rc so plain `bazel` calls in the job carry the same flags.

One switch rather than a knob per value: every value in the bundle is already overridable through the repo's own `.bazelrc`, so per-value args would add surface without adding control. The bundle's first entry is `--host_jvm_args=-XX:MaxRAMPercentage=40`; `performance_defaults_build_flags()` is empty today so the next command flag is one line and no new arg.

## Precedence rules of the bundle

- **Command flags beat the repo rc on a runner.** They are appended after the repo's flags via `extra_flags`, and Bazel takes the last.
- **Startup flags keep the repo's `startup` section last.** Bazel compares the server's JVM arguments in order, and the `~/.bazelrc` that `aspect ci bazelrc` writes is read before the workspace rc, so `aspect <task>` and plain `bazel` must both produce `[ours, repo's]` or they restart the shared server on every alternation. A repo's `-Xmx` wins over the percentage by JVM rule either way, and a repo's own `MaxRAMPercentage` wins by being last.
- **A new startup entry restarts the server once on upgrade.** Documented in the arg description. Command flags cost nothing to add.

## Why `MaxRAMPercentage` rather than a computed `-Xmx`

- The JVM already sizes against the cgroup limit in a container; a MemTotal probe would hand a 4 GB pod on a 256 GB node a 100 GB ceiling.
- HotSpot ignores `MaxRAMPercentage` whenever `-Xmx` is present, so no collision handling.
- The flag text is identical on every machine, so a bazelrc shared across hosts never differs.

## Not in this PR

- `--jobs`: the same profiles show the in-flight action count at the `--jobs=40` ceiling only 1 to 3% of the build, with the Bazel server itself at 3 to 4 of the 4 cores throughout. Concurrency is not the limiter on these runners; server CPU is.
- `--local_resources`: Bazel's 67% default assumes a small server, but local actions were near zero on these RBE runners. Left for a follow-up if a local-execution profile shows contention.

## Testing

`aspect dev test-bazelrc` — 14 tests (2 new: the bundle's exact flag text; on/off through `get_bazelrc_flags`, survival through `startup_flags_for_rc`, and the rendered `startup` line). `aspect ci bazelrc --help` and `aspect feature workflows` show the new switch.

## Field check: Ironclad

Decoded 54 complete invocations from the Workflows BEP archive of Ironclad's GCP deployment (objects uploaded 26 Aug and 3 Sep 2026; CircleCI runners on `c4-standard-16-lssd`, 16 vCPU / 60 GB; Bazel 9.2.0; remote execution with `--noremote_local_fallback`).

**Heap today.** No invocation carries any `--host_jvm_args`, so the server runs on HotSpot's default cap of 25% of RAM, about 15 GB. Against that cap:

| Signal (54 invocations) | Value |
|---|---|
| `OOM_ERROR` exits / `OUT_OF_MEMORY` aborts | 0 |
| Full GCs (`peak_post_gc_heap_size` non-zero) | 0 |
| Invocations with G1 old-gen collections | 33 of 54 |
| Builds over 5 min with old-gen collections | 11 of 11 |
| Eden bytes reclaimed per invocation, p50 / p90 / max | 23 / 522 / 698 GB |
| Skyframe nodes after invocation | 1.7M – 2.6M |
| Builds per server (servers persist across CI jobs) | median 23, max 91 |

The heap never filled, but the long builds do cross G1's ~45% initiating threshold (about 7 GB live), so a 6 GB cap would already be too small for them.

**What this PR changes for Ironclad at the current size:** the cap goes from ~15 GB to ~24 GB. No observable effect; there is no pressure to relieve.

**Where it earns its place:** the runner's Bazel server averages 1.1 cores of 16 during long builds, and 4 of 922,106 actions in the sample ran locally, so the right-size move is `c4-standard-8` (30 GB). There HotSpot's default drops to 7.5 GB, right where the long builds already live, and the resize alone would create the GC-thrash failure this PR exists to prevent. With the PR the cap is 12 GB and Ironclad needs no bazelrc change. The `--local_resources` caveat does not apply to them: nothing executes locally.

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/bazelrc.axl
+++ b/crates/aspect-cli/src/builtins/aspect/bazelrc.axl
@@ -18,10 +18,11 @@ What it writes depends on where it runs:
   - On a Workflows runner: the full set — generic flags, the remote-cache
     endpoint (from env), and the runner-specific flags (repository cache,
     per-runner output paths, identity header, the Workflows-cache compression
-    gates, the server heap cap). The heap cap must match the `Workflows`
-    feature's `server_heap_percent`: `--host_jvm_args` is a server startup
-    option, so a plain `bazel` reading a different value than `aspect <task>`
-    injects would restart the shared server on every alternation.
+    gates, and the `performance_defaults` bundle). The bundle switch must match
+    the `Workflows` feature's `performance_defaults`: it carries a server
+    startup option, so a plain `bazel` reading different JVM arguments than
+    `aspect <task>` injects would restart the shared server on every
+    alternation.
   - On other CI, or with an explicit `--output`: only the generic flags (plus
     the remote-cache endpoint when its env vars are set) — nothing tied to a
     runner's mounts or identity. Useful when an Aspect remote cache / BES is
@@ -43,7 +44,6 @@ Only file output is supported — there is no stdout mode.
 load("@aspect//bazel/build_metadata.axl", "version_gte")
 load(
     "./private/lib/environment.axl",
-    "DEFAULT_SERVER_HEAP_PERCENT",
     "detect_ci",
     "error",
     "get_bazelrc_flags",
@@ -148,7 +148,7 @@ def _bazelrc_impl(ctx: TaskContext) -> int:
         (startup_flags, build_flags) = get_bazelrc_flags(
             environment = environment,
             aspect_root_dir = ctx.std.env.aspect_root_dir(),
-            server_heap_percent = ctx.args.server_heap_percent,
+            performance_defaults = ctx.args.performance_defaults,
         )
         startup_flags = startup_flags_for_rc(startup_flags)
     else:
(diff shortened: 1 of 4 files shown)


cristifalcas:
Closing: superseded by an experiment on the silo side instead.

PR #1428, opened 2026-09-09
fix(bazel): key execlog reader on client pid; probe with startup flags

jeffpignataro:
## Problem

A customer's `aspect test` jobs started failing after a successful build:

```text
INFO: Build completed successfully, 2 total actions
...
error: execlog stream thread error: io error: process exited before the file was created
```

1. A pre-command hook ran `aspect ci bazelrc`, writing `--output_base=/mnt/ephemeral/output/dirt/dirt` into `~/.bazelrc`.
2. `Build::spawn` probed `bazel info server_pid release` **with no startup flags**. Bazel resolved that from the rc files, so the probe landed on the task's own output base but with rc-derived startup options, and returned that server's pid.
3. The real `bazel test` ran with the runner's explicit startup flags. Bazel saw different startup options on the running server and restarted it — `Starting local Bazel server (9.2.0)` three seconds after `bazel health check passed`.
4. The execlog reader had been started before spawn with the probed server pid as its liveness signal. That pid was gone, so `StreamingFile::open` returned `BrokenPipe` immediately.
5. `wait()` joined the streams after a green build, hit the stream error and `bail!`ed, never reporting bazel's exit status.

Until that hook change the rc files pointed the probe at a *different*, long-lived output base (Rosetta's `__main__`), so the wrong-but-alive pid masked the bug. It only bites with an exec-log sink registered (the artifacts feature's `upload_exec_log`).

## Fix

- **Probe the server the invocation will use.** `Build::spawn` now calls `server_info_with_startup_flags(&startup_flags)`, as `query` already did. The version probes behind version-gated flags (`resolve_flags_for_running_bazel`, `resolve_rc_version`) take the startup flags too, so they no longer read rc files and never start a stray server.
- **Bound the execlog reader by the client pid.** The `--execution_log_compact_file` path is chosen before spawn; the reader thread starts after `cmd.spawn()` with `child.id()`, mirroring the BES reader. Bazel's `CompactSpawnLogContext` opens the file in its constructor and closes it in `afterCommand`, both before the client exits, so the client's exit is the end-of-stream signal. `galvanize::StreamingFile` keys on `is_pid_alive(pid)` instead of "does `pid` hold the file open" — the server holds it, and the server may be replaced mid-invocation.
- **Keep bazel's exit code visible.** `wait()` collects stream errors instead of bailing on the first. If bazel failed, the stream errors are printed as warnings and bazel's status is returned. If bazel succeeded and a stream failed, the call still fails (a truncated artifact must not pass silently), but the message now carries the exit code.

Not changed here: `WorkspaceEventStream::spawn_with_pipe` and the BES `Pipe`'s `IfOpenForPid` policy still receive the pre-spawn server pid. With the probe fixed that pid is the right server, so they are correct unless bazel restarts for some other reason; making them client-pid-bounded is a follow-up.

## Test plan

- [x] `bazel test //crates/galvanize:galvanize_test` — two new tests: the reader gives up when the bounding pid is already dead before the file appears, and it keeps reading past EOF while the pid lives (with no process holding the file open) and ends only when the pid exits.
- [x] `bazel build //crates/axl-runtime`
- [x] `bazel test //crates/axl-runtime:test`
- [ ] Manual: on a Workflows runner with `aspect ci bazelrc` in a pre-command hook and `upload_exec_log` on, `aspect test` completes and the `.execlog.zstd` sink is written.

https://claude.ai/code/session_01HTa7Rk2dND4P5AXWZ7VBcV

### DIFF
--- a/crates/axl-runtime/src/engine/bazel/build.rs
+++ b/crates/axl-runtime/src/engine/bazel/build.rs
@@ -819,7 +819,12 @@ impl Build {
         announce: AnnounceSpawn,
         rt: AsyncRuntime,
     ) -> Result<Build, std::io::Error> {
-        let (pid, version) = super::info::server_info()?;
+        // Probe with the invocation's own startup flags so the answer describes
+        // the server this command will talk to. A bare `bazel info` reads the rc
+        // files instead, which can name a different output base — or the same
+        // one with different options, in which case the invocation restarts the
+        // server and the pid returned here is dead before the build begins.
+        let (pid, version) = super::info::server_info_with_startup_flags(&startup_flags)?;
 
         let span = tracing::info_span!(
             "ctx.bazel.build",
@@ -889,22 +894,17 @@ impl Build {
             }
         }
 
-        let mut execlog_stream = if execution_logs {
+        // The reader itself starts after `cmd.spawn()`, keyed on the child pid.
+        let execlog_out = if execution_logs {
             // If there is a CompactFile sink, let Bazel write directly to its path
             // so no separate temp file or tee step is needed for that copy.
-            let direct_path = if compact_paths.is_empty() {
-                None
+            let out = if compact_paths.is_empty() {
+                std::env::temp_dir().join(format!("execlog-out-{}.bin", uuid::Uuid::new_v4()))
             } else {
-                Some(std::path::PathBuf::from(compact_paths.remove(0)))
+                std::path::PathBuf::from(compact_paths.remove(0))
             };
-            let (out, stream) = ExecLogStream::spawn_with_file(
-                pid,
-                direct_path,
-                compact_paths,
-                !decoded_sinks.is_empty(),
-            )?;
             cmd.arg("--execution_log_compact_file").arg(&out);
-            Some(stream)
+            Some(out)
         } else {
             None
         };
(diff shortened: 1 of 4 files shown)


crates/axl-runtime/src/engine/bazel/stream/execlog.rs
```diff
@@ -181,33 +181,27 @@ impl ExecLogStream {
...
     ///
     /// The thread streams the file as Bazel writes it using [`galvanize::StreamingFile`],
-    /// which busy-polls for file existence at open time and retries reads while Bazel
-    /// holds the file open. It self-terminates when Bazel closes the file.
+    /// which busy-polls for file existence at open time and retries reads while the
+    /// client is alive.
     pub fn spawn_with_file(
-        pid: u32,
-        out_path: Option<PathBuf>,
+        path: PathBuf,
```
jbedard: `path` is a pretty horrible name here

crates/axl-runtime/src/engine/bazel/build.rs
```diff
@@ -819,7 +819,12 @@ impl Build {
         announce: AnnounceSpawn,
         rt: AsyncRuntime,
     ) -> Result<Build, std::io::Error> {
-        let (pid, version) = super::info::server_info()?;
+        // Probe with the invocation's own startup flags so the answer describes
```
jbedard: This seems like a pretty over detailed bloated comment

crates/axl-runtime/src/engine/bazel/stream/execlog.rs
jeffpignataro: yeah good call - changed to `execlog_path` for clarity.

crates/axl-runtime/src/engine/bazel/build.rs
jeffpignataro: Yeah, definitely. Tightened up the verbiage.

thesayyn:
to me this sounds like we should just forward the correct startup flags that should fix the issue here?

thesayyn:
Also relying on the client_pid is not an answer here. if you have bazel/tools wrapper, bazelisk etc the client_pid is never the real bazel client, even when it is, its not the client that writes this file, its the server that does it.

jeffpignataro:
Closing based on https://github.com/aspect-build/aspect-cli/pull/1428#issuecomment-5669056146 - I'll try to look for a simpler solution here.

PR #1439, opened 2026-09-12
fix(cache): honor Bazel test selection in cache diff

jeffpignataro:
`aspect cache diff -- //... -//excluded/...` currently constructs an invalid query, and query-based enumeration can report tests excluded by Bazel's test filters. Resolve the affected-test scope through Bazel's loading-only test phase so ordered patterns, suites, filters, and pattern files follow Bazel's selection behavior. Wildcards exclude `manual` tests; explicit manual targets remain selectable. Requires Bazel 7+.

Require complete selection events and reject loading failures or partial results. Disable BES for the private selection pass, honor command announcements, and release its temporary files promptly. Construct the attribution query RC only when a dependency miss requires it, retain resolved version-conditional flags, and render options through Bazel's common section so command-inapplicable flags are ignored by Bazel. Preserve pattern-file scope through the probe and precise prebuild.

Add an automatically discovered AXL regression suite and a GitHub Actions integration matrix for Bazel 7.4.0, 8.6.0, and 9.2.0. Each job downloads the CLI artifact built from this PR in the same workflow run, starts a checksum-pinned local `bazel-remote`, and checks real hits/misses against a controlled baseline. Execution markers assert that probes never run tests and overreport never runs build actions. Each job uses two Bazel workers and a 512 MiB JVM heap limit; no shared-cache credentials or Docker are needed. Downloads and port-bind races have bounded retries.

### Validation

- Parser/RC checks and 31 real-Bazel selection cases pass locally, including negative patterns, filters, suites, external targets, loading failures, and unreachable BES/cache endpoints.
- Expanded real-cache integration passes locally on Bazel 7.4.0, 8.6.0, and 9.2.0 in 118–129 seconds per version, storing 220–224 KiB.
- Integration covers unchanged hits, dependency misses, command-specific CLI flags, build-only configs on warm hits, announcements, suites, re-inclusion, precise prebuild, refreshed hits, and empty-scope pattern files/`--exec` suppression.
- Builtin import-style check and formatting pass locally.
- Selection was compared with native Bazel on a larger repository across 15 scenarios, with matching results.
- [CI passed on 6e95359d](https://github.com/aspect-build/aspect-cli/actions/runs/34728197294), including both shared-cache probes, full build/test jobs, and all 46 discovered AXL suites. The selection suite passed 31 real-Bazel cases in 83 seconds.
- The three GitHub-hosted cache-integration jobs passed in 47, 44, and 108 seconds of test time (Bazel 7, 8, and 9), or about 3.8 runner-minutes total including setup. Fixture timings do not establish whole-repository cache-diff performance.

### DIFF
--- a/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
+++ b/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
@@ -60,10 +60,12 @@ both sides or neither). The probe resolves flags from the same rc/config as
 `aspect test`/`build`, so they align by default.
 """
 
+load("@aspect//bazel/invocation.axl", "target_patterns")
 load("@aspect//bazel.axl", "BazelTrait", bzl = "bazel")
 load("@bazel//proto/remote_logging.axl", "remote_logging")
 load("@std//time.axl", "time")
 load("./private/lib/ansi.axl", "ansi")
+load("./private/lib/cache_selection.axl", "select_tests")
 load("./private/lib/environment.axl", "color_enabled", "warn")
 load("./private/lib/remote_executor.axl", "start_dummy_executor", "stop_dummy_executor")
 load("./private/lib/runnable.axl", "apparent_label")
@@ -131,6 +133,7 @@ def _invalidate_action_nodes(ctx: TaskContext, rc, endpoint_auth_flags: list, an
     err_path = "/tmp/aspect-cache-diff-invalidate-{}.log".format(ctx.task.name)
     flags = endpoint_auth_flags + [
         "--experimental_remote_require_cached",
+        "--target_pattern_file=",  # the anchor is explicit, even with a pattern-file scope
         "--remote_default_exec_properties=" + _SALT_SENTINEL + "=1",
         "--keep_going",
         "--remote_download_outputs=minimal",
@@ -241,9 +244,11 @@ def _build_missing_non_test(ctx: TaskContext, patterns: list, rc, endpoint_auth_
     returned on a hard build failure (a real broken target, not a cache miss)."""
     err_path = "/tmp/aspect-cache-diff-build-{}.log".format(ctx.task.name)
     _t = time.monotonic()
+    pattern_file = rc.flag_value("--target_pattern_file", command = "test")
+    pattern_flags = ["--target_pattern_file=" + pattern_file] if pattern_file else []
     status = ctx.bazel.build(
         rc = rc,
-        flags = endpoint_auth_flags + ["--remote_upload_local_results", "--keep_going"],
+        flags = endpoint_auth_flags + pattern_flags + ["--remote_upload_local_results", "--keep_going"],
         stdout = None,
         stderr = ctx.std.fs.create(err_path),
         announce_version = announce_version,
@@ -311,7 +316,7 @@ def _classify_and_stream(ctx: TaskContext, entries: list, test_targets: list, rc
             continue
         cause = {"target": e.label, "mnemonic": e.mnemonic}
         _tq = time.monotonic()
-        dep_tests = [apparent_label(t.name) for t in ctx.bazel.query("tests(rdeps(//..., {}))".format(e.label), rc = rc)]
+        dep_tests = [apparent_label(t.name) for t in ctx.bazel.query("tests(rdeps(//..., {}))".format(json.encode(e.label)), rc = rc)]
         rdeps_secs += time.monotonic() - _tq
         for label in dep_tests:
             if label in test_set:
@@ -338,7 +343,6 @@ def _emit_json(ctx: TaskContext, mode: str, test_targets: list, affected: list, 
 def _impl(ctx: TaskContext) -> int:
     color = color_enabled(ctx.std)
     patterns = list(ctx.args.targets)
-    union = " + ".join(patterns)
     mode = ctx.args.mode
     timing = {}  # seconds per phase — split bazel work vs our analysis at the end
 
@@ -354,6 +358,7 @@ def _impl(ctx: TaskContext) -> int:
     # `--nokeep_state_after_build --nouse_action_cache` (see `_run_probe`), so it
     # needs no separate or wiped base.
     rc = ctx.bazel.parse_rc(startup_flags = startup_flags, flags = flags)
+    patterns = target_patterns(ctx.args.is_explicit("targets"), patterns, "", rc, "test")
 
     # Explicit, up-front requirement: no remote cache → nothing to diff against.
     if not rc.flag_value("--remote_cache", command = "test") and not rc.flag_value("--remote_executor", command = "test"):
@@ -365,14 +370,15 @@ def _impl(ctx: TaskContext) -> int:
     # the probe and pre-build below both hit it.
     endpoint_auth_flags = bzl.rc_flags(ctx, rc, "test")
 
-    # Enumerate + reverse-dep under the same resolved flags as the probe/build,
-    # so a `--config`/`--bazel-flag` that changes package loading or the test set
-    # can't make the queries see a different graph than the probe.
+    # Bazel owns ordered patterns, suite expansion, and test filters. Query's
+    # tests() does not have the same selection semantics as `bazel test`.
     _t = time.monotonic()
-    test_targets = sorted([apparent_label(t.name) for t in ctx.bazel.query("tests(%s)" % union, rc = rc)])
+    test_targets = select_tests(ctx, patterns, rc, endpoint_auth_flags = endpoint_auth_flags)
     timing["enumerate"] = time.monotonic() - _t
+    if test_targets == None:
+        return 1
     if not test_targets:
-        _err(ctx, "No test targets in scope for: " + union)
+        _err(ctx, "No test targets in scope for: " + " ".join(patterns))
         return 0
 
     _err(ctx, ansi.style("Diffing {} test target(s) against the remote cache [--mode={}]…".format(len(test_targets), mode), ansi.BOLD, color))
@@ -519,7 +525,8 @@ diff = task(
             description = (
                 "Bazel target patterns scoping the affected set (e.g. " +
                 "`//services/...`). Defaults to `//...` — every test in the " +
-                "workspace. Intersected with `tests(<patterns>)`."
+                "workspace. Uses Bazel test filters. Separate negative patterns " +
+                "from flags with `--` (e.g. `-- //... -//personal/...`)."
             ),
         ),
     } | bzl.flags.args("the cache-diff probe") | bzl.announce.args("the cache-diff probe"),
--- a/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
+++ b/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
@@ -65,7 +65,7 @@ load("@aspect//bazel.axl", "BazelTrait", bzl = "bazel")
 load("@bazel//proto/remote_logging.axl", "remote_logging")
 load("@std//time.axl", "time")
 load("./private/lib/ansi.axl", "ansi")
-load("./private/lib/cache_selection.axl", "select_tests")
+load("./private/lib/cache_selection.axl", "attribution_rc", "select_tests")
 load("./private/lib/environment.axl", "color_enabled", "warn")
 load("./private/lib/remote_executor.axl", "start_dummy_executor", "stop_dummy_executor")
 load("./private/lib/runnable.axl", "apparent_label")
@@ -291,6 +291,7 @@ def _classify_and_stream(ctx: TaskContext, entries: list, test_targets: list, rc
     time (`timing["attribute"]`) from our own bookkeeping (`timing["classify"]`)."""
     _t_all = time.monotonic()
     test_set = {t: None for t in test_targets}
+    query_rc = attribution_rc(ctx, rc)
 
     # Frontier misses, de-duplicated by label (first mnemonic wins).
     missed = []
@@ -316,7 +317,7 @@ def _classify_and_stream(ctx: TaskContext, entries: list, test_targets: list, rc
             continue
         cause = {"target": e.label, "mnemonic": e.mnemonic}
         _tq = time.monotonic()
-        dep_tests = [apparent_label(t.name) for t in ctx.bazel.query("tests(rdeps(//..., {}))".format(json.encode(e.label)), rc = rc)]
+        dep_tests = [apparent_label(t.name) for t in ctx.bazel.query("tests(rdeps(//..., {}))".format(json.encode(e.label)), rc = query_rc)]
         rdeps_secs += time.monotonic() - _tq
         for label in dep_tests:
             if label in test_set:
--- a/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
+++ b/crates/aspect-cli/src/builtins/aspect/cache_diff.axl
@@ -8,6 +8,10 @@ miss, so the probe runs no actions. The gRPC log records one `GetActionResult`
 per action Bazel probed; a miss anywhere in a test's transitive closure trips at
 that action (the *frontier*), and the affected tests are attributed from there.
 
+Selection follows Bazel test patterns and filters (Bazel 7+). Wildcards exclude
+manual-tagged tests; explicit manual targets remain selectable. Negative
+patterns are processed in order and must follow `--`.
+
 Output contract (built for pipes):
 
   - stdout — the affected test labels, one per line, nothing else. Pipe it:
@@ -291,7 +295,7 @@ def _classify_and_stream(ctx: TaskContext, entries: list, test_targets: list, rc
     time (`timing["attribute"]`) from our own bookkeeping (`timing["classify"]`)."""
     _t_all = time.monotonic()
     test_set = {t: None for t in test_targets}
-    query_rc = attribution_rc(ctx, rc)
+    query_rc = None
 
     # Frontier misses, de-duplicated by label (first mnemonic wins).
     missed = []
@@ -317,6 +321,8 @@ def _classify_and_stream(ctx: TaskContext, entries: list, test_targets: list, rc
             continue
         cause = {"target": e.label, "mnemonic": e.mnemonic}
         _tq = time.monotonic()
+        if query_rc == None:
+            query_rc = attribution_rc(ctx, rc)
         dep_tests = [apparent_label(t.name) for t in ctx.bazel.query("tests(rdeps(//..., {}))".format(json.encode(e.label)), rc = query_rc)]
         rdeps_secs += time.monotonic() - _tq
         for label in dep_tests:
@@ -354,10 +360,9 @@ def _impl(ctx: TaskContext) -> int:
     flags = list(bazel_trait.base_flags) + bzl.flags.resolve(ctx)
     announce_version, announce_command = bzl.announce.resolve(ctx)
 
-    # One run command for the queries, the precise pre-build, and the probe — all
-    # on the default output base. The probe forces fresh remote lookups via
-    # `--nokeep_state_after_build --nouse_action_cache` (see `_run_probe`), so it
-    # needs no separate or wiped base.
+    # Share the RC and default output base across selection, precise prebuild,
+    # and probe. The salt-delete pass refreshes action execution without
+    # discarding analysis; --nouse_action_cache bypasses persistent action hits.
     rc = ctx.bazel.parse_rc(startup_flags = startup_flags, flags = flags)
     patterns = target_patterns(ctx.args.is_explicit("targets"), patterns, "", rc, "test")
 
@@ -374,12 +379,13 @@ def _impl(ctx: TaskContext) -> int:
     # Bazel owns ordered patterns, suite expansion, and test filters. Query's
     # tests() does not have the same selection semantics as `bazel test`.
     _t = time.monotonic()
-    test_targets = select_tests(ctx, patterns, rc, endpoint_auth_flags = endpoint_auth_flags)
+    test_targets = select_tests(ctx, patterns, rc, endpoint_auth_flags = endpoint_auth_flags, announce_version = announce_version, announce_command = announce_command)
     timing["enumerate"] = time.monotonic() - _t
     if test_targets == None:
         return 1
     if not test_targets:
-        _err(ctx, "No test targets in scope for: " + " ".join(patterns))
+        scope = " ".join(patterns) if patterns else "--target_pattern_file=" + (rc.flag_value("--target_pattern_file", command = "test") or "")
+        _err(ctx, "No test targets in scope for: " + scope)
         return 0
 
     _err(ctx, ansi.style("Diffing {} test target(s) against the remote cache [--mode={}]…".format(len(test_targets), mode), ansi.BOLD, color))
@@ -465,6 +471,9 @@ diff = task(
         "action results must already be uploaded (e.g. `bazel test //... " +
         "--remote_upload_local_results`). Against a sparse cache the result " +
         "over-reports.\n\n" +
+        "Selection follows Bazel test patterns and filters (Bazel 7+). Wildcards " +
+        "exclude manual-tagged tests; explicit manual targets remain selectable. " +
+        "Pass negative patterns after `--`.\n\n" +
         "Output: affected labels stream to stdout (one per line, pipeable: " +
         "`aspect cache diff | xargs bazel test`); progress, per-target reasons, " +
         "and the summary go to stderr. Pass --exec=\"<cmd>\" to run `<cmd> " +
--- a/crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
+++ b/crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
@@ -1,0 +1,81 @@
+"""Resolve cache-diff test scope through Bazel's test loading phase."""
+
+load("./runnable.axl", "apparent_label")
+
+def selected_tests(events: list, exit_code: int) -> list[str] | None:
+    """Accept only a complete, successful loading-only test selection."""
+
+    # With analysis disabled, `test` exits NO_TESTS_FOUND even when loading
+    # selected tests. Never treat that exit code alone as successful selection.
+    if exit_code != 4:
+        return None
+    labels = {}
+    aborted = {}
+    expansions = 0
+    finished = False
+    complete = False
+    for event in events:
+        if complete:
+            return None
+        complete = event.get("lastMessage", False)
+        if "expanded" in event:
+            expansions += 1
+            for child in event.get("children", []):
+                if "targetConfigured" not in child:
+                    return None
+                labels[child["targetConfigured"]["label"]] = None
+        if "aborted" in event:
+            label = event.get("id", {}).get("targetConfigured", {}).get("label")
+            if event["aborted"].get("reason") != "NO_ANALYZE" or not label:
+                return None
+            aborted[label] = None
+        if "finished" in event:
+            if finished or event["finished"].get("exitCode", {}).get("code") != 4:
+                return None
+            finished = True
+    if not complete or not finished or expansions != 1 or labels != aborted:
+        return None
+    return sorted({apparent_label(label): None for label in labels})
+
+def select_tests(ctx: TaskContext, patterns: list[str], rc, directory: str | None = None, endpoint_auth_flags: list = []) -> list[str] | None:
+    """Load the original patterns with test-command RC and filtering semantics."""
+    work = ctx.std.fs.mkdtemp(prefix = "aspect-cache-selection-")
+    ctx.defer(lambda: ctx.std.fs.remove_dir_all(work))
+    event_path = work + "/events.json"
+    error_path = work + "/stderr"
+    status = ctx.bazel.test(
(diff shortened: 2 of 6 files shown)


jeffpignataro (commented):
Automated recall-oriented review (extra-high effort): 14 findings, most severe first.

Reviewed at `360ca752`. A 15th finding — `attribution_rc` passing the `(value, condition)` tuples straight through, so the query rc re-evaluated version gates against assumed-latest — was fixed by that commit while this review was being written, and has been dropped.

Headline: `query_rc` is now computed eagerly, which turns a build-only `--config` into an unconditional hard failure including on the fully-cached happy path that previously never ran a query. The `_SELECTION_FLAGS` denylist is also worth a second look: it fixes six flags by name, but the same "Unrecognized option" breakage remains for every other build/test-only flag a user can type.

Anchoring note: two cache_diff.axl findings point at lines outside the diff hunks — the stale comment block at L357 and the task `description` at L458-479 — and are anchored to the nearest hunk line (L359 and L529) with the real target named in the comment.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
```diff
@@ -286,6 +291,7 @@ def _classify_and_stream(ctx: TaskContext, entries: list, test_targets: list, rc
     time (`timing["attribute"]`) from our own bookkeeping (`timing["classify"]`)."""
     _t_all = time.monotonic()
     test_set = {t: None for t in test_targets}
+    query_rc = attribution_rc(ctx, rc)
```
jeffpignataro: `attribution_rc` runs `rc.expand_all(command = "query")` on every invocation now, not just when there is a non-test miss to attribute.

`command_ancestors("query")` is `&[]` (crates/bazelrc/src/lib.rs:13-20), so a `build:ci`-only config section is undefined for `query` and `expand_configs` returns `UndefinedConfig` (crates/bazelrc/src/expand.rs:155-160). `parse_rc` at L361 passes no `skip_config_if_missing`.

So `aspect cache diff --config=ci` in a repo that defines only `build:ci ...` now dies with an uncaught AXL error on *every* run — including a fully-cached tree that previously printed `Affected 0 of N` without ever reaching a query. Compute `query_rc` lazily on the first non-test miss, or pass `skip_config_if_missing`.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,111 @@
...
+def attribution_flags(flags: list) -> list:
+    result = []
+    skip_value = False
+    for flag in flags:
+        if skip_value:
+            skip_value = False
+            continue
+        value = flag[0] if type(flag) == "tuple" else flag
+        if value.split("=", 1)[0] in _SELECTION_FLAGS:
+            skip_value = "=" not in value
```
jeffpignataro: `skip_value = "=" not in value` consumes the next token unconditionally, even when that token is another flag.

`is_two_token_pair` (crates/bazelrc/src/lib.rs:797-805) deliberately refuses to pair a bare flag with a following token that starts with `-`. Here `["--test_filter", "--keep_going"]` silently drops `--keep_going` from the query rc, so the attribution query aborts on the first bad package instead of continuing and the affected set comes back truncated.

Mirror the same guard: only skip when the next element does not start with `-`.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,111 @@
+"""Resolve cache-diff test scope through Bazel's test loading phase."""
+
+load("./runnable.axl", "apparent_label")
+
+def selected_tests(events: list, exit_code: int) -> list[str] | None:
+    """Accept only a complete, successful loading-only test selection."""
+
+    # With analysis disabled, `test` exits NO_TESTS_FOUND even when loading
+    # selected tests. Never treat that exit code alone as successful selection.
+    if exit_code != 4:
```
jeffpignataro: This pins the whole feature to an undocumented flag plus an exit-code accident, and it fails closed with no fallback.

`--analyze` is HIDDEN, and its own help says `--noanalyze` stops before analysis "returning zero iff the package loading completed successfully" — the 4 here comes from `TestCommand` post-processing finding no test results, not from the loading phase itself. Combined with the `expansions != 1` and `labels != aborted` equality at L36, any Bazel that exits 0 on a loading-only `test`, or that announces one extra BEP id (a `testSuiteExpansion` id would make the `event["id"]["targetConfigured"]` lookup on L28 yield `None`), makes `selected_tests` return `None` and `_impl` return 1 (cache_diff.axl:379-380).

The old `bazel query "tests(...)"` enumeration worked on every version. Worth either gating on a minimum Bazel version with a clear message, or falling back to the query when the loading events don't parse.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,111 @@
...
+                events.append(event)
+    tests = selected_tests(events, status.code)
+    if tests == None:
+        ctx.std.io.stderr.write("ERROR: cache diff could not establish the selected tests from Bazel's loading events (exit {}).\n".format(status.code))
+        ctx.std.io.stderr.write(ctx.std.fs.read_to_string(error_path))
+    return tests
+
+# These options select tests/patterns, not the dependency graph queried during
+# attribution. CLI-supplied options otherwise reach every RC command verbatim.
+_SELECTION_FLAGS = [
```
jeffpignataro: This denylist covers the six flags the integration test happened to trip, but the underlying problem is general: CLI flags land in the rc as `always` options (crates/bazelrc/src/lib.rs:175-190) and therefore show up in `expand_all(command = "query")` as regular flags — and `bazel query` accepts neither `RemoteOptions` nor the test options.

`aspect cache diff --test_env=FOO=1 -- //...`, `--runs_per_test=2`, `--build_tests_only`, `--remote_download_outputs=minimal` and `--remote_cache=grpc://...` all still kill the attribution query with "Unrecognized option" — the same class of failure this PR is fixing, five flags over.

The general fix is already in this diff's vocabulary: the case at cache_selection_test.axl:51 keeps `--default_override=0:common=...` entries precisely because Bazel ignores inapplicable options in the `common` section. Rendering the `always` options as common overrides covers every flag instead of six.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,111 @@
...
+    work = ctx.std.fs.mkdtemp(prefix = "aspect-cache-selection-")
+    ctx.defer(lambda: ctx.std.fs.remove_dir_all(work))
+    event_path = work + "/events.json"
+    error_path = work + "/stderr"
+    status = ctx.bazel.test(
+        rc = rc,
+        directory = directory,
+        # --build overrides a user's --nobuild (which changes test's exit
+        # code). --noanalyze still prevents analysis and all build/test actions.
+        flags = endpoint_auth_flags + [
```
jeffpignataro, line 55: `build_events = False` only suppresses the CLI's own `--build_event_binary_file` sink; the rc's `--bes_backend` still applies. A `build --bes_backend=grpcs://...` line (inherited by `test`, and present in every Workflows repo) means this loading-only run publishes a spurious empty invocation to the user's BES backend on every cache diff, with the upload latency on the critical path.

The integration fixture works around exactly this with `--bes_backend=` (tools/cache-diff-integration.py:506); real users get no such override. Add `--bes_backend=` to this flag list alongside the BEP-file flags.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
```diff
@@ -365,14 +371,15 @@ def _impl(ctx: TaskContext) -> int:
...
+    # tests() does not have the same selection semantics as `bazel test`.
     _t = time.monotonic()
-    test_targets = sorted([apparent_label(t.name) for t in ctx.bazel.query("tests(%s)" % union, rc = rc)])
+    test_targets = select_tests(ctx, patterns, rc, endpoint_auth_flags = endpoint_auth_flags)
     timing["enumerate"] = time.monotonic() - _t
+    if test_targets == None:
+        return 1
     if not test_targets:
-        _err(ctx, "No test targets in scope for: " + union)
+        _err(ctx, "No test targets in scope for: " + " ".join(patterns))
```
jeffpignataro: `patterns` is `[]` whenever the scope came from `--target_pattern_file` — `target_patterns` empties it on purpose (bazel/invocation.axl:84-88). So with `test --target_pattern_file=/ci/patterns` in the rc and no explicit CLI targets, this prints `No test targets in scope for: ` with nothing after the colon and no hint that the scope came from a pattern file. Name the pattern file when `patterns` is empty.

tools/cache-diff-integration.py
```diff
@@ -0,0 +1,180 @@
...
+    aspect = str(Path(shutil.which(args.aspect) or args.aspect).resolve())
+    bazel = str(Path(shutil.which(args.bazel) or args.bazel).resolve())
+    started = time.monotonic()
+    with tempfile.TemporaryDirectory(prefix="cache-diff-integration-") as tmp:
+        root = Path(tmp)
+        arch = {"x86_64": "amd64", "aarch64": "arm64"}.get(platform.machine(), platform.machine())
+        target = f"{platform.system().lower()}-{arch}"
+        binary = root / "bazel-remote"
+        url = f"https://github.com/buchgr/bazel-remote/releases/download/v{CACHE_VERSION}/bazel-remote-{CACHE_VERSION}-{target}"
+        with urllib.request.urlopen(url, timeout=60) as response:
```
jeffpignataro: A single unretried request with a 60s timeout for a ~25 MB release asset, and no proxy handling — a transient GitHub hiccup fails the `cache-diff-integration` check for reasons unrelated to the change. Worth a small retry loop.

Separately, L54 indexes `CACHE_SHA256[target]` before the assert message can help: on any platform not in the table (`windows-amd64`, a musl or s390x runner) this raises a bare `KeyError: 'windows-amd64'` rather than an unsupported-platform error. Check membership first.

tools/cache-diff-integration.py
```diff
@@ -0,0 +1,180 @@
...
+CACHE_VERSION = "2.6.2"
+CACHE_SHA256 = {
+    "darwin-amd64": "fc3b8f5d8edcf94f578cf6ffa8071d8859f5f02a38c14ae00cd3d22aa44974b6",
+    "darwin-arm64": "1d6bc5fe5be7a66c7aeea6723280402414e280ae579016d3b5e6e5c77278ad04",
+    "linux-amd64": "62e236bf8396e69396928e0d0c32062fbd5575f20fe55dc10a82eb791297e1a0",
+    "linux-arm64": "b2cabd5bf674e8d0649de2c95dddfea8d875db1a06ed9d431674c9293df273ed",
+}
+
+
+def free_port():
```
jeffpignataro, line 33: TOCTOU: the socket closes when the `with` block exits, so the port is free between here and the `Popen` at L535. On a busy runner another process can take it, bazel-remote exits with "address already in use", the readiness loop hits `server.poll() is not None`, and the job fails with the server log — an unexplained flake. Hold the sockets open until after `Popen`, or retry the bind.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,111 @@
...
+        if value.split("=", 1)[0] in _SELECTION_FLAGS:
+            skip_value = "=" not in value
+        else:
+            # expand_all already evaluated version conditions against the source RC.
+            result.append(value)
+    return result
+
+def attribution_rc(ctx: TaskContext, rc):
+    _, flags = rc.expand_all(command = "query")
+    return ctx.bazel.new_rc(startup_flags = rc.startup_flags(), flags = attribution_flags(flags))
```
jeffpignataro: Nit: `rc.startup_flags()` already ends in `--ignore_all_rc_files` (`invocation_startup_flags`, crates/bazelrc/src/lib.rs:673-686), and the query's own `resolve_for_command` calls `invocation_startup_flags()` again on the new rc — so the attribution query is spawned with the flag twice.

Harmless today (same boolean value, so no server restart), but it makes `--announce-bazel-command` output confusing in a repo that keeps a dedicated `bazelrc-e2e` test asserting no server thrash.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,111 @@
...
+            "--nokeep_going",
+            "--build_event_json_file=" + event_path,
+            # Keep this private log local; uploading its referenced files would
+            # contact the cache even for an empty selection.
+            "--nobuild_event_json_file_path_conversion",
+        ],
+        build_events = False,
+        stdout = None,
+        stderr = ctx.std.fs.create(error_path),
+        announce_version = False,
```
jeffpignataro: `announce_version` / `announce_command` are hardcoded `False`, while `_build_missing_non_test` (cache_diff.axl:253-254) and `_run_probe` (cache_diff.axl:197-198) both thread the resolved values through.

This is the one new Bazel invocation that can fail the task, so a user debugging `cache diff could not establish the selected tests from Bazel's loading events (exit 1)` under `--announce-bazel-command=true` sees every command line except the one that failed. Take them as parameters and pass them from `_impl`.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,111 @@
...
+                return None
+            finished = True
+    if not complete or not finished or expansions != 1 or labels != aborted:
+        return None
+    return sorted({apparent_label(label): None for label in labels})
+
+def select_tests(ctx: TaskContext, patterns: list[str], rc, directory: str | None = None, endpoint_auth_flags: list = []) -> list[str] | None:
+    """Load the original patterns with test-command RC and filtering semantics."""
+    work = ctx.std.fs.mkdtemp(prefix = "aspect-cache-selection-")
+    ctx.defer(lambda: ctx.std.fs.remove_dir_all(work))
```
jeffpignataro: Everything this function needs is read at L69-80, so the temp dir can be removed before `return tests` rather than held to task end. As written, `cache_selection_test.axl` calls `select_tests` 30 times and accumulates 30 `aspect-cache-selection-*` directories plus 30 deferred closures for the life of the `dev test-cache-selection` task.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
```diff
@@ -354,6 +359,7 @@ def _impl(ctx: TaskContext) -> int:
     # `--nokeep_state_after_build --nouse_action_cache` (see `_run_probe`), so it
```
jeffpignataro: (This is about the comment block starting at L357, which isn't in the diff — anchoring here as the nearest hunk line.)

It's stale in two ways now: there is exactly one query left (the rdeps attribution at L320), not "the queries"; and `_run_probe` no longer uses `--nokeep_state_after_build` — it was replaced by the salt-delete pass described in the module docstring at L46-53. A reader following this comment goes looking for a flag that is no longer passed anywhere in the file.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
```diff
@@ -519,7 +526,8 @@ diff = task(
             description = (
                 "Bazel target patterns scoping the affected set (e.g. " +
                 "`//services/...`). Defaults to `//...` — every test in the " +
-                "workspace. Intersected with `tests(<patterns>)`."
+                "workspace. Uses Bazel test filters. Separate negative patterns " +
```
jeffpignataro, line 538: Good that this says "Uses Bazel test filters" — but this is the only place it's said.

The task `description` at L458-479 and the module docstring's "Output contract" section (both outside the diff) still promise that `//...` means "every test in the workspace", and neither mentions that wildcard selection now drops `manual`-tagged tests, which the old `tests(//...)` query included. `aspect cache diff --help` renders the `description` first, so a user reading help gets the pre-PR semantics. Worth a line there and in the docstring.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,111 @@
+"""Resolve cache-diff test scope through Bazel's test loading phase."""
+
+load("./runnable.axl", "apparent_label")
+
+def selected_tests(events: list, exit_code: int) -> list[str] | None:
```
jeffpignataro: `selected_tests` has one non-test caller (`select_tests`, same file) and is otherwise loaded only by cache_selection_test.axl:3.

docs/axl.md §11: "Symbols can be exported via `testonly_` prefix for writing unit tests." The repo follows it elsewhere — `testonly_parse_endpoint` (private/lib/deliveryd.axl:342), `testonly_dispatch` / `testonly_suggest_fixes` (private/lib/lifecycle.axl:1107-1108), `testonly_config_names` / `testonly_flags_delta` (bazel.axl:19-20). Rename to `testonly_selected_tests` so the parser doesn't read as public surface.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,110 @@
...
+        value = flag[0] if type(flag) == "tuple" else flag
+        if value.split("=", 1)[0] in _SELECTION_FLAGS:
+            skip_value = "=" not in value
+        else:
+            result.append(flag)
+    return result
+
+def attribution_rc(ctx: TaskContext, rc):
+    _, flags = rc.expand_all(command = "query")
+    return ctx.bazel.new_rc(startup_flags = rc.startup_flags(), flags = attribution_flags(flags))
```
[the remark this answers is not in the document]
jeffpignataro: Fixed in 360ca752 and retained in 6e95359d. The attribution RC uses flags after the source RC has evaluated version conditions, so rebuilding it cannot re-evaluate `<8` against an assumed-latest version. Added a regression with a Bazel 7.4 RC that retains the `<8` graph flag and drops the `>=8` flag.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
jeffpignataro: Fixed in 6e95359d by restoring lazy query-RC construction on the first dependency miss requiring attribution. Fully cached and direct test-miss paths no longer expand query configs. The real-cache integration now covers a warm tree with a build-only config; it passes locally and in CI on Bazel 7.4, 8.6, and 9.2.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Fixed in 6e95359d. Replaced flag stripping with common-section rendering, and a following `--keep_going` is preserved as its own option. Added regressions for the bare-filter/following-flag case, ordinary two-token values, and negative filter values such as `--test_tag_filters -noci`.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Addressed in 6e95359d with an explicit Bazel 7+ requirement and a clear version/selection error. The parser accepts either exit 0 or 4 only when the matching, complete loading-event contract is satisfied. Added a real-cache CI matrix for Bazel 7.4.0, 8.6.0, and 9.2.0; all three passed. Incomplete or unsupported events still fail closed: a query fallback would silently change filtering and exclusion semantics.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Fixed in 6e95359d. Attribution options now go through Bazel's common section, letting Bazel ignore command-inapplicable options instead of maintaining a six-flag denylist. The real-cache dependency-miss test now supplies test_env, runs_per_test, build_tests_only, remote_download_outputs, and remote_cache explicitly, including a two-token value. It passes in CI on Bazel 7, 8, and 9.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Fixed in 6e95359d by explicitly passing `--bes_backend=` to the loading-only selection invocation. Added a selection regression with an unreachable configured BES endpoint; selection succeeds without publishing to it.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
jeffpignataro: Fixed in 6e95359d. Empty pattern-file scope now names `--target_pattern_file=<path>` in the diagnostic. The real-cache integration checks that message and verifies that empty scope still suppresses `--exec`; it passes across the three CI Bazel versions.

tools/cache-diff-integration.py
jeffpignataro: Fixed in 6e95359d: validate platform membership before downloading/indexing the checksum table, and retry download I/O failures up to three attempts with bounded backoff. Checksum mismatches still fail immediately. `urllib.request.urlopen` already honors the standard proxy environment variables. Locally fault-injected a transient download failure and an unsupported platform to verify both paths.

tools/cache-diff-integration.py
jeffpignataro: Fixed in 6e95359d with bounded retries using fresh ports after an address-in-use startup failure. Holding the reservation sockets through Popen would prevent bazel-remote from binding them, so the server startup retries the race instead. Locally fault-injected a bind collision and verified that the next attempt uses new ports and succeeds.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Fixed in 6e95359d. The derived RC removes the existing `--ignore_all_rc_files` before constructing the new RC, whose invocation adds it once. Other resolved startup flags remain intact.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Fixed in 6e95359d. The selection helper now takes the resolved announce_version and announce_command values from cache diff. The integration test enables `--announce-bazel-command=true` and asserts that the loading-only `--noanalyze` invocation is visible; this passes on all three CI Bazel versions.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Fixed in 6e95359d: remove the temporary directory after reading the events and diagnostics, before returning. A small guarded deferred cleanup remains for exceptional exits, but successful calls no longer retain their directories until task completion.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
jeffpignataro: Fixed in 6e95359d. Rewrote the comment to describe the shared RC/output base and the current salt-delete plus `--nouse_action_cache` probe, removing the stale multiple-query and `--nokeep_state_after_build` description.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
jeffpignataro: Fixed in 6e95359d. Both the module documentation and the task description now state Bazel test-selection semantics, wildcard exclusion of manual-tagged tests, explicit manual-target support, ordered negative patterns, and the Bazel 7+ requirement.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Fixed in 6e95359d: renamed the parser to `testonly_selected_tests` and updated its production caller and tests. The attribution flag helper also uses the testonly prefix. The builtin import-style check passes.

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
```diff
@@ -131,6 +137,7 @@ def _invalidate_action_nodes(ctx: TaskContext, rc, endpoint_auth_flags: list, an
     err_path = "/tmp/aspect-cache-diff-invalidate-{}.log".format(ctx.task.name)
     flags = endpoint_auth_flags + [
         "--experimental_remote_require_cached",
+        "--target_pattern_file=",  # the anchor is explicit, even with a pattern-file scope
```
jbedard, line 140: Adding this empty ensures no targets are selected? I don't see how targets could have been added at all, although this seems like a reasonable safe-guard 🤷

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
```diff
@@ -311,7 +321,9 @@ def _classify_and_stream(ctx: TaskContext, entries: list, test_targets: list, rc
             continue
         cause = {"target": e.label, "mnemonic": e.mnemonic}
         _tq = time.monotonic()
-        dep_tests = [apparent_label(t.name) for t in ctx.bazel.query("tests(rdeps(//..., {}))".format(e.label), rc = rc)]
+        if query_rc == None:
```
jbedard: If we're only doing this once why is it in the loop?

tools/cache-diff-integration.py
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env python3
+"""Exercise cache diff against an isolated REAPI cache. Requires Bazel and the CLI.
```
jbedard, line 2: I find this type of test very hard to read or maintain. But I guess this is the only real test we have of the actual cache functionality? We need to create an isolated repo with its own cache and know the exact cache state in order to test `aspect cache` with proper assertions?

We could easily run plenty of `aspect cache ...` commands as a plain shell script to test different CLI args, but without a generated isolated repo+cache like this we can't really assert the results?

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,118 @@
...
+    ctx.defer(lambda: ctx.std.fs.remove_dir_all(work) if ctx.std.fs.exists(work) else None)
+    event_path = work + "/events.json"
+    error_path = work + "/stderr"
+    status = ctx.bazel.test(
+        rc = rc,
+        directory = directory,
+        # --build overrides a user's --nobuild (which changes test's exit
+        # code). --noanalyze still prevents analysis and all build/test actions.
+        flags = endpoint_auth_flags + [
+            "--bes_backend=",
```
jbedard, line 56: Is this to explicitly override and remove any `--bes_backend` configured somewhere like .bazelrc?

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
```diff
@@ -0,0 +1,118 @@
...
+        announce_command = announce_command,
+        *patterns
+    ).wait()
+    events = []
+    if ctx.std.fs.exists(event_path):
+        for line in ctx.std.fs.read_to_string(event_path).splitlines():
+            if line:
+                event = json.try_decode(line, None)
+                if type(event) != "dict":
+                    events = []
```
jbedard: One bad entry wipes out all good entries? With no error message or anything?

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
jeffpignataro: Yeah that's fair. I'll adjust it so it happens first.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Yep, shouldn't be silent. I'll fix that.

crates/aspect-cli/src/builtins/aspect/private/lib/cache_selection.axl
jeffpignataro: Yeah that's the intent. So that was don't send an invocation to the backend (just in case).

tools/cache-diff-integration.py
jeffpignataro: Would it be better to have concrete fixtures for stuff like `BUILD`, `defs.bzl` and `bazelrc` in the tools folder? Would that make it easier to read/maintain?

crates/aspect-cli/src/builtins/aspect/cache_diff.axl
jeffpignataro: If the user passes in `--target_pattern_file=` then Bazel will error with `"Command-line target pattern and --target_pattern_file cannot both be specified"` so this is avoids that and makes the anchor the only taget.

Yes, just a safe-guard.

thesayyn:
code is hard to read but i understand the intent.

PR #1443, opened 2026-09-14
feat(helpers): expose file_uri_to_path for resolving Bazel output URIs

burmudar:
Adds `file_uri_to_path(env, uri)` to the public `@aspect//helpers.axl` facade so user AXL can resolve a Bazel output `file://` or `bytestream://` URI to a local path, the way the previous Aspect client's `Client.GetFilePath` did. `bytestream://` URIs resolve through the runner's bb_clientd mount, whose location is read from `env` (`ctx.std.env`), matching the `env`-first shape of `detect_build_url`, `detect_ci_host` and `aspect_web_ui_url`.

`bb_clientd_root` is a runner platform detail and stays private in `lib/bazel_results.axl`; it now takes `env` instead of `ctx`. The explicit-root form lives on as `uri_to_path_under_root(uri, bb_root)` for the lint task, which substitutes `None` after probing whether the mount is live, and for `process_event`, which has no env.

```python
load("@aspect//helpers.axl", "file_uri_to_path")

path = file_uri_to_path(ctx.std.env, file.file)  # "" when no local path exists
```

- `@aspect//helpers.axl` gains `file_uri_to_path(env, uri)`, resolving a Bazel output `file://` or `bytestream://` URI to a local path

### Test plan

- New test cases added: `aspect tests axl` covers the public `file_uri_to_path(env, uri)` under each bb_clientd env configuration, plus `uri_to_path_under_root` and `bb_clientd_root`; `aspect dev test-helpers` exercises the facade re-export


gregmagolan:
Thanks for the PR, William. I pushed a commit onto your branch (the PR allows maintainer edits) that reshapes the public surface: `bb_clientd_root` is a runner-platform detail, so it stays private, and `file_uri_to_path` now takes `env` like the other facade helpers and reads the mount configuration itself. Usage becomes:

```python
load("@aspect//helpers.axl", "file_uri_to_path")

path = file_uri_to_path(ctx.std.env, file.file)
```

Will land with this addition. Let me know if that shape doesn't cover what you need and we can fix forward.

gregmagolan (approved):

PR #1447, opened 2026-09-15
feat(axl): graceful child shutdown: Child.terminate() and a live-children registry

jbedard:
This is a prereq for `run --watch` which may need to terminate the child process.

### Test plan

- New test cases added


jbedard:
@codex review

crates/axl-runtime/src/engine/bazel/live.rs
```diff
@@ -1,71 +1,31 @@
...
-use std::sync::Mutex;
 use std::sync::OnceLock;
 
-use super::process;
+use crate::engine::process::{self, ProcessGuard, ProcessSupervisor};
 
-fn registry() -> &'static Mutex<Vec<u32>> {
-    static REG: OnceLock<Mutex<Vec<u32>>> = OnceLock::new();
-    REG.get_or_init(|| Mutex::new(Vec::new()))
+fn supervisor() -> &'static ProcessSupervisor {
```
thesayyn, line 7: i am curious what other design spaces are open here. maybe unix process groups?

jbedard: I think that's what codex was recommending against: https://github.com/aspect-build/aspect-cli/pull/1447#discussion_r4010736601

Or is that only recommending to not be the same group as the CLI?

jeffpignataro (approved):

crates/axl-runtime/src/engine/bazel/live.rs
thesayyn: i am not sure sure because i don't know the details. however i believe thats a cleaner api because we can't track all children because we don't control what subprocesses do.

thesayyn:
Can we include a description here that at least explains why we need this?

jbedard:
After talking with @thesayyn he pointed out how this is just expanding a mess and making it worse and this should not be done at all.

PR #1449, opened 2026-09-15
feat(axl): spawn stdin/env, Writable.try_write, and runinfo target tags

jbedard:
Generic child-I/O surface used by watch mode but independent of it: r.spawn gains stdin= (Command.stdin modes) and env= (applied last, over the target's RunEnvironmentInfo and runfiles vars); the runinfo aspect captures the target's tags and determine_entrypoint exposes them; and Writable.try_write is a best-effort write-and-flush returning False instead of raising when the receiving child has exited or wedged — the probe a notify loop needs to talk to a child that may be gone.

### Test plan

- New test cases added


crates/axl-types/src/stream.rs
```diff
@@ -394,6 +451,35 @@ fn writable_methods(registry: &mut MethodsBuilder) {
...
+            return Err(anyhow!("try_write() expects a string or bytes"));
+        };
+        match &*io {
+            Writable::ChildStdin(stdin) => {
+                let guard = stdin.lock().unwrap();
+                let mut borrowed = guard.borrow_mut();
+                Ok(borrowed
+                    .as_mut()
+                    .and_then(|writer| nonblocking_write(writer, data).ok())
+                    .unwrap_or(0) as u32)
```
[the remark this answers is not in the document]
jbedard: Returning 0 for non-Unix is the intended contract rather than an accidental conflation: callers are written to treat 0 as "this child cannot be reached over stdin" and fall back to their non-stdin path (the watch loop restarts the child instead of notifying it), so an unsupported platform degrades to the same safe behavior as a closed pipe. Propagating Unsupported would instead turn every notify attempt on those platforms into a hard error. Documented this on the try_write docstring in the latest push.

crates/axl-types/src/stream.rs
```diff
@@ -394,6 +451,40 @@ fn writable_methods(registry: &mut MethodsBuilder) {
...
+    /// Attempts one nonblocking write to a child's stdin and returns the number
+    /// of bytes accepted. Returns 0 when the stream is closed, the pipe is full,
+    /// or the write fails. A short write may occur; retry only the unwritten
+    /// suffix. Other writable stream kinds are not supported.
+    ///
+    /// On non-Unix platforms nonblocking child-stdin writes are unsupported
+    /// and every call returns 0. That is deliberate: callers treat 0 as "this
+    /// child cannot be reached this way" and fall back to their non-stdin
+    /// path (e.g. the watch loop restarts the child instead of notifying it).
+    fn try_write<'v>(
```
thesayyn, line 464: i am little hesitant to introduce a variant that is non blocking and does not support other types of stream. if this isn't generic method then it shouldn't probably be here.

jbedard: What do you mean other types of streams? This is used when trying to write to stdin for ibazel support in #1450

jbedard: Just [making this try_write generic](https://github.com/aspect-build/aspect-cli/compare/1fb5d67de1954aa61e037413ec3f7718d9dda858..49f1bb015a3b88e79923bc178edcb99bcef3d045)?

PR #1455, opened 2026-09-15
fix(bes): stream a feature-injected BES backend from every task

joshcbarnes:
`--remote=bes` reached `build` and `test`, and silently streamed nowhere for `format`, `gazelle`, `lint`, `run`, `warming`, and `delivery`.

## The bug

The `Deployment` feature puts its backend on `BazelTrait.bes_backends`, but `collect_bes_sinks` took those URIs as an `extra_backends` parameter rather than reading the trait it is already handed. Only `bazel/invocation.axl` — the spawn path behind `build`/`test` — passed them:

```python
extra_backends = list(trait.bes_backends)
sinks = collect_bes_sinks(ctx, trait, rc, command, extra_backends = extra_backends)
```

The six tasks that build their own sinks call it with no `extra_backends`, so the default `[]` dropped the feature's backend on the floor. No error and no disclosure — those builds simply reported to nothing.

The cache half never had the problem, which is what made it confusing to diagnose: `--remote=cache` lands in `BazelTrait.base_flags`, which the spawn path reads generically, so it reaches every task. Only the BES half needed each caller to cooperate.

Observed on a GitHub-hosted CI job: `format` got `--remote_cache=grpcs://cache.aspect.build` and the credential helper, but no BES stream — so the invocation never appeared in the Web UI, while `build` and `test` from the same run did.

## History

This completes what #1374 set out to do. It introduced the seam and stated the goal:

> `--remote` and `--deployment` … now reach every bazel-driving task rather than just build/test.

It added `BazelTrait.bes_backends` without a consumer. #1326, a day later, wired it up — in `invocation.axl` only, touching `build.axl`, `test.axl`, `trait.axl` and `bazel.axl`. The six tasks with their own sink construction were never revisited.

## The fix

`collect_bes_sinks` and `dropped_bes_backends` read `bazel_trait.bes_backends` themselves, so a feature-injected backend reaches a task by construction rather than by each call site remembering. That's deliberately the opposite of threading the parameter through six more call sites: a parameter that must be passed identically everywhere is what produced the bug.

`extra_backends` stays for a caller with a backend of its own, unioned with the trait's and deduped, so a caller that still passes them gets one sink rather than two. The now-redundant threading in `invocation.axl` is dropped.

Endpoint precedence is unchanged: a backend Bazel uploads to itself via `--bes_backend` is still dropped from the CLI's sinks and still named in the "Not streaming …" disclosure.

## Test plan

- New `_test_feature_injected_backend_reaches_every_task` in `build_events_test.axl`: a backend on the trait is streamed with no `extra_backends` from the caller; it is still dropped when Bazel uploads there itself and still named by `dropped_bes_backends`; a caller passing the same backend is deduped. `_fake_trait` grows a `bes_backends` field.
- **Confirmed the test fails with the fix reverted** — `fail: streamed with no extra_backends from the caller` — rather than passing either way.
- End-to-end against a build from this branch, simulating a GitHub-hosted runner with `--remote=cache,bes`: `aspect format` streams to `grpcs://bes.aspect.build` 0 times before the fix and 1 time after.
- `aspect dev test-bes-sinks` (19), `test-bazel-invocation` (11 sections), `test-deployment-flags` (19), `test-bazel-flags` all pass; `aspect format` clean.

## Notes for reviewers

- Independent of #1442 — that PR's only `build_events.axl` change is two hunks in `bes_results_url_flag`, a different function, so these merge cleanly. The two are complementary: #1442's `--config=aspect-cloud` routes BES through Bazel's own `--bes_backend` and bypasses the CLI-sink path, while `aspect <task> --remote` still goes through it and hits this bug.
- Visible effect for anyone using `--remote=bes` (or `--remote` bare): `format`, `gazelle`, `lint`, `run`, `warming`, and `delivery` now produce Web UI invocations where they previously produced none. In this repo's CI that takes a pass from 2 hosted invocations to roughly 14.

- `aspect <task> --remote` now streams build events from every Bazel-driving task. `format`, `gazelle`, `lint`, `run`, `warming`, and `delivery` previously wired the remote cache but silently streamed no build events, so their invocations never reached the Aspect Web UI.


jbedard (approved):
> silently streamed nowhere for format, gazelle, lint, run, warming, and delivery.

Can those be tested somehow?

PR #1457, opened 2026-09-16
fix(auth): coordinate refresh across CLI processes

jeffpignataro:
When multiple CLI processes encounter an expired bearer, they can submit the same single-use refresh token. One refresh succeeds; the others receive a rejection and report that the session expired. This can make a normal bearer expiration look like a requirement to log in again.

Coordinate refresh through one transaction shared by MCP, the Bazel credential helper, and AXL credential lookup. Waiting callers re-read the store and adopt the replacement token. A store-wide advisory lock also protects login, logout, and deployment removal from lost updates to the shared credential map. File-backed credentials use private temporary files and atomic replacement.

Lock acquisition is bounded to 60 seconds and automatic refresh to 30 seconds. Refresh rejection is distinguished from transient failures, which advise retrying and preserve the stored credentials; raw token-endpoint response bodies are not exposed.

On Unix, keyring locks use a private, ownership-checked directory keyed by the effective OS user ID, independent of `HOME`, `TMPDIR`, and `XDG_RUNTIME_DIR`. Windows uses the OS known-folder API.

### Validation

- `cargo test -p axl-runtime --lib engine::aspect::`: 132 passed.
- `cargo test -p aspect-cli --test auth_refresh`: 3 passed. Separate credential-helper and AXL processes share exactly one grant; logout is not overwritten by a pending refresh; failed refreshes preserve credentials.
- The same subprocess concurrency test fails against released `v2026.37.6` with the original `session expired` message.
- `cargo fmt --all --check`, BUILD formatting, and `git diff --check` passed.
- `bazel test //crates/aspect-cli:auth_refresh_test --test_output=errors`: passed at `392167c3` (final test result reused from cache).
- `bazel test //crates/axl-runtime:test --test_filter=engine::aspect::credential_store:: --test_output=errors`: passed at `a01a773b`.
- Subprocesses with different, unusable home directories and different temp/runtime overrides contend on the same test lock. Restoring the old HOME-derived path makes this test fail. Ownership, permission, and symlink checks are covered.
- Repository pre-commit formatting hook passed.

### Scope and rollout

Tests use synthetic credentials and a loopback issuer. Live keychain integration and Frontegg session policies were not tested. This addresses the reproducible refresh race; it does not establish the cause of a particular 24-hour logout or extend provider session lifetimes.

Restart long-running MCP processes after upgrading: older binaries do not participate in the lock. Independently copied credential stores remain separate and cannot safely share a rotating refresh token.


crates/axl-runtime/src/engine/aspect/credential_store.rs
```diff
@@ -90,8 +94,68 @@ impl CredentialStore {
...
-    pub(crate) fn save_all<T: Serialize>(&self, map: &HashMap<String, T>) -> anyhow::Result<()> {
+    /// Serialize the entire read/modify/write transaction, including a rotating
+    /// refresh grant. All writers share this lock, across threads and processes.
+    /// The sidecar is never deleted: replacing it would split the lock domain.
+    pub(crate) fn update<T: Serialize + DeserializeOwned, R>(
+        &self,
+        update: impl FnOnce(&mut HashMap<String, T>) -> anyhow::Result<R>,
+    ) -> anyhow::Result<R> {
+        let path = match self {
+            Self::Keyring => default_file_path()?.with_file_name("keyring-credentials.lock"),
```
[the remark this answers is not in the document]
jeffpignataro: Fixed in a01a773b490d45000601ef46349ec7ebdb36cf8f. Unix keyring locks now use `/tmp/aspect-keyring-<effective uid>/credentials.lock`, independent of HOME, TMPDIR, and XDG_RUNTIME_DIR. The directory must be owned by that OS user with mode 0700; symlinks and unsafe permissions are rejected, and lock files are opened with O_NOFOLLOW. Windows uses the OS known-folder API.

Added a subprocess regression test that overrides HOME with unusable paths and varies TMPDIR/XDG_RUNTIME_DIR, while verifying contention on the same separate test lock without accessing real credentials. It passes with the fix and fails with the old HOME-derived path. All 132 focused auth/runtime tests, 3 refresh integration tests, and the Bazel credential-store tests pass.

crates/aspect-cli/tests/auth_refresh.rs
```diff
@@ -0,0 +1,313 @@
...
+            .as_secs()
+            + 86400);
+        let grants = Arc::new(AtomicUsize::new(0));
+        let stop = Arc::new(AtomicBool::new(false));
+        let (first_tx, first_grant) = mpsc::channel();
+        let (release, release_rx) = mpsc::channel();
+        let worker_grants = grants.clone();
+        let worker_stop = stop.clone();
+        let worker_bearer = bearer.clone();
+        let thread = std::thread::spawn(move || {
```
thesayyn, line 58: i don't quite understand the use of thread here if this code meant to be single threaded.

jeffpignataro: The thread isn't there to serve requests in parallel. It allows the main thread to keep going while a connection is parked. I'll adjust the wording though.

🤖 details:
`wait_for_grant()` returns once child #1's grant reaches the issuer, and the issuer then blocks on `release_rx` holding that connection open. Meanwhile the main thread spawns the second helper and the `AXL` child, which is exactly the window the test needs: both competing processes hit the locked refresh transaction while #1's grant is still in flight, and their requests queue on the listener. Without coordination they'd submit the old refresh token and the issuer would `401` them — that's the assertion `grants == 1` is protecting.

On one thread you'd need a state machine interleaving accept polling with `Command::spawn` to get the same interleaving. I'll move the rationale up to the spawn so it's not buried in the loop body.

PR #1459, opened 2026-09-16
fix(metadata): report the PR head commit, not the synthesized merge

joshcbarnes:
## The bug

`get_build_metadata_flags` overlaid `git show HEAD` over whatever the CI collector resolved, and treated the result as authoritative for every `COMMIT_*` field. On a PR build that's the wrong commit twice over: on GitHub Actions `pull_request`, `GITHUB_SHA` **is** the synthesized `refs/pull/N/merge` commit, and `actions/checkout` checks out that same ref by default — so both sources agree on a commit that is in neither branch's history.

What went out on every default GHA PR build:

| Field | Value |
|---|---|
| `COMMIT_SHA` | a SHA absent from the PR's commit list |
| `COMMIT_MESSAGE` | `Merge <head> into <base>` |
| `COMMIT_AUTHOR` | `GitHub <noreply@github.com>` |
| `COMMIT_TIMESTAMP` | when the merge ref was synthesized |

GitLab merged-result and merge-train pipelines have the same shape via `CI_COMMIT_SHA`. `pull_request_target` fails differently — there `GITHUB_SHA` and the checkout are the *base branch tip*, so every `COMMIT_*` described the target branch rather than the PR. Buildkite and CircleCI check out the PR head and were already correct.

This repo already knew about this and fixed it elsewhere: `github.axl:detect_commit_sha` prefers `pull_request.head.sha` and `gitlab_detect.axl:_resolve_sha` prefers `CI_MERGE_REQUEST_SOURCE_BRANCH_SHA`, each with a docstring naming this exact failure. Build metadata used neither.

## The fix

Resolve the head commit first, then describe *that* commit:

- **`detect_build_commit_sha`** (new, public) delegates to the two existing helpers. GitLab is consulted first, because `detect_commit_sha` falls through to a bare `CI_COMMIT_SHA` — the synthesized merge on those pipelines.
- **`_git_show`** takes a rev instead of hardcoding `HEAD`, and honors the exit status rather than only checking for empty stdout.
- Where there is no synthesized merge, the resolved SHA *is* `HEAD`, so this is the same read those hosts did before — Buildkite, CircleCI, GHA `push`, GitLab detached and local builds are unchanged.
- **`detect_gitlab_commit_sha`** (new, public) wraps `_resolve_sha`: build metadata runs on branch pipelines too, so it needs the SHA without the rest of the MR context `detect_gitlab_mr` returns. It honors `ASPECT_GITLAB_SHA` on the same terms `detect_gitlab_mr` does — gated on `ASPECT_GITLAB_MR_IID` — so a runner posting to a GitLab MR from another CI host reports one commit, not the override's SHA on the MR widget and the host's own in the metadata.

## Behavior changes worth reviewing

**`pull_request_target`:** `COMMIT_SHA` is now the PR head (matching what status checks attach to) and the description fields are dropped, where before it was the base branch tip's SHA and description.

**Shallow checkouts:** on the default `fetch-depth: 1` merge-ref checkout the head commit's object isn't fetched, so `git show` finds nothing and the description fields are omitted — the SHA is still correct and still emitted. Verified against a real shallow merge-ref clone: both `<head_sha>` and `HEAD^2` return `rc=128`, so there is no free local recovery. Getting the description back needs a `git fetch --depth=1 origin <head_sha>` on a path that runs for every Bazel invocation; that seemed the wrong trade, since omitting a field beats attributing the build to `noreply@github.com`. Happy to add it if reviewers disagree — repos on `fetch-depth: 0` are unaffected either way.

## Tests

New `bazel/build_metadata_test.axl`, registered in `.aspect/config.axl` and auto-discovered by `tools/run-axl-suites.sh`, so it runs in both CI pipelines:

- `detect_build_commit_sha` across 11 host/event shapes — GHA `pull_request` / `pull_request_target` / `push` / `merge_group`, a GHA PR with no event payload on disk, GitLab `merged_result` / `detached`, the `ASPECT_GITLAB_SHA` override with and without the MR IID, Buildkite, CircleCI, and off-CI.
- `get_build_metadata_flags` end to end against real scratch git repos built into the GHA `pull_request` shape: a feature commit, a forge-authored `--no-ff` merge standing in for `refs/pull/N/merge`, and an event payload naming the feature commit as `head.sha`. The fixture asserts its own merge really is authored by `noreply@github.com`, so it fails loudly rather than silently testing nothing.
- The unfetched-head case (SHA alone, no description) and a Buildkite branch build (hosts that were already correct read identically).

Confirmed the suite fails with the fix reverted — `COMMIT_SHA` comes back as the merge commit — so it exercises the changed path rather than passing incidentally.

`bazel test //crates/...` 18/18, `aspect format` and `aspect lint` clean.


joshcbarnes:
Pushed `a9f6e766`: the resolved SHA reached `git show` unvalidated.

`detect_build_commit_sha` draws on `GITHUB_SHA` / `BUILDKITE_COMMIT` / `CIRCLE_SHA1` / `CI_COMMIT_SHA` / `ASPECT_GITLAB_SHA` / the event payload — environment anyone who can trigger a build controls — and `git show` reads a leading `-` as one of its own options. `git show --output=<path>` overwrites that path; I reproduced it directly against a scratch repo before fixing.

Gated on `is_commit_sha` (7-40 hex), falling back to HEAD otherwise, which is also the better read for `BUILDKITE_COMMIT=HEAD` (what Buildkite sets when no commit is pinned) — that previously ran `git show HEAD` by accident and now does so deliberately.

The unfetched-head behaviour this PR documents is untouched: a well-formed SHA absent from a shallow checkout still yields the SHA with no description rather than falling back. Added `_test_commit_sha_shape` and `_test_a_non_sha_never_reaches_git_show`; the latter asserts the named file still holds its contents, and with the guard reverted it fails with the file emptied.

One unrelated thing noticed while working here: this branch is based on a main that predates #1442, so its `bazelrc.axl` is the pre-config-groups version. Worth a rebase before merge.

PR #1462, opened 2026-09-16
refactor(wrapper): trim tools/bazel to the routing logic

gregmagolan:
Tightens the vendored `tools/bazel` wrapper from 479 lines to 169 with no behavior change. The goal is a script a team can read top to bottom when they vendor it.

What got simpler:

- **Header.** The 95-line preamble is now a three-row routing table and an env-var table. The long form lives in `tools/bazel.md`, which the header points at, along with `aspect setup tools-bazel-wrapper` as the way it gets vendored.
- **One PATH scan, not two.** The anti-inception block and `resolve_bazel_real` each walked `type -aP bazel` and canonicalized paths by hand to skip the wrapper itself. Now one loop resolves `BAZEL_REAL` up front using `[[ ! "$candidate" -ef "$0" ]]`, and the `ASPECT_CLI_RUNNING` bypass is a single `exec` line after it.
- **One membership test.** `contains` plus `ASPECT_VERBS_STR` / `BAZEL_VERBS_STR` / `KNOWN_VERBS_STR` collapse into `has <needle> <items...>`, a string match over `" $* "`.
- **`run_bazel`.** The five `trace_exec bazel "" …; exec …` pairs become one helper, and `trace` loses its category/shadow parameters: callers pass the suffix.
- **One aspect route.** The wrapped-verb and custom-task branches differed only in the vanilla-bazel fallback when `aspect` is missing, so they merge. A verb bazel also has gets the `ASPECT_WRAPPER_SKIP=1` nudge and the fallback; anything else gets neither.
- **`on_recognized_ci`** is one `[[ -n "${CI+x}${BUILDKITE+x}…" ]]` test inline in `trace`.

```
For tools/bazel after the trim:

┌──────────┬────────┬───────┐
│   Kind   │ Before │ After │
├──────────┼────────┼───────┤
│ Code     │ 203    │ 96    │
├──────────┼────────┼───────┤
│ Comments │ 248    │ 51    │
├──────────┼────────┼───────┤
│ Blank    │ 28     │ 22    │
├──────────┼────────┼───────┤
│ Total    │ 479    │ 169   │
└──────────┴────────┴───────┘
```

Contracts kept: the `# aspect-build/tools-bazel-wrapper` marker on line 2 and `ASPECT_BAZEL_WRAPPER_VERSION="2"` (the CLI parses both), every env var, exact trace and install-hint text, exit codes, and bash 3.2 compatibility. The version stays at 2 since the routing rules and env-var contract are unchanged.

Also updates three comments in `tools/wrapper-test.sh` that named removed helpers, and the size note in `repro_commands.axl` that said the wrapper is ~28 KiB (it is ~7 KiB now; the 64 KiB probe cap is unchanged).

---

- Searched for relevant documentation and updated as needed: yes (`tools/bazel.md` describes behavior, not implementation, and needed no change)

- The vendored `tools/bazel` wrapper is a third of its former length with the same routing; re-vendor with `aspect setup tools-bazel-wrapper` to pick it up.

### Test plan

- Covered by existing test cases: `./tools/wrapper-test.sh` passes all 87 cases, and again with `WRAPPER_BASH=/bin/bash` (macOS bash 3.2). `shellcheck tools/bazel` is clean.


jeffpignataro (approved):
:white_check_mark:

PR #1465, opened 2026-09-16
feat(setup): add `aspect setup workspace-data` for --workspace_status_command

joshcbarnes:
## Problem

A vanilla `bazel` call reaches a BES backend with no attribution. Commit, branch, PR and run URL are injected per invocation by `aspect <task>` (`register_bes_metadata`), so a build started any other way shows up in the Web UI unlabelled. Repos have been closing the gap with hand-rolled bash that reimplements `get_build_metadata_flags` and then drifts from it.

The obvious fix — write the flags into an rc file — doesn't work. `--build_metadata` accumulates rather than overrides, so two concurrent jobs sharing a `$HOME` don't overwrite each other's lines, they merge: every build carries two `COMMIT_SHA`s. And the values are stale the moment the commit moves. Per-invocation data does not belong in config.

## What this adds

`aspect setup workspace-data` prints the same metadata as Bazel workspace status — `KEY value` lines on stdout:

```sh
bazel build //... --workspace_status_command="aspect setup workspace-data"
```

Bazel runs the command at the start of every invocation, so the values are always the current build's. That inverts the problem usefully: **the flag naming the command is identical for every job, commit and workspace**, so it is safe in a shared `~/.bazelrc` and reaches every vanilla `bazel` call on a CI runner — including sibling workspaces a repository's own `.bazelrc` cannot reach — while the data behind it is never shared between two builds.

The keys are the ones `aspect <task>` already sends as `--build_metadata`, and the Web UI merges both sources with `--build_metadata` winning (`lib/bazel_results.axl::merged_meta`), so the two routes reach the same fields. They are written unprefixed so they land in `volatile-status.txt` — a `STABLE_` key invalidates every stamped action when it changes, the wrong trade for data that changes every commit.

`get_build_metadata` now holds the collection, with both renderers built on it, so the flags and the status lines cannot drift.

## Relationship to #1459

These two started as one change and have been separated. **#1459 owns the metadata's correctness** — resolving the PR head commit instead of the forge-synthesized merge, across GitHub `pull_request` / `pull_request_target` and GitLab merged-result and merge-train pipelines. This PR owns **where the metadata can be delivered**, and touches `build_metadata.axl` only to split collection from rendering.

No file overlap remains: an earlier revision of this branch carried its own narrower version of the commit fix and its own `bazel/build_metadata_test.axl`, both dropped. #1459's version is strictly better — mine used `detect_commit_sha` alone, which falls through to a bare `CI_COMMIT_SHA` and so would not have fixed GitLab at all.

They are independent and can merge in either order. Until #1459 lands, this command prints the merge commit's details on a GHA PR build, exactly as `aspect <task>` does today.

## Verification

Against Bazel 9.0.1, in throwaway workspaces:

- A `--workspace_status_command` with arguments is accepted, and a bare command name resolves on `PATH` — no wrapper script or absolute path needed.
- Values containing spaces survive unquoted (the value is the rest of the line). The same data as a `--build_metadata` flag would need shell-style quoting; unquoted, Bazel splits it and errors.
- Keys land in `volatile-status.txt`, not `stable-status.txt`.
- Task narrative goes to stderr; stdout carries only the data, so Bazel's parse is clean.
- Exits 0 in a bare directory with no git, no `.aspect/` and no `MODULE.bazel` — the shape sibling workspaces have. Bazel fails a build outright when the status command exits non-zero, so this matters; the residual exposure is a workspace whose own `.aspect/config.axl` fails to evaluate, or `aspect` missing from `PATH`, both conditions under which `aspect <task>` is already broken there. Documented in the module docstring.

`aspect dev test-workspace-data`, all 50 suites via `tools/run-axl-suites.sh`, and the 99 `aspect-cli` Rust unit tests pass.

## Cost

~353ms per invocation (release build), on the critical path of every `bazel` call using it. Negligible for a CI build, less so for a script calling `bazel` in a loop. Almost all of it is CLI startup rather than the `git` call, so a Rust fast path that skips AXL task loading would cut it substantially if it turns out to matter.

## Not covered

Workspace status carries attribution, not wiring. A vanilla `bazel` in a sibling workspace still needs `--remote_cache` and `--bes_backend` from somewhere — that half is unchanged and still has no off-runner story.


crates/aspect-cli/src/builtins/aspect/workspace_data.axl
```diff
@@ -0,0 +1,55 @@
+"""`aspect setup workspace-data` — print this build's commit, branch, PR and CI
+metadata as Bazel workspace status, for `--workspace_status_command`.
+
+    bazel build //... --workspace_status_command="aspect setup workspace-data"
```
gregmagolan, line 4: Lets also add alternate instructions for calling 'aspect setup workspace-data' from an existing workspace_status script

crates/aspect-cli/src/builtins/aspect/workspace_data.axl
```diff
@@ -0,0 +1,55 @@
...
+so stdout carries them and nothing else — every narrative line this CLI prints
+already goes to stderr, which Bazel leaves alone. A write error other than a
+closed pipe propagates; a closed pipe means Bazel is no longer reading, and
+there is nobody left to tell.
+"""
+
+load("@aspect//bazel/build_metadata.axl", "get_build_metadata", "workspace_status_lines")
+
+def _workspace_data_impl(ctx: TaskContext) -> int:
+    lines = workspace_status_lines(get_build_metadata(ctx.std))
```
gregmagolan, line 57: Such a short impl function 🌮

gregmagolan (approved):
🌮

PR #1469, opened 2026-09-17
fix(delivery): Retry the deliveryd health probe before failing the task

b3cramer:
The delivery task probes deliveryd's `/health` once and exits 1 on anything but 2xx. On a fresh runner that is the daemon's first Redis dial (2s timeout), so one slow dial fails the whole step. Bedrock's nightly hit this: `HTTP 503: unhealthy: redis connection failed` while MemoryDB was healthy.

The probe now retries a 503 or transport error up to three times, one second apart. Other statuses are not retried. `pause` and `probe_attempts` are injectable for tests.

Refs CS-286.

---

### Test plan

`aspect dev test-deliveryd` pins retry-then-succeed, give-up after three probes, no retry on 500, and `probe_attempts = 1`. Not verified on a live runner.


jeffpignataro (approved):
:white_check_mark:

PR #1470, opened 2026-09-17
ci(release): stop dirtying the tree during release, bump action-gh-release to v3

jeffpignataro:
The tag release job takes ~28 minutes on a 32-core runner, and the last two tag runs (v2026.38.18, v2026.38.19) failed in the release-upload step. Two independent fixes, one commit each.

**1. Write release assets under `RUNNER_TEMP`, not the checkout.** The three `bazel run //...:release` steps copy binaries into `upload_root/assets` inside the git checkout. `--config=release` stamps via `bazel/workspace_status.sh`, whose `STABLE_BUILD_SCM_LOCAL_CHANGES` key runs `git status --porcelain`. After the first copy lands, the tree has untracked files, the key flips clean → dirty, and the stable-status file that every stamped rustc action takes as an input changes. Every first-party crate is stamped (`bazel/rust/defs.bzl` sets `stamp = -1` for any `version_key` other than `None`), so the second and third `bazel run` recompile the world serially. On the last successful tag run that was 549s + 255s, against 2s for the first `bazel run`, which ran while the directory was still empty. The shipped binaries were also stamped as a dirty build. Writing the assets outside the checkout keeps the tree clean; upload-artifact roots the artifact at the common ancestor of the matched files, so the downloaded layout (`assets/...`) is unchanged for `publish_release.yaml`.

**2. Bump `softprops/action-gh-release` v2.6.2 → v3.0.3.** The failures were `Error uploading` and `Error creating asset temp dir`. Neither string is in the action's source at the pinned commit; both are response bodies from GitHub's asset upload service. v2.6.2 fires all ~26 uploads at once and only retries a 422 `already_exists` race, so the first 5xx fails the step. v3.0.2 hardens the streamed uploads and fixes small checksum-asset uploads; v3.0.3 adds safe classification of malformed API errors. v3 runs on Node 24, which the hosted runner already forces. The blank-line removals in `publish_release.yaml` are the repo's YAML formatter, applied by the pre-commit hook.

Not in this PR: the remaining ~10 minutes is the `lto=fat` link tail (16 single-threaded links), and the build runs with no remote cache because setup-aspect cannot configure vanilla `bazel` on that runner without a newer aspect-cli. Both left alone on purpose; the release build must not depend on the aspect binary it is producing.

---

### Test plan

- Manual testing: `build_release_artifacts.yaml` dispatched from this branch (link in the PR comments once it finishes). Expect the second and third `bazel run` steps to be cache hits and the job to drop from ~28 to ~15 minutes, with the uploaded artifact still laid out as `assets/*`.
- The action bump is exercised by the next real tag run; the build-only dispatch does not reach `publish_release.yaml`.

https://claude.ai/code/session_01BG3pdPMNfVGCgJus2gbcyo


jeffpignataro:
Test dispatch of `build_release_artifacts.yaml` from this branch: https://github.com/aspect-build/aspect-cli/actions/runs/35278419203 (success).

| Step | main (run 35170402445) | this branch |
|---|---|---|
| `bazel build` (all bins) | 878s | 884s |
| `bazel run //crates/aspect-launcher:release` | 2s | 2s |
| `bazel run //crates/aspect-cli:release` | 549s | 2s (3505 action cache hits) |
| `bazel run //crates/axl-docgen:release` | 255s | 1s (146 action cache hits) |
| Build job total | 28m 29s | 15m 16s |

The uploaded artifact has the same layout (`assets/*`, 39 files, 208MB), so `publish_release.yaml` needs no change. The `bazel build` step is unchanged, as expected: that's the fat-LTO critical path, not touched here.

gregmagolan (approved):

PR #1471, opened 2026-09-18
fix(bes): drain the response stream while a bidi send is in flight

joshcbarnes:
## The bug

`tokio::select!` runs a chosen branch's body to completion, so awaiting the request-channel send inside `drive_stream`'s loop stopped the ack branch from being polled until it resolved — up to `send_stall_timeout`, 60s by default. Measured at 2.26s against a 2s timeout with a backend that stops reading.

Acks alone survived that: h2 buffers them locally and `drain_acks` recovers them on the failure path. A server **status** did not. `drain_acks` loops on `while let Ok(Some(Ok(resp)))`, so it stops at the first `Err` and drops it, and the sink returned its own synthesized `deadline_exceeded` instead.

So a non-retryable rejection — bad credentials, invalid argument — arriving while a send was stalled got classified as retryable, replayed for the whole retry budget with backoff, and finally reported as:

```
giving up, build events were not delivered: request stream stalled for 60s
mid-send (server stopped reading)
```

hiding a `PERMISSION_DENIED` behind a misleading cause.

## The fix

`send_bounded` becomes `send_draining_acks`, selecting over the send, the response stream, and the stall deadline, so acks are harvested throughout the wait and a status short-circuits it.

Acks are collected into a caller-owned `Vec` rather than applied directly, which lets the replay loop — holding a borrow of `state.buffer` — share the same helper. That path had the same blind spot *within* each send, between its non-blocking drains.

Response-stream ends now route through one `response_end_outcome`, which the main loop's error arm also uses instead of repeating the retryable/fatal mapping. `ReplayResponseEnd` becomes `ResponseEnd`, since it is no longer replay-specific.

`context` now takes the same bare phrase everywhere. It had two conventions — a pre-spaced `" during replay"` for `response_end_outcome`, a bare one for `send_draining_acks` — passed a few lines apart at one call site. The mid-send call passed `""`, making a close observed mid-send byte-identical in logs to one seen on the response arm; it now reads `... prematurely mid-send`.

## What this is not

I chased a flow-control deadlock first and could not build one. Acks are ~20 bytes against a 64KB HTTP/2 window, so the server's ack writes never block, and any request the server reads frees the single slot a pending send needs. Two attempted repros completed `Done` in ~0.1s. The defect is the lost status, not a hang.

## Tests

Adds `StopReadingThenFatal` and `AckThenStopReading` server modes.

`server_status_during_a_stalled_send_is_not_masked` guards the fix — with the drain arm deleted it fails:

```
expected Fatal(PermissionDenied), got Transient(status: 'Deadline expired
before operation could complete', self: "request stream stalled for 10s
mid-send (server stopped reading)")
```

`stalled_send_recovers_acks_written_before_the_stall` asserts the ack accounting across a stall. It passes with or without the fix, since `drain_acks` recovers the same acks on the failure path — its doc comment says so rather than implying coverage it doesn't have.

All 39 sink tests and the full `//crates/axl-runtime:test` pass.


gregmagolan (approved):
lgtm

thesayyn (approved):

